PR #99, opened 2015-01-14
Add kOS CPU and Telemachus Datalinks to all ModuleCommand

jwvanderbeck:
This just makes it easier to play with kOS and Telemachus instead of having to add the parts.


Felger:
Suggestion: Add Telemachus to antennas, not modulecommand.  Makes more sense.  Does Telemachus interplay with RT at all?

jwvanderbeck:
Many command pods have a built in passive RT antenna, so why would it need another antenna for Telemachus?

Not positive on integration with RT directly, but I don't think so.

Really Telemachus doesn't modify gameplay in anyway, so this is just a "nice" thing.

Felger:
So it'd be caught by the moduleantenna rather than modulecommand (assuming that's the right module name, would have to check.)  Either way it doesn't really matter that much. I just think it's slightly more realistic to have Telemachus tied to the communication equipment rather than the processor equipment. No Antenna = No Telemetry.

jwvanderbeck:
How about I add telemachus to antennas and modules that have an RT2 passive antenna.  That makes sense.

Felger:
That works, if you want to get fancy, you could do something like:
    @PART[*]:HAS[MODULE[ModuleCommand]]:NEEDS[!RemoteTech]:FOR[RealismOverhaul]
    {
        //add telemachus
    }
    @PART[*]:HAS[MODULE[ModuleRTAntenna],etc]:NEEDS[RemoteTech]:FOR[RealismOverhaul]
    {
        //add telemachus
    }
Which would add telemachus to antennas only if RT is installed, otherwise it would install to command modules.  Thoughts?

jwvanderbeck:
Yeah something like that.  Once I finish up my work on Explorer 1 here i'll take a second pass at this.

jwvanderbeck:
Updated configs to only apply Telemachus to a part if:

* IF RT2 is installed and the part has an RT2 Antenna, passive or active
* IF RT2 is NOT installed and the part has a stock antenna (ModuleDataTransmitter)

PR #102, opened 2015-01-14
Atomic Age support, bugfixes for UA1207, add UA-156x and AJ-260 configs

ferram4:
Included are engine configs and RealPlume configs for both engines from Porkjet's Atomic Age mod.  The names and exact performance are somewhat speculative due to these being based on study and concept designs, but are based in reality: the Nuclear Lightbulb is a United Tech engine, since they did most of the research on it (as well as creating the report referenced for its performance) and the Pratt & Whitney TRITON is one of the few LANTR designs that I can find.

FX for the TRITON is supposed to change with increasing LOX mixing, but requires future support from Real Fuels to shut off the unused effects.

In addition, there was a small discrepancy between the Isps of the FASA UA1207 and KW 1207, with the latter having Isps 5 seconds higher across the board; I brought it down to the FASA numbers in order to maintain consistency.


ferram4:
I've added some more tweaks and fixes for the UA1207; apparently, the FASA version wasn't having its thrust curve applied and was using the UA1206 thrust curve, and a name change at some point meant that the RealPlume FX weren't applied to it anymore.

Further, I started a WIP config for KW stuff, and added the same FX that was added to the FASA UA120x motors to the KW UA1207 and USRM, since they should be similar.

PR #103, opened 2015-01-15
Explorer1 battery balance

jwvanderbeck:
Rebalances Electric Charge and Consumption in FASA Explorer 1 to as closely mimic the real probe as possible.

Reference: http://en.wikipedia.org/wiki/Explorer_1

No direct data on total battery power could be found, but the mission profile called for batteries that could power a high power transmitter for 31 days and a low power transmitter for 105 days.  Due to KSP limitations on consumption rates, and the lack of dual battery setups, I have set this up to work as follows:

Two antennas.  Low Power Transmitter is an always on passive omni antenna with lower range.  High Power Transmitter is an activated omni antenna with twice the range of the LP transmitter.

Mission profile of: HPT + LPT for 31 days, LPT only for the remaining 74 days.

**NOTE:** I changed :AFTER[RemoteTech], to :FINAL because it was not working properly with the former wildcard.  For some reason that I can't explain the values in RemoteTech were not being changed.  Changing it to FINAL fixed this.


jwvanderbeck:
I cleaned up this PR after you accepted a previous one that caused conflicts.  So this should be good to merge cleanly again.

Felger:
Awesome stuff, thanks!

PR #104, opened 2015-01-16
Fixes, fixes, fixes

Hattivat:
(the first two commits are meaningless, I used them to update my fork)

- fixed Lionhead's resources
- added a linearRCS model (copied from RP-0, meaning they can now delete it from their folder and link to this one, I guess) in order to fix the problems stemming from Ven's Stock Revamp messing with stock LinearRCS model (causes it to behave funky when welded to something else)
- simplified Vostok's radios


Felger:
Hattivat, you're the best!

PR #105, opened 2015-01-17
Partial ESA support

Hattivat:
as we discussed, for now it's Ariane 4, Ariane 5, Diamant, and Europa 1. A minor bug in one of the other configs slowed me down, so ATV should be ready tomorrow.


Hattivat:
ATV ready to roll

PR #110, opened 2015-01-26
added VSR Mk1-2 Pod

bax-:
Heat shield added to Mark 1-2Pod by Ven’s Stock Revamp factored in


bax-:
The next version of VSR will probably add the heat shield to the Pod dynamically itself so this config file won't be necessary:

https://github.com/VenVen/Stock-Revamp/commit/3de3cbc34a361ee7f0042e77ce205c116b59daea

Please close

PR #118, opened 2015-01-31
fix for MM 2.5.9, tweak for changine RC sizes

bax-:
- fix for MM 2.5.9 (& to ,)
- apply RealChute sizes dynamically: Instead of deleting MODULE[ProceduralChute] we could just delete the sizes. Then the config doesn't break that quick if RealChute changes something in the Module in later versions


NathanKell:
Weird that the first commit is already in master.  Ah well, the second looks cool. :)

PR #123, opened 2015-02-02
Engine config updates

TheVoid2:
Updated configs for RL10 and J-2 versions. Cleaned up names and descriptions of Atlas and Saturn engines.


Felger:
Replaced by #126

PR #127, opened 2015-02-02
Lr101 fixes

jwvanderbeck:


jwvanderbeck:
Updated LR-101 stats as per ferram4's input.

PR #128, opened 2015-02-02
Update RO_CSS.cfg

advicedawgg:
Just a title fix. SSRB's are not 5 segment.


jwvanderbeck:
What is the proper name for these?

advicedawgg:
Well they are 4 segment space shuttle solid rocket boosters. I'm not sure there is a technical name for them.
They were manufactured and refurbished by Thiokol until they merged with ATK. But that was very late in the shuttle program.
Space Shuttle Solid Rocket Boosters?

PR #143, opened 2015-02-12
FASA RVE Behaviour Consistency

stratochief66:
All FASA capsules will now behave consistently with directional
heatshields under DRE in RO.

Coded such that they do not rely on any of the original FASA files
having MODULE[ModuleHeatShield]

### DIFF
--- a/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -336,7 +336,7 @@
 	@mass = 0.152
 	!MODULE[ModuleHeatShield]
 	!RESOURCE[AblativeShielding]
-	MODULE[ModuleHeatShield]
+	MODULE
 	{
 		name = ModuleHeatShield
 		direction = 0, -1, 0
--- a/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -335,7 +335,9 @@
 	@description = This part is the ablative heat shield for the Apollo Command Module.
 	@mass = 0.152
 	!MODULE[ModuleHeatShield]
+	{ }
 	!RESOURCE[AblativeShielding]
+	{ }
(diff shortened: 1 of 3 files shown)


NathanKell:
Looks cool! Just a note--when you delete a node, you need to do
!NODE {}
(the {} are important--otherwise MM will try to delete a *key* called NODE, not a node called NODE).
Can you fix that?

stratochief66:
No worry. I wasn't sure what the reasoning was for it, as it appears that sometimes curled braces seem to affect the behaviour in a case like this, and other times the did not.

NathanKell:
Well, if the node already doesn't exist, then failing to delete it wouldn't matter, so that might be the 'did not' cases? But if not, tell sarbian, that's a bug. :D

stratochief66:
Does it look good now, NathanKell?

NathanKell:
Oops--and I missed this on the Gemini pull so that needs fixing too: When you see @key,x = foo, that means "at the x'th key, change its value to foo. So first, don't do key,1, and second, include *all* the module, not just the changed stuff (it looks like basically what you did for the curves was just remove the @'s.... :] ). If you are adding, rather than changing, it needs to look like this (obviously use the correct values instead of these for the curves...)
https://github.com/KSP-RO/RealismOverhaul/blob/master/RealismOverhaul/RO_SuggestedMods/AIES/RO_AIES_Command.cfg#L447-L464

NathanKell:
(Sorry I didn't notice that issue with the Gemini pull at the time; I guess I was kinda sleepy >.> )

stratochief66:
No problem. Thanks for helping me along, it is very educational.

In a key set like this, what are the trailing 0's for?

key = 2000 160 0 0

I imagine that the first value is the temperature, and the second is the intensity of ablative loss.

Is MM code written in a specific language, or what coding language is it most like?

NathanKell:
The trailing zeroes are the in and out-tangents. [See here](http://forum.kerbalspaceprogram.com/threads/93332) for information about KSP float curves. Ideally we'd actually use better tangents, but 0 0 works ok for now.

MM code is not a specific language; it's just a set of first-character operators (and a few other things) grafted onto ConfigNode (which Mu wrote for KSP as a way of saving and loading data).

It all looks good now--anything else before I merge?

stratochief66:
Nope, good to merge!

PR #153, opened 2015-02-17
Adding a J-2X Config for the FASA J-2 Engine

JandCandO:
I already added a J-2X config for AIES's J-2, so I though I would extend it to FASA's J-2.

I just need to know if the ModuleEngineIgnitor declaration in the J-2X config overrides the one in the ModuleEngineConfigs' declaration. Or if there will be nasty clashes....

### DIFF
--- a/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
+++ b/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
@@ -916,6 +916,53 @@
 				key = 1 200
 			}
 		}
+		CONFIG
+		{
+			name = J-2X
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
I think we are keeping the J-2X as a separate engine part, not with the other J-2's

JandCandO:
Oh ok if that's the case, then I'll delete this pull request. I kind of want this to float around for a day before that so people can see it and say what they think should be done, though.

NathanKell:
Yes, the modern J-2X has as much in common with the J-2, as Delta IV does with Thor-Delta ("they lifted up the name and built a new rocket underneath"). Thanks though!

JandCandO:
Ok thanks for the info on your plans! I'm deleting this now.

Issue #144, opened 2015-02-12
Engines checklist


NathanKell:
- [x] Constelacion -> J-2X
- [x] Fix 1kn and 0.5kn thrusters to obey propellant thrust scaling (and the AIES one to be 2kN base since 2 nozzles). That means call it 1kN class, maybe? Since with NTO/AZ50 it'll be more like 2kN...
- [x] Give the above (generic thrusters) EngineIgnitor configs; pressure-fed unlimited-restart ullage=false.
- [x] Backport VSR engines
- [x] Add LR-105-3
- [x] Port over VSR-engine changes (to SquadEngines) to other same-config engines
- [x] mea culpa: CECE should be an RL10 alt config, not a new part.
- [x] RLA Kingfisher -> Bell 8xxx engine/Agena engine (was CECE)
- [x] RLA Albatross -> VINCI (was J-2)
- [x] KW Globe X-10S "Thor II" SRB -> 5-seg Shuttle SRB (was UA-1207)
- [x] Set AIES Exper to RL-60 (was RL-10B-2)

Regarding the generic thrusters: Right now, RCS configs (check out an RCS module) have differing thrust and Isp based on what propellant mixture is used; coldgas has low thrust and Isp, hydrazine is decent, and storable hypergolics are quite good indeed. The "generic thruster" parts, however, only increase Isp based on propellant change, not thrust. They should be scaled in thrust as well, like RCS.

NathanKell:
Possibles:
- [ ] Set Exper to RL-60 or CECE not RL10B-2

jwvanderbeck:
Working on the first right now.  Have the base configs in place need to test in game to ensure proper scaling (which will no doubt make me want to punch babies)

NathanKell:
I'm working on the backport for the record, about 1/4 done, although right now I'm dealing with figuring out what happened to RLA.

jwvanderbeck:
You also need to let me know what I should do with the RLA J-2.  Either keep it as the J-2 line but remove the J-2X, or make it a J-2X only engine and remove the J-2 line.

jwvanderbeck:
Which mixture should be the "1.0" point on the generic thrusters?  The 0.5kN thruster in RLA seems to have already been done and its 1.0 point is in the MMH+NTO mixture.  Yet when I look at say the Squad RCS, the 1.0 point is in the Hydrazine mixture.

jwvanderbeck:
[...]

PR #160, opened 2015-02-23
Migrated back ported engine changes to other part packs

jwvanderbeck:
For #144


NathanKell:
@jwvanderbeck  you have to also add a RESOURCE block otherwise the RD-58 won't ignite at all...it should have 2.5 units of TEATEB. Unless the diff isn't showing that?

PR #163, opened 2015-02-24
Additional SRM correction, UA-1207

stratochief66:
I missed a line in my changes last night, this one will make the UA-1207 configuration the default, such that the thrust curve applies to the part by default.


jwvanderbeck:
Thanks

PR #165, opened 2015-02-25
Updated .version to 8.1.1

jwvanderbeck:


Felger:
I really do need to set up Travis to do these releases so I don't forget
stuff like this.

jwvanderbeck:
Well if nothing else, I can easily port the one I use for TestFlight to RO.  It autogenerates the .version file based on the tag.  But yeah you need to setup travis integration first.

NathanKell:
Well shoot. And yes, we do need travis...

PR #170, opened 2015-03-06
FASA Capsule Changes

stratochief66:
Changed the behaviour of a number of Mercury capsule parts. The capsule
cap will now pop off on re-entry. The RCS pods are now attached to the
base capsule corrected, RCSFX functions. LES now stages as the Apollo
LES does.

Added a txt file for keeping track of issues ID'd and corrected as I go
through the capsules. Will be removed on completion of my pass through.

### DIFF
--- a/RealismOverhaul/RO_SuggestedMods/FASA/FASA bug notes.txt
+++ b/RealismOverhaul/RO_SuggestedMods/FASA/FASA bug notes.txt
@@ -1,0 +1,11 @@
+Mercury:
+
+Made Mercury LES stagable, so it behaves like the Apollo LES and is jetissoned on staging
+
+Made Mercury cap attach to the capsule, rather than the parachute
+
+Attached the roll RCS and properly, corrected RCSFX config
+
+Removed stock reaction wheel from Mercury capsule attitude control
+
+Made the parachute attach to the capsule, adding an attachment point to the capsule to support
--- a/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -96,7 +96,7 @@
 	@MODEL
 	{
 		@scale = 1.0, 1.0, 1.0
-		@rotation = 0, 34, 0
+		@rotation = 0, 0, 0
 	}
 	@node_stack_top = 0.0, 0.357075, 0.0, 0.0, 1.0, 0.0, 1
 	@node_stack_bottom = 0.0, 0.157075, 0.0, 0.0, 1.0, 0.0, 1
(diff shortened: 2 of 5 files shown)


NathanKell:
Ah, thanks!
A few questions on this:
1. Since there's now @MODULE[ModuleRCS] is it because FASA now adds its own ModuleRCS module, which we're modifying?
2. The "model =" should stay as they are--we bundle that model, because the VSR one has different transform names so we can't just assume the existing one is correct.
3. I'm a bit confused by the parachute and cap change--why was that change made? As things were, the cap still would decouple on reentry...

stratochief66:
1. Yes, the current FASA Mercury Pod has a ModuleRCS. Before my change, the pod still had the ModuleRCS set to consume Monoprop plus a new ModuleRCSFX.

2. The models in the previous version appeared to have no texture/colouring. I will try switching them back and make sure things work fine with the previous ones.

3. The parachute and cap were sticking together when the cap was staged off during re-entry. I tried a number of other (more obvious) attempts to get them to separate as expected and they didn't work. Previously the cap would pop off, but stick to the capsule, even with a very high decoupling force. With this change, the cap will pop off.

I noticed the way it was before my change, the cap would pop off if staged at the launch pad, but not during re-entry.

NathanKell:
1. Ah, I see. Good catch!

2. They are untextured, yes, but should be small enough as to be nearly invisible. If they're visible, maybe slightly decrease X and Z so they're more under-the-surface? Then again the linear RCS port is basically white even textured. :]

3. Huh. I would not have thought that the change would help; the reason for the "sticking" is due to the capsule having a lower ballistic coefficient than the cap, and thus a lower terminal velocity.

Also you might want to rename the PR to "FASA Capsule changes" since it now has the CSM/LEM too :)

stratochief66:
2. I will try moving them in just under the surface. My reason for fiddle with them to was to make the RCS gas puffs appear.

3. Are you sure? With the different in balistic coefficient, shouldn't the cap still pop off briefly if the decoupling force is high enough? To me, it looks like the models were sticking together, colliding or something. I would also ideally like the cap to have a higher ballistic coefficient than the capsule, because I imagine the cap would decouple then be 'left behind' in the sky as the cap is less aerodynamic than the capsule, but I could be mistaken in that mental image.

NathanKell:
1. If the model is just under the surface the jet *should* still appear, it should be long enough, but only testing will tell...

2. Higher BC = higher mass/surface area = higher terminal velocity = falls faster. But yes, while one might expect the cap to have a lower BC than the capsule, and it would in real life since the cap is *hollow*, that might not be true in KSP; it's a roughly 0.7m diameter flat plate while the capsule is a 1.9m one, so about 2/15ths the drag, and if the cap masses more than a hundred kilos or so it will have a higher BC. However, if the issue is collider failure due to movement, then never mind. :] Either way, however, your solution solves the problem, so...

stratochief66:
Jets appear fine now, other than the jets appearing from the ahistorical radial rcs ports on the nose.

PR #171, opened 2015-03-06
Removed leftover git conflict

aw94:
regex said the ModuleEnginesFX version is likely to be the correct one, so I
went along with it.


aw94:
Not too sure about this...

jwvanderbeck:
No that's not correct but thanks for highlighting there was an issue.  Someone needs to smack @ferram4  for the mess and probably @NathanKell for merging it :p

aw94:
:( Should have taken regex's advice and tested it. Bit too hasty...

jwvanderbeck:
It would have worked either way, but it would have been wrong :)

- John Vanderbeck
- http://www.johnvanderbeck.com

aw94:
Ah, never mind then.

PR #178, opened 2015-03-09
Fix TACLS scubbers and purifiers

### LINKED ISSUE #177: CO2 scrubbing with lithium hydroxide produces water?
Technically, CO2 scrubbing with lithium hydroxide produces water:

2LiOH + CO2 → Li2CO3 + H2O

The TACLS scrubbers in RO use this reaction.

While it's true that lithium carbonate isn't very soluble in water, I can't find any record of crew drinking the water from the CO2 scrubber, because doing so would probably cause them to *seriously chill out* due to its pharmacological effects. I believe this is undesirable in manned space flight.

I'd suggest the scrubbers produce waste-water rather than water.

pjf:
Note that 'true' in output resources means it's okay for that resource
to be thrown away. Reference:

https://github.com/taraniselsu/TacLifeSupport/blob/master/Source/TacGenericConverter.cs#L61-L64

This is backed up by the converters that come with TACLS itself.

This commit fixes the following:

- The air filter will no longer run in a fully oxygenated environment.
- Water splitters will no longer run in a fully oxygenated environment.
- Water purifiers will no longer vent water, but insist on keeping waste.
- CO2 scrubbers no longer fascinated with keeping waste, either.
- CO2 scrubbers produce WasteWater when appropriate, not Water. (Closes #177)
- Potassium superoxide scrubber now produces water, rather than requiring it.
- CO2 scrubbers will continue to run as long as CO2 is available. No
  more kerbonauts dying from CO2 exposure because the wastewater tank
  was full, or the cabin was fully oxygenated.

### DIFF
--- a/RealismOverhaul/RO_RecommendedMods/TACLS/RO_TACLS_Tanks.cfg
+++ b/RealismOverhaul/RO_RecommendedMods/TACLS/RO_TACLS_Tanks.cfg
@@ -91,7 +91,7 @@
 	{
 		@conversionRate = 3.0	// # of people - Figures based on per/person
 		@inputResources = IntakeAir, 0.035, ElectricCharge, 0.010
-		@outputResources = Oxygen, 0.0073, true
+		@outputResources = Oxygen, 0.0073, false
 	}
 }
 @PART[TacCarbonExtractor]:FOR[RealismOverhaul]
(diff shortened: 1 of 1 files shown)


stratochief66:
Nice! I have been eyeing how those bugs were affecting the FASA hardware. I will test this code out.

pjf:
@stratochief66 : Oh drat, do we have similar bugs in FASA parts as well? I was looking at the TACLS parts last night as I was placing them for RP-0, but I hadn't glanced at any of the others.

I'm pretty sure I once wrote some terrible Kerbal fan-fic due to one of these bugs; my fuel-cells refused to generate electricity because the water tanks were full, and so Bill, Barb, and Jeb all had to drink as much water they could to keep the life-support running.

PR #185, opened 2015-03-15
Adding configs for RD-270 and RD-270M

jbengtson:
This includes config files for the RD-270 and RD-270M.  The configs generate two engines so that the pentaborane variant can use a green backflame in its SmokeScrren effects.  I have added these configs as additional files because they copy engines that already exist.


jwvanderbeck:
In my opinon they should be in the original file though.  If we made new files every time we copied an engine it would be a nightmare.

jbengtson:
Excellent point but the distinction here, in my mind, is that these engines never actually flew.  The RD-270 was extensively tested and the 270M was a "study" based on the 270.

NathanKell:
I think it's a useful enough distinction in this case, one that was done for the M55, and one I probably should have done for the Aerobee Sustainer. I have my issues with the monolithic-cfg approach. :]

jwvanderbeck:
In that case then the file names should still have some logic to them so we actually can find these things later.  Maybe for engines that never flew, put them in a separate file named "speculative" or something.

PR #186, opened 2015-03-16
Overhaul RCS to standard configs and costs.

pjf:
This PR means that rather than copying and pasting the same RCS configs all over the place (and occasionally copying the wrong ones), parts can now just say:

    useRcsConfig = RCSBlock
    useRcsConfig = RCSBlockHalf
    useRcsConfig = RCSBlockQuarter
    useRcsConfig = RCSBlock15x

And theough the magic of ModuleManager, they'll get the right configs.

Note that this PR is from a branch in the RO repo itself, so you can amend and append to it as required.


pjf:
Also hell yes, 1.5k fewer lines of code with this change.

jwvanderbeck:
While I **really** like this, I hate the fact that we are moving even more stuff into :FINAL to make this work and would **really** prefer to see us move to a pre-processor instead, especially since it has a much wider use than hacky MM files.

(don't take this the wrong way please, this is awesome work, i'm just looking at the long road)

NathanKell:
@pjf, very cool! A few thoughts (of course).
1. FINAL should not be necessary; AFTER[RealismOverhaul] should work. I don't believe even the cloned parts are not FOR so it should be fine.
2. What is the 1.5x block in SXT? In SXT there's a RCS-on-boom, which is the regular block, and a 3x linear block.
3. I notice the thrust was changed on one of the ModuleRCSFX blocks.
4. The RLA block with "extra thrust"--is it maybe a 2-nozzle block?

@jwvanderbeck actually we can use this for engine configs too. For example, all RL10s can get useEngineConfig = RL10. :)

jwvanderbeck:
Yes it can do engine configs but I just see this as a really hacky and very confusing way of managing things.

Again please don't think i'm dismissing this, i'm not.  I just think it isn't the best way to approach it and will make the configs even more confusing.

JandCandO:
I like this idea. I think we should do whatever we can to unify things so we don't have as much of a mess of configs as we have now.

pjf:
1. **I agree entirely that a pre-processor or build system is better.** My usual yardstick for software development is "does this make it better than before?", to which I hope the answer is yes. (This also makes it easier to write a pre-processor or build system, since more RCS configs are aggregated.) Like many things, this started as a "quick change" that ended up not being so quick. ;)
1. **Yeah, :FINAL sucks. It's in there not because it's necessary, but because it made testing easier.** In `RO_Squad_Utility.cfg` we `+PART` a few things `:AFTER[RealismOverhaul]`. I wasn't sure of how MM orders rules when there are directories involved, and there's no `:AFTER[:AFTER[RealismOverhaul]]`.
1. **The 1.5x block was a surprise to me too!** It's the `SXTRCSRack` part. It appeared to be 50% more powerful than a regular RCS block, but the engine configs looked like they were 20% or 0% more expensive when I examined it. So I settled on a straight 20% more expensive for consistency, and left everything else the same.
1. **I changed the thrust on one of the ModuleRCSFX blocks because it claimed to be using hydrazine as default, but didn't provide hydrazine's thrust by default**. That *looked* like a mistake, but double-checking would be appreciated.
1. **Ideally I'd like to get rid of the ModuleRCSFX copy/paste as well**, but not this PR.

NathanKell:
@pjf 
1. Quite. :)
2. Mea Culpa for some of that. I believe that no AFTER[RO] patches in RO are before their original, so we can make them FOR[RO] safely, although we will then be dependent on path ordering. As an alternative, you could do :FOR[RCSFixing] or something, which would run after RO (and indeed after every plugin-based mod, because MM isn't purely by alpha), but still before RP-0 (its plugin is named RP0 so RP-0 is a later pass).
3. Evidently linearRCS has half the power of rcsBlock. That's how that happened. The cost for RCSRack was based on linearRCS x 3...
4. Ah, good catch.

PR #193, opened 2015-03-23
KVD-1 / CE-7.5 hydrolox engine configs

ferram4:
Includes:
Engine configs
RealPlume FX configs
hex-edited versions of the LV-1R meshes (stock and Ven's) to function as verniers, redistributed with permission

Also added a WIP "Credits" file for handling attribution if we redistribute assets and for handling any other credits we might need.


RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
```diff
@@ -3327,3 +3327,309 @@
...
+
+EFFECTS
+{
+  running
+  {
+    AUDIO
+    {
+      channel = Ship
+      clip = sound_rocket_spurts
+      volume = 0.0 0.0
```
NathanKell: @ferram4  this should be volume 0.0 1.0 unless the engine has perfect throttling.

ferram4: That only applies to direct throttle effects.  Running effects use thrust/max thrust as their input value.

PR #204, opened 2015-03-29
FASA Reproduction Craft

stratochief66:
Craft files for the basic NASA craft in RSS/RO constructed from FASA parts.

Includes: Mercury Redstone, Mercury Atlas, Atlas Agena, Gemini Titan,
Apollo Saturn V with LEM.

### DIFF
--- a/RealismOverhaul/Reproduction Craft Files/FASA Craft Files/Ships/VAB/Apollo Saturn V Lunar.craft
+++ b/RealismOverhaul/Reproduction Craft Files/FASA Craft Files/Ships/VAB/Apollo Saturn V Lunar.craft
@@ -1,0 +1,14348 @@
+ship = Apollo Saturn V Lunar
+version = 0.90.0
+description = 
+type = VAB
+size = 20.14136,115.527,20.14137
+PART
+{
+	part = FASAApollo.CM_4294479486
+	partName = Part
+	pos = 0.536364,104.8279,-0.05546219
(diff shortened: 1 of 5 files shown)


avielmenter:
Atlas-Agena Targeting Vehicle launch went well without any changes to the .craft file.

Gemini-Titan was really really close, but at fuel exhaustion I was in a 131 x 161km orbit. 

For Mercury-Redstone, the abort sequence wasn't set in action groups. I also noticed that, for some reason, it's rotated 90 degrees away from where rockets usually are by default. This lead to me going south instead of east, because I wasn't paying attention to the numbers on the nav ball. Otherwise, it worked fine and the staging sequence was good.

The Mercury-Atlas had the same small issues as the Mercury-Redstone (no abort sequence and rotated 90 degrees), but I was able to launch it into orbit and then deorbit the mercury pod successfully nonetheless.

The Saturn V is the only one for which I didn't do a full launch test (I examined it in the VAB). I noticed that the LEM staging wasn't set at all. While docking does tend to mess staging order up, if you have it such that the LEM staging is set then it's usually preserved correctly when you detach the LEM from the CSM in lunar orbit. I also noticed that you did not attach floats, or include an abort sequence. Otherwise, it looked fine. The staging sequence for the rocket itself was good.

TL;DR some minor usability nitpicks, but no major issues. All rockets could be launched successfully and accomplish their respective missions as they are in these craft files.

PR #215, opened 2015-04-02
Update RO_BobCat_SovietEngines.cfg

A1Ch1:
Added AJ26 config for NK-33 and adjusted thrust: the original can throttle from 50%-105% rated (source http://www.lpre.de/sntk/NK-33/index.htm) while the AJ26 can throttle 56%-108% (source http://www.spacelaunchreport.com/taurus2.html & engine mass source http://www.spaceflight101.com/antares-launch-vehicle-information.html)

Broke out manufacturer from title into %manufacturer under the impression this is the preferred format, if not apologies.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/BobCat/RO_BobCat_SovietEngines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/BobCat/RO_BobCat_SovietEngines.cfg
@@ -6,15 +6,16 @@
 	%node_stack_bottom = 0.0, -1.589812, 0.0, 0.0, 1.0, 0.0, 2
 	%node_stack_shroud = 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0
 	!node_attach = DELETE
-	%title = SNTK Kuznetsov NK-33
-	%description = Originally built for the N1F rocket, while the rockets were scrapped, the engines survived and are now being refurbished to power new designs, and manufacturing of new engines is starting.
+	%title = NK-33
+	%manufacturer = SNTK Kuznetsov
+	%description = Originally built in the late 1960s/early 1970s for the Soviet N1F rocket. Though the N1F was scrapped, the engines survived. Aerojet aquired several NK-33 engines in the 1990s and refurbished them as AJ26-62 engines for Orbital Science's Antares launch vehicle.
 	%attachRules = 1,0,1,0,0
 	%mass = 1.222
 	%maxTemp = 1700
 	@MODULE[ModuleEngines*]
 	{
-		%minThrust = 941.92
-		%maxThrust = 1682
+		%minThrust = 841
+		%maxThrust = 1766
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
Yes separating the manufacturer into its own field, and removing it from the title is our plan moving forward so thanks for cleaning those up.

As for throttling, we've had some discussions on how to handle the SSME which also is supposed to throttle to something like 104% and then an even newer model could push that to like 109%, but doing so was seldom done due to the fact that it severely increased the chance of engine failure.

Since KSP doesn't have any system for indicating over 100% thrust, and since the only way to handle the higher chance of failure is through TestFlight, I think we had said we were going to make engines in RO stop at 100% and then handle anything above 100% with TF and have it induce higher chance of failure.

However at the time it was only one engine so it wasn't a large concern.  Now with a couple more, maybe we should re-open the topic.

So, pinging @NathanKell for some discussion.

A1Ch1:
Yes, and to make things more confusing throttle levels over 100% don't always indicate a higher rate of failure - in the case of the RS-25, 104% was the "standard" throttle (until Block II or thereabouts, when it was increased to 104.5%) and did not affect failure rates. 109% was the "contingency" throttle, but was actually planned to be used on regular flights when Shuttle was to carry Centaur - it wasn't so much an increased risk of failure as decreased engine life. Later RS-25s could be pushed all the way to 111% for a contingency abort with a somewhat increased failure rate, and certifying them to run at 106% was one of the options explored re: heavy payloads to ISS.

Usually when throttle is indicated to be over 100%, it's because the engine was found to be able to safely supply more thrust than it was originally baselined at (eg the RS-68 operates at 101%), 108% was the standard throttle for the AJ26-62 (yeah, and a few of them exploded - but it seems corrosion from 40 years in storage was the main contributor), The Soyuz 2-1v uses an unmodified NK-33 and runs at 105% as "standard" throttle and is able to run at 108%, so I'd guess 105% doesn't represent a significantly increased failure risk and is just the engine being able to run at higher power than the design goal, while bumping the engine to 108% decreases your chances of going to space today. NK-33 has been tested at up to 130% which would definitely be RUD territory.

So if standard flights use higher throttle, it seems that's the number that should be used: the RS-25 should run at 104 or 104.5% (which I think the RO version already does), and the NK-33 at 105%. Increasing the throttle beyond that, eg to 111% for RS-25 should be for TF, reflecting the increased failure rate at these throttles in the real world.

NathanKell:
I know in the case of the SSME, the over-100% throttling wasn't so much "overboost" as "we improved the engine but don't want to change the avionics." So I do think in that situation (and same for NK-33) we should set maxThrust at the maximum normal thrust. As for chilled-LOX, we already have chilled etc Kerosene (Syntin) so why not.

A1Ch1:
Turns out Energia did use subcooling. 78 K for LOX; LH2 chilled to 18 K to reduce boil-off loss.

PR #218, opened 2015-04-03
Squad and KW H-1/RS-27

A1Ch1:
Used the same engine and ignitor configs on both so weight, thrust, and Isp are consistent. Updated weight of H-1: according to NASA sources the engine weighed around 2000 lbs, Astronautix was the only place that listed it as ~400 kg. And couldn't find any information about H-1B, it seems to only exist on Astronautix. Saturn IB used regular H-1 engines at about 910 kN.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/KWRocketry/RO_KWRocketry_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/KWRocketry/RO_KWRocketry_Engines.cfg
@@ -1054,7 +1054,7 @@
 	}
 	%title = H-1/RS-27 Series [2.5m]
 	%manufacturer = Rocketdyne
-	%description = First stage engine for the Delta II vehicle, burning Kerosene and LOX.
+	%description = The H-1 and RS-27 are first stage gas-generator engines burning LOX/kerosene. The H-1 was used on the Saturn I and Saturn IB launchers and was later modified into the RS-27 for use on the Delta rocket family. The H-1, H-1B, and RS-27 are optimized for use as first stage main engines. The RS-27A, used on the Delta III and most Delta II launchers, is optimized for performance at altitude since these rockets used solid boosters to augment thrust at liftoff.
 	@MODEL,0
 	{
 		%scale = 1.25, 1.25, 1.25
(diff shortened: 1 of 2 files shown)


NathanKell:
We've found the stats [here](http://www.b14643.de/Spacerockets_2/Diverse/U.S._Rocket_engines/engines.htm) to generally be the best, and that definitely shows a difference between the H-1s on Saturn I block I and block II, and Saturn IB.

A1Ch1:
Ah, yes, the engine certainly was uprated, I didn't mean to imply Saturn IB used the same thrust H-1 as the Saturn I. I just meant the NASA documentation refers to the engines on Saturn IB as H-1 and they were rated 910 kN, at sea level :) Now that it's clear H-1 is referring to Saturn I engines and H-1B is referring to Saturn IB engines, I can clean up the configs.

306 s Isp just doesn't look right though. The Saturn 1B Skylab flight manual (http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19740021163.pdf) predicted about 295 s, in line with Astronautix and (https://books.google.com/books?id=l_TLBQAAQBAJ&lpg=SA99-PA1278&ots=pk78UXiepQ&dq=saturn%201b%20specific%20impulse&pg=SA99-PA1278#v=onepage&q=saturn%201b%20specific%20impulse&f=false).

NathanKell:
Ah, apologies! Misunderstood. Yes, by H-1B we mean H-1s on IB, as it were.
I guess that must be a typo--perhaps it's an "off by 10" typo? Let's go with 296 then, and--that doc says 265 for sea level?
Thanks for digging! :)

Also, feel free to stop by #RO on espernet so we can talk in realtime if you prefer--it's usually a pretty active channel.

A1Ch1:
Thanks for inviting!

I think the problem with naming the engine the H-1B is that H-1C and H-1D refer to the engines in the center and the perimeter of the booster, so it's a bit confusing. I was thinking perhaps rename H-1 to "H-1-188klbf" and H-1B to "H-1-205klbf" analogous to the convention used for the J-2 versions. That takes out any ambiguity over which version of the engine is being modeled.

NathanKell:
Hmm, looks it can't be automatically merged?
Also, I think I'd prefer to have them H-1-SaturnI and H-1-SaturnIB, that will be clearer to the people trying to make the replicas. That ok?
(Naming the J-2-200klbf J-2-SLS would be rather confusing, since it's not for *that* SLS)

@jwvanderbeck  would you mind if that change was made to the H-1s? I know it'll take a fix in TestFlight, but a pretty minor one...

A1Ch1:
Sure thing, I wasn't so happy with how the names ended up looking when I booted KSP up either

NathanKell:
Thanks! Assuming you can get the PR such that I can merge it (check the conflicts with current RO master...) I'll merge it--and assuming @jwvanderbeck  doesn't kill us :P

A1Ch1:
OK, bio-major question: where can I find out how to resolve merge conflicts?

NathanKell:
You'll have to either do what you just did, or do it from the command line (the latter isn't hard though). Since you opened the new one, however...

PR #221, opened 2015-04-03
KW LE-7A

A1Ch1:
The LE-7A is throttleable to 72% (Astronautix gives 438 Isp, but all other sources give 440, so went with that)


A1Ch1:
NK happened to approve my last RS-27 pull just as the edit was in progress, so manually undid conflict... hope I did it right!

PR #222, opened 2015-04-04
+part AIES SpaceX Kestrel

A1Ch1:
Duplicate LMAE to create SpaceX Kestrel. Don't think there are any pressure-fed kerosene/LOX engines in RO yet. Model looks very similar to Kestrel, and should be scaled approximately the same. Demo image is using same 1.7 m diameter as Falcon 1.
ro kestrel
kestrel
kestrel diagram

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/AIES/RO_AIES_Engine.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/AIES/RO_AIES_Engine.cfg
@@ -1704,7 +1704,7 @@
 	%node_stack_bottom = 0.0, -0.125, 0.0, 0.0, 1.0, 0.0, 1
 	@title = Kestrel
 	@manufacturer = SpaceX
-	@description = Pressure-fed engine used in the Falcon 1 launcher.
+	@description = Pressure-fed kerosene/LOX engine that was used in the second stage of the Falcon 1 launcher.
 	%mass = 0.052
 	@MODULE[ModuleEngines*]
 	{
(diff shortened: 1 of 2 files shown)


NathanKell:
Good plan!
A problem, though--rescaleFactor is bugged, it isn't safe to use with MODEL nodes. Instead, apply the scaling in the MODEL node and scale the node_ lines to match and leave scale (outside the MODEL node) and rescaleFactor at 1.0.

NathanKell:
Also, noticing this on going back through stuff--can you give actual descriptions for what you do in each commit? Git will already tell us you updated that file--that's the one piece of info we already know from the commit. ;)

A1Ch1:
Argh I actually did scale the thing using MODEL and the nodes, then felt silly for having such an inelegant solution that I went back and rescaled it, thinking the bug had finally been corrected... At any rate still got the numbers used so should be easy fix. Thank you for the tip re:descriptions, I work with cells in a lab all day so this is my first learning experience on anything like this.

NathanKell:
Heh, yeah, it's half-fixed: it applies kinda correctly normally, but still breaks on revert. And yeah, git is a funny thing to get used to, I'm still learning myself, but generally more info = better in most things, right? :)

Also, I'll be away for a bit, but feel free to drop by #RO if you run into any other questions and @Felger, @ferram4, @pjf , @jwvanderbeck  (Agathorn), etc will be happy to help.

A1Ch1:
Didn't even notice that commit descriptions were a thing :) Thought that box was for pull requests

A1Ch1:
Or maybe didn't fix it... if scale has to =1 as well... hmm

jwvanderbeck:
Don't forget you will have to adjust your node locations as well.

NathanKell:
This all good now?

A1Ch1:
screenshot8a
screenshot12a
Added plume configs; should be all done

NathanKell:
Daaang, nice work!

PR #223, opened 2015-04-04
Gimbal range configs for GEM 46 and GEM 60

A1Ch1:
Update to GEM 46 and GEM 60 with new gimbal range configs (thanks NK) for varying mass depending on TVC ability.

TVC-capable SRMs were appended with -TVC, open to alternatives. Couldn't think of a non-clumsy way to indicate TVC-less configs.

Data from (http://www.b14643.de/Spacerockets_2/Diverse/ATK-Thiokol/index.htm) and (https://www.orbitalatk.com/flight-systems/propulsion-systems/GEM-strapon-booster-system/docs/orbital_atk_motor_catalog_(2012).pdf)


NathanKell:
Awesome idea! You know, I didn't even think of this use when I added that to RF...

PR #226, opened 2015-04-07
Remove gimballing on NK-33

A1Ch1:


Phredward:
Does the nk-33 really not gimbal?  http://www.rocket.com/aj26 implies that they are modernizing the gimbaling, but I couldn't find anything specifically about the nk-33 one way or the other.

A1Ch1:
Yes, the original NK-33 didn't gimbal at all, and steering on the N-1 was controlled by differential throttling, with some engines at angles to control roll. The Soyuz 2.1v uses vernier engines for control.

NathanKell:
The original indeed did not gimbal; attitude control on the N1 was done by throttling. However, since (AFAIK) there's no solid, well-used mod to do that, gimbaling seems a fair nod to usability. Thoughts?

Phredward:
My preference would be for an alternative config "NK-33 original (no gimbal").  I agree that for playability no gimbaling makes this engine nigh-unusable when used by itself.  In addition, it seems like it was a very simple manner to add gimbaling to the engine (a technology that totally existed at the time as well), as evidenced by AJ-26.  Perhaps in the future, making an aj-26 variant w/ gimbal and an nk-33 w/o will make sense?

A1Ch1:
Fair enough - the AJ26 config already retained the gimbal at the cost of higher mass, but no gimbal on the original is hard to use without proper verniers

PR #231, opened 2015-04-12
Set propellant to Kerosene for RSSRO configured jet engines

pianojosh:
There is probably a cleaner way to do this but I don't know MM well enough. This fixes the issue I referred to here:
http://www.reddit.com/r/RealSolarSystem/comments/32ao6r/jet_engines_trying_to_consume_liquidfuel_instead/

That I believe is a side-effect of:
https://github.com/NathanKell/AJE-1/commit/230b96aac78f78609c3bd225f5d494166f5204fb

As an aside, if there is a more general way to handle this. I'd love to see it.


NathanKell:
Hah, whoops. Oversight on my part. Thanks for this!

PR #235, opened 2015-04-14
Fixed RS-68 rumble at zero throttle before liftoff

A1Ch1:


NathanKell:
I would suggest 0.001, since there will be times throttle is >0 but <10%. (Since throttle merely lerps between minthrust and maxthrust).

A1Ch1:
Done

PR #252, opened 2015-05-12
LionHead ESA Pack

A1Ch1:
Added config for Vulcain, including reduced O/F ratio so EPC holds correct propellant mass for that engine. Cloned EPC H173 to create EPC 158, with reduced dry mass. Cloned EAP P241 to create P238 with increased dry mass and less fuel. Also reduced length of EPC core so boosters interface at correct location above propellant tanks, modified thrust profile of EAP to simulate burnout and modified dry mass where appropriate.

Simulation using added Ariane 5 G parts for single-burn GTO mission: https://youtu.be/M3M6C624CIk


A1Ch1:
Will modify as needed to comply with KSP 1.0 changes, still figuring out exactly what is required

A1Ch1:
Waiting until 1.0 update to add Ariane 6 + Vega configs

PR #256, opened 2015-05-17
RN_Skylab Update

raidernick:
Remove now defunct skylab_unbroken
Subtract 3 from the mass of skylab as the telescope is now a separate part which weighs 3 tonnes
Skylab should "technically" be able to hold 6 people, the IVA supports this many, and there are 2 docking ports even though 1 was unused
Added definition for telescope solar panel


NathanKell:
@raidernick , @jwvanderbeck , this needs to be targeted to branch 'KSP10' not 'master' ....

PR #260, opened 2015-05-20
Remove references to old FASA parts

Wetmelon:


NathanKell:
@Wetmelon it would be better to create the parts as new parts (copy the old cfg and then make the specified changes) since the assets are still in FASA, just not the cfgs. That means RedAV8R's careful FASA fairing creations can still be used....

PR #299, opened 2015-05-24
Update KWRocketry.cfg

John-Long-Silver:
Changed "emission = 0.0 x" to "emission = 0.0 0" and "directThrottleEffectName" to "powerEffectName" as suggested by NathanKell.
I hope this is in any way helpful.

### DIFF
--- a/GameData/RealismOverhaul/SmokeScreen_EffectCfgs/KWRocketry.cfg
+++ b/GameData/RealismOverhaul/SmokeScreen_EffectCfgs/KWRocketry.cfg
@@ -11,7 +11,7 @@
 				name = flamethrust
 				modelName = RealismOverhaul/SmokeScreen_MP_Nazari_FX/KWbooster
 				transformName = NozzleTransform
-				emission = 0.0 1.75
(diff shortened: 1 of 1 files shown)


John-Long-Silver:
Is this better now?

NathanKell:
Yeah, looks good. However, last night @stratochief66  found that these changes _may_ not be necessary after all, so I'll try with and without them and see what happens, and if they're not needed I'll strike it from the todo list. Meanwhile I'd ask if you might look into the other bits on the todo list.

Thanks! :)

John-Long-Silver:
Ok, so let me know if you still need them, in that case I will create a PR for the other files.
Regarding the rest of the todo list I might need some further instructions.

NathanKell:
Sure! Can you pop by #RO on espernet? ( [webchat link if you don't have an IRC client](http://webchat.esper.net/?channels=RO) )
If not can do it here.

NathanKell:
@John-Long-Silver  ?

John-Long-Silver:
Let's do this here for now...

PR #310, opened 2015-05-24
ModuleJettison and Node_stack removed

Replica17:


NathanKell:
This is actually fine as it is; node_stack_bottom should stay. I was talking about nodes that RO adds, rather than changes.

PR #320, opened 2015-05-27
Added Realplume configs for KW and AIES

Replica17:
Tested ingame 1.0.2

Working:
All AEIS working. Alignment is correct.
All KW working. 

Added Procedural Parts.

To do: Different size of realplume for different SRB.


NathanKell:
@Replica17  as for ideas, verify that the transform names in the EFFECTS stuff (like `thrustTransform` ) matches the ones in the KW configs. I note that KW engines appear to already have EFFECTS nodes, so you can grab the names from them (seems like multiple transforms?). Also, make sure you add
!EFFECTS {}
on the line above your
EFFECTS

to clear out the old EFFECTS node.

PR #324, opened 2015-05-30
NearFutureSolar parts names changed.

FennexFox:
Nertea changed the names of his parts in more organized way. So do I.


NathanKell:
This PR appears to include both that small change to NFS parts, but also all the DCSS work. Intended?

FennexFox:
Wow, didn't noticed that. I must read the manual of GitHub first.

PR #325, opened 2015-05-30
RealPlume configs for the rest of NP boosters

John-Long-Silver:
Did all the engines in RO_NovaPunch_SolidBoosters.cfg
I messed around a bit with the parameters, so please check, whether you like the result.

### DIFF
--- a/GameData/RealismOverhaul/SmokeScreen_EffectCfgs/NovaPunch.cfg
+++ b/GameData/RealismOverhaul/SmokeScreen_EffectCfgs/NovaPunch.cfg
@@ -708,6 +708,212 @@
 		@gimbalTransformName = thrustTransform
 	}
 }
+@PART[NP_srb_radialbooster]:FOR[RealPlume]:NEEDS[SmokeScreen] // Zenit Energia Liquid Booster, scaled up from RD-107 above.
+{
+	!fx_exhaustFlame_blue = DELETE
+	!fx_smokeTrail_light = DELETE
+	!fx_exhaustLight_blue = DELETE
+	!fx_exhaustSparks_flameout = DELETE
+	!sound_vent_medium = DELETE
+	!sound_rocket_hard = DELETE
+	!sound_vent_soft = DELETE
+	!sound_explosion_low = DELETE
+	!EFFECTS {}
+	EFFECTS
+	{
+		powerflame // name of effect to be added to the Engine Module
+		{
+			MODEL_MULTI_PARTICLE_PERSIST
+			{
+				name = flamethrust
+				modelName = RealismOverhaul/SmokeScreen_MP_Nazari_FX/KWflamelarge
+				transformName = thrustTransform
+				localRotation = 0,0,0
+				localPosition = 0,0,0.4
+				offsetDirection = 0,0,1
+				fixedScale = .75
+				sizeClamp = 50
+				initialDensity = 0.8
+				physical = False
+				stickiness = 0.9
+				dragCoefficient = 0.1
+				singleEmitTimer = 0
+				randomInitalVelocityOffsetMaxRadius = 0
+				logGrow
+				{
+					density = 1.0 10
+					density = 0.5 10
+					density = 0.0 10
+				}
+				logGrowScale
+				{
+					density = 1.0 0.5
+					density = 0.79 2
+					density = 0.0 3
+				}
+				linGrow
+				{
+					density = 1.0 -0.1
+					density = 0.79 0.1
+					density = 0.005 0.1
+					density = 0.0 0.5
+				}
+				speed
+				{
+					density = 1.0 3
+					density = 0.79 3
+					density = 0.005 3
+					density = 0.0 3
+				}
+				emission
+				{
+					density = 1.0 3.0
+					density = 0.79 1.0
+					density = 0.5 0.5
+					density = 0.0 0.25
+				}
+				energy
+				{
+					density = 1.0 1.2
+					density = 0.5 1.6
+					density = 0.1 1.7
+					density = 0.0 1.9
+				}
+				offset
+				{
+					density = 1.0 0.083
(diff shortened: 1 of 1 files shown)


NathanKell:
Sorry, these mostly seem to be taking as their model configs, engines with completely different area ratios--did you modify the density-based plume changes for the new area ratios?

(Also, sorry, was away for weekend and slow getting back to things...)

John-Long-Silver:
I do think so, however, as I coudn't find a documentation on how exacty the parameters are being used by RealPlume, I had to revert to guessing in many instances. If you could show me something that enables me to recalculate that stuff I'll be more than happy to do so.
Currently there is no version of  Suttons "Rocket Propulsion Elements" available at our university library that I'm allowed to take home, so I admittedly used rather incomplete Information both on the theoretical side, as well as on the implementation side/the formulas that RealPlume uses. Therefore I mostly reverse engineered from stuff like: Is the approximate height of flow separation right, etc.

PS: Replica17 used some of my configurations in the KW files. If something is wrong here, we will have to check those as well.

NathanKell:
I see. The best docs we have are https://drive.google.com/file/d/0B8d75CnONrbmdlh0bkZ6MEdwREE/view?usp=sharing and the Mainsail example (.bak) in the SmokeScreen folder. Beyond that ask Felger or ferram4 in chat. I'm sorry to say this is _not_ an area of expertise for me. I'll test these out with NovaPunch, and apologies for missing where you #*(%@ _explicitly stated_ you modified the values, in the PR itself. Oy. Sorry.

John-Long-Silver:
Thanks, but this file is only helpful regarding the looks of certain flames, not for the relation between physical parameters and game parameters.
I know the Mainsail example, but it doesn't tell me what will happen for non-positive values of the log-base...
Anyways I will try to contact Felger or ferram4 some time soon, however I won't be available for the next few days. So I'm sorry if this causes delays.

Felger:
FYI - if you type @Felger, it emails me and tells me I was referenced in a conversation.  Kind of like this: @John-Long-Silver 

On the subject of plume configs, I've been thinking about that myself recently, I wonder if it wouldn't be worth writing a tool that would take engine parameters and generate a set of numerically correct plume configurations. @ferram4 can comment more on how possible this is than I can, I'm only a little familiar with the math.

I'd expect you'd have to input nozzle ratio, isp (to determine exhaust velocity) and fuel density, perhaps a few other parameters.

Once you had that set of configs, you could take it in-game, size the initial plume effect, and adjust the offset by density to line it up with the engine bell itself.

As an aside, @sarbian, is it possible to stretch the particles horizontally only? (perpendicular to the flow?) The smallest plume you can make is effectively just a sphere, which doesn't look quite right for vacuum configs.

sarbian:
@Felger try adding renderMode = "Stretch" to the effect. It s been a long time since I works with SmokeScreen...
I also see that there is a lengthScale parameter in the particle system that I do not use. I'll add a new curve to expose it tonight

And I need to start my work on 3D plumes...

John-Long-Silver:
@Felger (I like how this @ works): Generating config files automatically sounds like an awesome idea. However for more complicated fuels and especially solid boosters the exact mixture of gases in the exhaust might add some uncertainties and could also limit the applicability for different ratios of fuels. However, this is not at all my field of expertise, but @ferram4 I would absolutely love to learn about physics of plumes, any ideas where I can start? Or should I just try to derive it from what I remember from thermodynamics?

Felger:
So, doing a little research on the subject, here's my thoughts on a good growth config:
logGrow dominates in the atmosphere, due to mixing with the atmosphere near the plume constricting it.

Outside the atmosphere, or very high, logGrow tapers off, replaced by linGrow, since there's no atmosphere to push the plume back in and come to an equilibrium with the plume.

To simulate the quick dispersion of the plume to vacuum, I often set emission very low, and energy very low at vacuum, but be careful with this as it can look a little funny.  Can try selecting different effect models.

For reference (if you're not already aware) I uploaded screenshots of the [MP_Nazari models](https://drive.google.com/file/d/0B8d75CnONrbmdlh0bkZ6MEdwREE/view?usp=sharing) to google drive, so you can use those as a reference.

Felger:
Okay, so I got a little obsessed, made a spreadsheet with assumptions that would probably make @ferram4 cringe.  It's probably not the worst first start though.

[Have a look](https://docs.google.com/spreadsheets/d/1b4Y3THZCTXYfUvsU-s9LRIiYkg2Gk9L4kdIh_FTwNuA/edit?usp=sharing)

John-Long-Silver:
You are aware, that you created a simple 1D harmonic oscillator (\frac{d^2 R}{d x^2}=- const. * R)? I don't quite understand why you solve the second most trivial of all differential equations numerically :P Anyways, apart from numerical errors your result is an undamped oscillation.

Felger:
Hahaha, yeah, that's what I get for doing it quick and dirty without any paper.  The rest of the math should be a good 0th order approximation though.  Replace the crappy linear integrator with a proper solution and we're good to go.  I also probably screwed up the expansion math with my pressure assumption.

In any case, I opened up the doc to edits by anyone with the link, so feel free to tinker with it, or do whatever you want.  I wonder if we could pipe a similar thing into SmokeScreen for live manipulation of the plume based on the atmospheric condition.

NathanKell:
So, the conversation having left the pull, I think it's best I merge this in and y'all continue either here or on a new issue?

John-Long-Silver:
@Felger I was thinking that maybe one can use this new transsonic cross-section that ferram introduced apparently in "Ferri" to improve upon the undamped situation by giving the surounding flow some a) pressure profile and b) rigidity due to it's wish to minimize the surface for laminary flows. Mixing and eventually loss of laminar flow (roughening transition) should then be managed by a different plume.
I have no problem with moving this discussion, but right now I'm unaware of any better place...

PR #327, opened 2015-05-31
Fasa newbits1

stratochief66:
Adding configs for FASA Castor and MB3 Delta engine.


stratochief66:
Unsure why those other commits are showing here, any ideas? Will that cause trouble?

NathanKell:
Probably because you took a merge from master at some point?

PR #363, opened 2015-06-18
Update RO_FASA_LEM.cfg

MartinFleming:
At the moment the LEM has an offset center of mass created by the landing leg with ladder part, which weirdly seems to have much more mass than stated in the configs. This makes it impossible to fly as is.
To counter this I have reduced the mass of the part and the LEM shelf to 0.005, there landing on the Moon is much easier :)

### DIFF
--- a/RealismOverhaul/RO_FASA_LEM.cfg
+++ b/RealismOverhaul/RO_FASA_LEM.cfg
@@ -544,7 +544,7 @@
 	@node_stack_top = 0.0, 0.182, 0.4, 0.0, 1.0, 0.0, 0
 	@title = Apollo Lunar Module Leg - Ladder
 	@description = The landing leg with ladder.
-	@mass = 0.1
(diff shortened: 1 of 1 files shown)


NathanKell:
@MartinFleming  can you rebase against the current master? I can't auto-merge it.
Also, what do you mean by "has more mass than stated"? If you add an octo2 core (50kg) and then add a ladder to it, what does Mechjeb say the mass is?

stratochief66:
It is ok for the LEM to have a slightly off centre CoM, I believe. The descent stage has a gimballing engine that can take care of that (to direct the thrust through the true CoM) and the ascent stage has plenty of RCS to attain and hold your chosen orientation.

@NathanKell, can you test (or have someone test) that the current FASA RO config or FASA Saturn V craft LEM has these properties? They work for me, but I made them.

PR #364, opened 2015-06-20
Correcting Plume Position for RD-191

JandCandO:
I'm still a bit confused about the realplume prefab system. Is this the proper way to correct an engine plume if the auto-assigned plume is a bit off?


Felger:
Fix it in the PLUME node if the part has one. It'll be in the smokescreen
configs folder. If you're happy with the effect, you could adjust it and
submit a pull to make our effects better.

JandCandO:
So since this pull request is creating the plume node, I'm pretty much good to go?

JandCandO:
I'm not going to make any more changes to this, it is fine for now.

PR #372, opened 2015-06-24
enlarge science storageRange for mk1landerCan und mk2landerCan

glilienthal:
Otherwise it was only possible from the top of the can to store data in it or retrieve data. 
The external access to making crew reports might still be screwed, but that should be done internally anyway.


NathanKell:
Thanks! :)

PR #381, opened 2015-06-28
Adding 1.04 Heatshields for Gemini & Apollo

stratochief66:
A bit crude, but quite functional.

Both can survive lunar velocity re-entry.

Welded the heathshield onto the Apollo capsule as otherwise the capsule
kept exploding while in angled re-entry mode.

A bit different from each other. Apollo hits the desired high
temperature regime but doesn't burn off much Ablative. Gemini doesn't
reach as high a temp but loses much more ablative.


stratochief66:
Someone better at thermals is more than welcome to tweak these further, my work feels crude.

Ideally, the Gemini will be able to survive lunar speed re-entry but nearly burn through the Ablative and reach the maximum skin temperature (difficult, but possible)

The Apollo should burn some Ablative (20kg?) and approach 2800 - 3000 degree skin temp. Less marginal re-entry, much more easily survivable in Apollo

PR #384, opened 2015-07-01
Updated forum link to Procedural Parts thread

hoorayimhelping:
Currently, the link points to an older forum thread. A new thread was created and this doc was never updated.


Felger:
Thanks for the catch!

For future reference, KSP forums only needs the link up to the thread number (106975 in this case).

PR #394, opened 2015-07-05
Bobcat N1 Overhaul

stratochief66:
Abides by new LS standard.

Includes a few simplifications. For instance, Soyuz LOK main engine is
modelled as RCS, as the normal engine would ignite on the launch pad.

Most importantly, this version can actually perform the landing and
return safely!

May need a heatshield rebuild to work with RealHeat release.

Craft file uses the parachute as the base part, or else the main capsule
shrinks on vessel reload.


NathanKell:
You can fix the shrinking by using scale inside the MODEL node, rather than a non-1.0 rescaleFactor.

PR #396, opened 2015-07-08
Update RO_FASA_LEM.cfg

smartdummies:
Change Node attach to allow the landing legs to attach to the Lunar Descent Module


stratochief66:
Good catch!

PR #397, opened 2015-07-08
Update RO_FASA_ApolloCSM.cfg

smartdummies:
Changed node attach for Apollo SM to allow the SM antenna to attach correctly


stratochief66:
Again, good catch!

PR #399, opened 2015-07-09
raidernick v1 rescale

raidernick:
Fixed scale to be correctly 1:1, the original blender scale was off when the model was made.


raidernick:
fix name

PR #400, opened 2015-07-09
Raidernick scout update

raidernick:
Fixed rcs issues, fixed drawgauge


NathanKell:
Two %key in succession will fail. Instead use
@key,0 = 0 x
and
@key,1 = 1 y
(where x and y are the desired Isps).
This is because @value or %value will always modify the first found value unless told otherwise, so %key followed by %key will just change the first key twice.

GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_USRockets.cfg
```diff
@@ -308,21 +482,21 @@
...
-			ratio = 1.0
+			%name = HTP
+			%ratio = 1.0
 		}
-		atmosphereCurve
+		%atmosphereCurve
 		{
-			key = 0 167
-			key = 1 57
+			@key = 0 167
```
NathanKell: Needs to be @key,0 = 0 167

GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_USRockets.cfg
```diff
@@ -308,21 +482,21 @@
...
+			@key = 1 57
```
NathanKell: needs to be @key,1 = 1 57

PR #401, opened 2015-07-10
Fix Non-RO parts appearing in fuel tanks category

OtherBarry:
Tested it, works on mine.


OtherBarry:
Probably worth someone else checking, just to be safe.

PR #412, opened 2015-07-15
luna update 2

raidernick:
added luna 11 config


NathanKell:
DO NOT COMMIT UNTIL @raidernick SAYS SO.

PR #426, opened 2015-07-19
Update Soviet Engine Realplumes

JandCandO:
They now use the plume|flareScale and plume|flarePosition features from RealPlume v0.4.1.


Felger:
Looks good! Merging

PR #430, opened 2015-07-21
RN proton sep temp fix

raidernick:
Temporary fix for rp sep motors


NathanKell:
Tagging @Felger so this gets undone when SmokeScreen updates.

PR #431, opened 2015-07-21
Add TRANSMITTER to passive antennas

batuhantasli:


NathanKell:
@batuhantasli  Great, thanks! Can you do a quick check of other files too? It may be more widespread.

PR #467, opened 2015-07-31
RN remove molniya rcs fuel

raidernick:
fix molniya, it has no rcs thrust transforms it uses RW, so the rcs fuel
and module are useless


stratochief66:
does the historical hardware have a reaction wheel??

NathanKell:
I take it astronautix is wrong about it then? (It mentions 78kgf-sec of RCS impulse)

PR #472, opened 2015-08-06
Added Cygnus/e mass values.

advicedawgg:
Mass was previously 4kg dry weight for both Cygnus spacecraft. Added real values.


advicedawgg:
Engine configs responsible for mass error. Disregard this pull

PR #473, opened 2015-08-06
Cygnus/Cygnus_E mass value correction

advicedawgg:
Prior mass values were 4kg! The weight of one BT-4 Engine. This was a copy from the moduleengineconfigs. This has been corrected to the true mass of these spacecraft.


stratochief66:
Interesting catch. In  a case like this, I am not sure if having origMass even serves a purpose.

@advicedawgg Have you checked this change in game to make sure it doesn't double the mass of the part (1.45 + 1.45) or any other unexpected behaviour?

advicedawgg:
@stratochief66 Looks like there is no mass duplication, the part comes out correct. Though origMass can safely be removed entirely.

PR #477, opened 2015-08-09
RN fixed R7 tanks and engines

raidernick:
Made the R7 tank and engines values realistic. By fixing the engines I
was forced to also redo the tanks. I also redid all the craft files and
subassemblies.

So far everything looks great. I tested sputnik 1 and Luna 2. They
worked. I didn't test the rest but all the ASL and VAC DV values are the
same or higher than before for all rockets. That should mean they should
all be working.


raidernick:
The main issue was the soyuz spacecraft and rockets not making it to space without using almost all of the soyuz fuel, which is not realistic. With the proper engine values the craft has about 600 more DV now than previously and that "should" work considering that is more than the entire spacecraft holds, you should actually end up with some extra on a perfect ascent.

raidernick:
Also, the weight of the interstages and engines wasn't subtracted from the weight of the core stage. The old weights matched the TOTAL mass of the core stage with everything on it.

PR #482, opened 2015-08-15
LR-89 Ignition fix

stratochief66:
Some smart person(s) pointed out earlier in the week that this engine was failing to light on the pad. This appears to fix that.


gruneisen:
I think the same bug also affects the FASA LR89 engine. Should this change extend to those configs too?

PR #490, opened 2015-08-30
Rename parts post-RLA update. Reassign some engines

A1Ch1:
RLA has been updated. Almost all parts were renamed and some were removed. Fixed part names for engines that are still in pack and reassigned some engines to different real counterparts.


NathanKell:
Awesome!

PR #500, opened 2015-09-06
Added Procedural Crew Tubes

codepoetpbowden:
This pull request adds the concept of "procedural crew tubes" for Realism Overhaul. crew tubes are structural parts that allow crew to pass through them from a CLS point of view. They are more massive and costly than regular structural parts, and have a minimum size. They add an extra realism consideration to station and base building.
I am not terribly experienced with module manager, but have tested this in both career and sandbox modes and it seems to do what I intended. It is humbly offered to the community, I hope it is seen as an  improvement.

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
@@ -51,6 +51,7 @@
 		}
 	}
 }
+
 @PART[proceduralTankRealFuels]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
@@ -65,6 +66,7 @@
 		typeAvailable = LifeSupport
 	}
 }
+
 @PART[proceduralTankKethane]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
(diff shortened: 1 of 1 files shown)


Ravenchant:
Good idea! I think the 2m minimum diameter is a little on the large side though.

codepoetpbowden:
Yep, I suppose it depends if you are sizing for kerbals or humans to be able to move around inside.
I went for 2m because that is the diameter for the standard size docking ports in RO, so a docking port fits snuggly on the end of one of these tubes, and I assumed that there needed to be some width allowed for sheilding, blankets etc that distinguish these parts from normal structural parts. 
However I am not precious,and was only offering it up - change it as you see fit.

NathanKell:
Looks good to me!
However, could you do the clone on :FISRT (or at least FOR[RealismOverhaul] )? If you clone a part on final, nothing can change it.

codepoetpbowden:
I played around with NEEDS and FOR, but could not figure out how to make it work. I will have another go when I get home tomorrow.

NathanKell:
@codepoetpbowden sent you a PR explaining what I meant; in doing it I realized the original thing I was saying wouldn't quite work, but I think what I sent does.

codepoetpbowden:
Made another change to mark the crew tube parts as RO supported ones. 
Now all tested and ready to be pulled.

NathanKell:
Great! :)

PR #507, opened 2015-09-09
RemoteTech_Settings: Set HideGroundStationsBehindBody = True

pjf:
This setting means that ground stations on the other side of the Earth
aren't visible *through* the Earth.

Tested on RemoteTech v1.6.7, RSS v10.2.0, KSP 1.0.4


NathanKell:
Oooh! Right, I forgot that got added!

PR #538, opened 2015-09-22
Hatch added to Conic Cockpit

Randazzo:
Updated model with hatch and ladder


NathanKell:
Looks like this still has the merge markers in the file. Can you remove them and leave only the correct lines? I *think* they'll be the ones marked HEAD in either case, since that's from your side. The RO/master ones are the lines in the original cfg you uploaded.

Randazzo:
Should be corrected now. Not sure how I went about fudging it in the first place.

Issue #519, opened 2015-09-14
RT config should add groundsites iff RSS is installed.


NathanKell:
.

NathanKell:
Verifying I can do this via https://github.com/RemoteTechnologiesGroup/RemoteTech/issues/490

PR #540, opened 2015-09-23
RemoteTech_Settings update

Peppie23:
I've removed all ground stations because they will be shiped with the RSS mod (https://github.com/KSP-RO/RealSolarSystem/pull/35) in the feature. i've also removed all values that should not be overwritten. The module manager NEEDS-block should not be required, i only want to be safe.

Ref Task: https://github.com/RemoteTechnologiesGroup/RemoteTech/issues/490
Closes https://github.com/KSP-RO/RealismOverhaul/issues/519


Peppie23:
It would be nice if someone could cross check this change, and also 
https://github.com/KSP-RO/RealSolarSystem/pull/35

NathanKell:
No problem! I'll merge manually, and thanks again ever so! :)

PR #545, opened 2015-09-26
Add thrustCurve for AJ-60A

A1Ch1:
Not totally precise, estimated based on 

rwv atlas 5 521 accel

from Atlas user guide. In game acceleration:

rlv atlas 5 521 accel


B787-300:
The boosters burn out at 95 s or so into the flight of Atlas, which is the first time it peaks over 2g acceleration.  Plus that graph includes the RD180 firing at the same time.  The dip at about 15 sec seems to be throttling down for maxQ.

A1Ch1:
The dip at 20 s definitely isn't the RD-180 throttling down, it throttles down to 92% at T+35, then back up to 100% at T+65. Unlike Orbital ATK, AR hasn't made the booster thrust curve public, so the total acceleration is the best we've got. The boosters definitely reduce thrust starting at around T+20, the second dip is the RD-180 throttling down, and the bump at T+50 is the boosters increasing thrust.

B787-300:
I agree with that and from what AJ says on their website, http://www.rocket.com/atlas-v-solid-rocket-motor , but you seem to have the full curve in there for the entire rocket applying to the booster.

A1Ch1:
Yes, the total acceleration (2 boosters + core) is the only information available. It would certainly make things easier to have the booster thrust itself.

B787-300:
let me see if i can get it.  I have to email AJ-R asking about other motors for a Senior Design Project, so I might be able to coax it out of them.  But this is much better than having the motors putting out 100% thrust for the entire burn time. So I thank you.

A1Ch1:
Yes, the lack of an AJ-60A thrust curve has been bugging me for a long time. Many kudos if you can get that information!

A1Ch1:
The more I play with the thrust curves, the less I trust the RD-180 throttle levels cited in the Atlas payload guide. Although it specifies 92% thrust from 35 s to 65 s for a 521, the lack of any bump in acceleration at 65 s, and the giant spike at 50 s, makes it seem more likely there's a deeper throttle that ends earlier.

PR #548, opened 2015-09-26
Changed stats of dish antennas.

Schnobs:
All dishes have a range of either 250Gm (inner planets) or 40Tm (well past Pluto), but vary by weight and power demand.

Antenna descriptions now indicate their useful range in the game.


A1Ch1:
This looks great! Dishes have needed some TLC for a while

NathanKell:
We discussed this in the channel regarding more range bands etc., so temporarily closing this pending a bit of a rework (I believe that was the upshot of the conversation?)

PR #553, opened 2015-09-30
RO compatibility for the stock radiators.

PhineasFreak:
Most of the data for the radiator wings came from the "ISS ATCS Overview" pdf file.


NathanKell:
Awesome, thanks!

PR #555, opened 2015-09-30
Update RO_AJE.cfg

RedAV8R:
Some better naming and description changes.

Give credit to mod (ex. AJE) as 'manufacturer' per agreement with others.

### DIFF
--- a/GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
+++ b/GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
@@ -1,6 +1,8 @@
 @PART[aje_f404]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
+	%title = General Electric F404-GE-402
+	%manufacturer = AJE (Squad)
 	@MODULE[ModuleEngines*]
 	{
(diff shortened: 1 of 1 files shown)


NathanKell:
So, after discussing it with part makers, and on the channel, we had a reevaluation--we do, now, use the correct manufacturer name for the given item. So that's one less complication! :)
(That's why there have been some patches to change manufacturer name)

RedAV8R:
Ok, that's fair.

I still believe some naming of the 'title' should be a bit more descriptive, especially for stock items. Like "Fuel Tank - 1.25m", or something like that. Something a bit more standardized, and easy to located. Will also group a whole lot nicer when sorted by part name, etc.

NathanKell:
Ah, sure, I totally agree! That makes good sense.

RedAV8R:
BTW.. I"M BACK!!! :)

RedAV8R:
Leave me be, no need to be a big part, but i'll be attempting to clean some stuff up and make PRs as I go.

NathanKell:
And a hearty welcome back to you! :)

Hope things are well Chez Red, and all! :)

NathanKell:
Will do. Always awesome, always appreciated. :)

Issue #63, opened 2014-12-02
Realchute stack "hovers" above other parts.


pjf:
Actual behaviour: The Realchute stack (`RC_stack` in `RealismOverhaul/RO_RealChute.cfg`) has top and bottom nodes which aren't on the surface of the compoent itself, causing it to 'hover' when placed in an actual stack.

Expected behaviour: Attachment nodes are on the surface of the part itself.

Steps to reproduce: Select the RC_stack part in the VAB and attach it to a part. (I'm using just an okto2 and the stack chute, and there's a very noticable gap.)

Note: I haven't tried RealChute without RO, so this could be stock behaviour.

Mod loadout:

```
KSP found at /home/pjf/.steam/steam/SteamApps/common/Kerbal Space Program

KSP Version: 0.25.0

Installed Modules:

✓ AGExt 1.24a
✓ AJE 1.6.4
✓ AlternateResourcePanel 2.6.1.0
✓ Chatterer 0.7.1
✓ CIT-Util 1.0.4-unofficial
✓ CommunityResourcePack 0.2.3
✓ CommunityTechTree 1.1
✓ CrossFeedEnabler v3.1
✓ CustomBiomes 1.6.8
✓ CustomBiomes-Data-RSS v8.3
✓ DDSLoader 1.7.0.0
✓ DeadlyReentry v6.2.1
✓ DMagicOrbitalScience 0.9.0.2
✓ DogeCoinFlag 1.02
✓ EngineIgnitor-Unofficial-Repack 3.4.1.1
✓ EVE-Overhaul-Core 0.0.2014.11.25
✓ FerramAerospaceResearch v0.14.4
✓ FinalFrontier 0.5.9-177
✓ FinePrint 0.59
✓ FirespitterCore 7.0.5398.27328
✓ HotRockets 7.25
✓ InfernalRobotics 0.19.2
✓ Karbonite 0.4.4
✓ KAS 0.4.9
✓ KerbalAlarmClock v3.0.5.0
✓ KerbalConstructionTime 1.0.3
✓ KerbalJointReinforcement v2.4.4
✓ kOS 0.15.3.0
✓ KronalVesselViewer 0.0.4_0.25
✓ MechJeb2 2.4.0
✓ ModuleManager 2.5.1
✓ ModuleRCSFX v3.3
✓ NathanKell-RVE-Haxx 0.0.2014.11.25
✓ ORSX 0.1.3
✓ PartCatalog 3.0_RC8
[...]

PR #556, opened 2015-09-30
Renaming for standardization/commonality - Other Fixes

RedAV8R:
Title only. One description changed for added information.


NathanKell:
Um, with the Manufacturer property available why are you also adding it to the title? I thought the idea was that we'd shift Manufacturer name to the manufacturer field, and then have just the actual designation in the title...

Sorry if I missed something...

RedAV8R:
No...didn't miss anything.

Half or more had the manufacturer in the title, half didn't. I went the way that seemed more descriptive. Manufacturer and model.

So...

Sure the manufacturer is now in the manufacturer field. In KSP basically allowed searching via manufacturer in the parts list, if a couple other things were properly set up. With FilterExtensions, that's kinda removed to a more 'per mod' basis. Which I like. Without FE, nobody could search via manufacturer. So with title including it, things were more or less sorted by manufacturer.

So. How exactly do we want things. A small title of "F404-GE-429" isn't much descriptive, and kinda wastes space in a way. And So. Surely there is something else that can be put there. Perhaps every engine could have a prefix of "TurboJet:", "TurboFan:", "Rocket:"...or something to that effect. Even go as far as put air intakes (at least dedicated ones) in the engine category with a "Air Intake:" prefix or something.

RedAV8R:
Any thoughts?

A1Ch1:
I rather like the convention of using manufacturer names in the titles of jet engines, while leaving them out of the titles of rocket engines, as this mirrors Wikipedia Aviation and Spaceflight article titles.

RedAV8R:
While not always completely true with Wiki. I personally rather like having the manufacturer in the title regardless of it's type.

I'd be ok leaving it out with rocket engines, while keeping it for air breathing cousins. As rocket engine manufacturers have been a lot more 'dynamic' Aerojet, Rocketdyne, then Aerojet Rocketdyne, etc. Which of the three should be put down?

What it boils down to for me, is I'd like to see better standardized naming so that searing for a specific part isn't more of a crap shoot like it is now. I have no issues working on that. Most of the work is done, just need to form PRs to get it where they need to be. At least when it comes to the more generic/stock items.

NathanKell:
Sorry, was a bit busy last night and today.

The way I see it, if we include a manufacturer prefix, then engines are always sorted by manufacturer; if we don't, you can sort by title alone (i.e. roughly sorting by year, for US J/F/LR or Russian RD), or by manufacturer (via Filter Extensions, though it might be worth asking how much need be done to allow non-stock manufacturers to be handled stock).

Regarding rocket engines, the consensus that was arrived at about a week ago when this came up (on IRC rather than git) was:
title is just the designations (LR91 Series, AJ10-137 Service Propulsion System, etc)
manufacturer is as it says on the tin
description is prefixed with a short role statement (i.e. Vacuum optimized, hypergolic, deeply throttleable, low TWR, mid 1960s) followed by a newline (\n) and then the long-form description.

This serves the purposes of:

1. Not stating the role in the title (some existing engines say Booster or Sustainer or Upper....); we don't want to prejudge a player's use.
2. Nonetheless we do want to provide some easy-access use-relevant data to aid in the selection of engines by those unfamiliar with rocket history.

Regarding jets, at least in the US case, turbofans and turbojets will already sort differently so a prefix isn't needed, though I can see the merit in adding a postfix.

Getting rather sleepy so I'll end there, I'll be cogitating more tomorrow...

Tagging some others involved in the aforementioned discussion (sorry if I missed anyone)
@eggrobin @ferram4 @saabstory88 @pjf

RedAV8R:
I do like the agreement with rocket engines. I really do.

Not seeing how air-breathers will sort differently, other than that basically won't. But I'm sure ideas have been passed that will. We'll wait for ya sometime today:)

RedAV8R:
Changed engine titles to model number/designation plus suffix.
Also Fixed issue #63

NathanKell:
Bleh, sorry, KSP bugfixing is taking *way* more of my time and mental energy than expected.

I agree about the diameters, also for the same reasoning as above (RO is supposed to *break* one of the habit of using engines of the "right size"). For jets, I meant that US jet designations (and Russian RD designations, for those jets which used them) are roughly chronological, in that J34 comes before J57 comes before J79. They will also sort out into jet (J), turbofan (TF, later F)--that's what I meant about sorting differently, they're already segregated.

NathanKell:
(And let me know if I should merge this in, or await further commits)

RedAV8R:
Figure as much about the engines.

Since it'll be a long process getting names changed. Do we want to just do individual PRs as they are worked on, or just do one massive PR when it's done?

If individual, let me do a few more changes on the files currently committed and I'd go ahead and merge it. I'll post again when ready.

RedAV8R:
Good to go. Merge at your convenience.

NathanKell:
Cool, doing so!

Btw, @SirKeplan started a wiki entry on this stuff. https://github.com/KSP-RO/RealismOverhaul/wiki/Part-Naming

PR #563, opened 2015-10-02
SXT Kossak cockpit

winged7:
This one: https://www.youtube.com/watch?v=yca5tnt5Zo8

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Command.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Command.cfg
@@ -267,7 +267,7 @@
 	@node_stack_bottom = 0.0, -1.93, 0.0, 0, -1, 0, 2
 	@MODEL
 	{
-		%scale = 1.52, 1.52, 1.52
+		%scale = 1.52, 1.52, 1.625
 	}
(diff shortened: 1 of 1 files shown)


NathanKell:
I know the Osaul bits are supposed to be An 225; do we know from Lack what this is based off? His parts are usually realworld clones, and so should get realworld stats...

winged7:
Cockpit is very similar to B-737, also SXT fuselage is supposed to be B-737 or A320 fuselage: http://i.imgur.com/YBS3EbG.jpg 
SXT fuselage is 3,75m x 3,75m - the same as Kossak after resizing.

Real B-737 fuselage dimensions are : 4,01m x 3,76m  http://i.imgur.com/9gcRXbD.png

I think that 3,75m x 3,75m is better because 3,76m x 4m will make them incopatible with Procedural Tanks ( You can't get such irregural shape)

NathanKell:
Given that we have the fuselage parts in #567 - and I think there's an aircraft tail in 2.5m floating around SXT as well--then it might well make sense to use the real dimensions, especially since IIRC there's a mod-of-a-mod for Proc Parts that does allow ellipsoid proc parts.

Anyone else want to chime in on this?

winged7:
Size updated to 3,76m x 4,01m

Could you give me a link for that mod (ellipsoid proc parts)?

PR #565, opened 2015-10-03
PF transparency, CMES updates

PhineasFreak:
Added the ability for the Procedural Fairings to become transparent when hovering above them in the VAB (like the stock ones, although it's not perfect).

Updated the CMES MTV habitation module as a "TransHab II", derived by the original TransHab concept.

Updated the CMES Orion configurations to better differentiate between the Orion EFT and the Orion MPCV versions. Still need to implement the RCS system on both of them.

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
@@ -72,3 +72,16 @@
 {
 	%RSSROConfig = True
 }
+
+//	==================================================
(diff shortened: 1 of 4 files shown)


NathanKell:
I'd suggest touching base with @Felger about the fairing sides--it seems like the sort of thing PFFE might want to do, but if RO does it we maybe should do it based off module, not partname?

PhineasFreak:
For some reason, it never occurred to me to check for the fairing module instead of the part module name, doh!

I will contact Felger and ask him if he can implement it.

PhineasFreak:
Ok, Felger agreed to add the feature to PFFE.

PR #566, opened 2015-10-03
SXT cockpits: Marble, EntenteCordiale

winged7:


NathanKell:
Um, these parts are very clearly B-52 and Concorde cockpits, so rather than generic sizeups they should be modeled on those specific aircrafts' sizes, masses, avionics wattage, etc. When parts are generic (like the Mk2 spaceplane parts, who knows what they should be) we do generic stuff, but when there's a clear real life analogue (these, Bonanza, Mk1 pod, etc) we use the real stats. :)

winged7:
Concorde fuselage dimensions: 
http://www.concordesst.com/dimentions.html
http://www.concordesst.com/graphics/frontondim.gif

"Fuselage max external width: 2,88m
Fuselage max external height: 3,32m"

i think that it would better to make 3,1m x 3,1m so it will be compatible with procedural tanks. The same case as for Kossak (B-737) cockpit - you can't get such irregular shapes with procedural tanks.

winged7:
B-52 data from wikipedia:
https://en.wikipedia.org/wiki/Boeing_B-52_Stratofortress

-Wingspan: 185 ft 0 in (56.4 m)
-Height: 40 ft 8 in (12.4 m)


My estimations for cockpit dimensions:
http://i.imgur.com/z4O9JGP.jpg

4,00m - height
3,33m -width

NathanKell:
Great work! :)

Let me ping @ferram4  in case he has ideas about the cockpit masses on either of these.

PR #567, opened 2015-10-04
SXT: B-737 Fuselage

winged7:


winged7:
Unfortunately I'm forced to give up here.They are already supported by RO, so I can't see my config updates in game. The same story for aircraft tail: http://i.imgur.com/TTq1zQO.jpg

Config for them is probably in some random place outside of SXT folder.

PR #574, opened 2015-10-07
KK Launchers Atlas V

A1Ch1:
Adjust mass, rename boosters to AJ-60A, add RL10C-1, adjust STAR-5F sep motor

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/KK Launchers/RO_KK_AtlasV_Adapters.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/KK Launchers/RO_KK_AtlasV_Adapters.cfg
@@ -3,10 +3,11 @@
 	%RSSROConfig = True
 	@rescaleFactor = 1
 	%scale = 1.0
+	@title = AJ-60A Decoupler
 	@mass = 0.5
 	@maxTemp = 1500
 	@manufacturer = Lockheed Martin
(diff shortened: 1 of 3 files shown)


raidernick:
Your isp values are wrong. The RL10A-4-2 is 446.4 and the RL10-C1 is 458

A1Ch1:
Source? Everything I've read shows C-1 has lower Isp since it is a converted B-2 with the extension removed. The origMass value is needed because the C-1 is higher mass than the A-4-2.

http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20110015783.pdf RL10C-1 448.5
http://www.astronautix.com/engines/rl10a42.htm RL10A-4-2 451

raidernick:
astronautix is 100% wrong 100$ of the time, never use them for anything unless you have no other choice. This is what pretty much all of the RO guys for for isp values. http://www.b14643.de/Spacerockets/Diverse/U.S._Rocket_engines/engines.htm

A1Ch1:
What about the AR website itself: http://www.rocket.com/rl10-engine

Specific impulse in vacuum 444.4–451.0 sec; mixture ratio: 5.1–5.5:1

451 at optimal mixture ratio

A1Ch1:
b14543 is a nice site, but those numbers are calculated rather than actual data.

raidernick:
the mixture ratio is going to change the isp, if you are going to change the isp, change the mix rate then also, otherwise don't change the isp

raidernick:
Looks good now. I agree with all the other changes too. NK should double check and it should be good to go.

PR #575, opened 2015-10-07
Remove gimbal from KK boosters

A1Ch1:
All Delta II boosters used fixed nozzles


raidernick:
Nathankell I had forgotten to do this, it should be ok to merge

PR #581, opened 2015-10-11
LiqHype - A6 - Remove extra "."

RedAV8R:
Are people actually testing their configs in game??? Because this flat out refused to load the way it was. Like crash KSP before it even finishes loading. Just hangs.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Liquidhype/RO_LqdH_Ariane6.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Liquidhype/RO_LqdH_Ariane6.cfg
@@ -139,7 +139,6 @@
 		}
 	}
 }
-
 @PART[upperStage]:FOR[RealismOverhaul]:NEEDS[RemoteTech]
 {
 	MODULE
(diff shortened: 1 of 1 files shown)


A1Ch1:
Just a typo. Thanks for fixing.

RedAV8R:
I'm not mad, and I don't wanna be a dick, but this is about the biggest 'just a typo' that one can have. Like Mariner I big.

I would have hoped that configs are actually tested in game before a PR is done. Sure some details can be missed here and there that one can tweak as time goes on, but obviously this one wasn't tested. At all.

For now on, if you haven't at least basic function tested items. Please put "NOT TESTED" in big print across any config and PR. So that others know.

*End rant*

A1Ch1:
If you really want a timeline, here's how it went down: created config. Tested. Created PR. Later, noticed it needed bottom node. Created bottom node. Copy/paste into PR, accidentally added "." that wasn't in own config. It happens. Again, thanks for finding and fixing.

A1Ch1:
My advice is to please keep in mind that this is a large collaborative project, with dozens of PRs and hundreds of commits between releases. Bugs will pop up, and even massive game-breaking bugs can be produced by simple human error rather than gross incompetence, so it does no one any good to assume the latter. I created a bug by copy-pasting into an existing PR, so I'll be sure to triple-check the next time that situation arises, with no ranting necessary to reinforce the lesson.

PR #586, opened 2015-10-14
SSTU Orion Config 1

stratochief66:
Start configuration of Orion CM, SM, LAS from SSTU.

Real scaling done, burns RealFuel. Fuel and masses still need tweaking,
TACLS not added yet.


Niemand303:
Awesome! Gonna find some info on the Altair lander. I think maybe we should try that Methane engine which was planned for making possible the ISRU refueling (hyped about Real ISRU), a sabatier cycle would work there.

stratochief66:
@Niemand303 
http://imgur.com/a/gDB2x

PR #591, opened 2015-10-15
Create SSTU_SRB

FFCJoseEduardo:
2 and 3-segment booster need to be done (I have no idea where can I get the information for these) and the solid fuel in the 4-stage one needs to be converted to PBAN


stratochief66:
Thrust curve doesn't apply correctly in-game, that needs to be fixed first.

PR #595, opened 2015-10-17
Update SSTU Upper Stages.cfg

FFCJoseEduardo:


stratochief66:
Tested in game, works well. Burns RF, mass and scale good. needs some fairing adjustment, which I plan to do shortly.

Thanks a ton @FFCJoseEduardo

PR #605, opened 2015-10-20
RN KK delta iii adds

raidernick:
-Added configs for delta III rocket


stratochief66:
@raidernick looks good, one thing confused me though:

https://github.com/raidernick/RealismOverhaul/commit/362618e897395ad8f61a63d7a7ffd52c0c34c397#diff-305ff8117220aee8e075fd6574ad7e0dR32

from the part name, I would expect it to contain RP, but you remove oxidizer and add LOX, which would make sense if it was originally configured as an oxidizer tank. Is the naming of the part just funny?

RedAV8R:
From what I can see the top large diameter tank was in fact the Kerosene tank, while the original diameter tank on the bottom was LOX. So they need switched around.

raidernick:
stratochief66 I did it based on what was in the tank originally, perhaps there was an error in the original tank as RedAV8R pointed out, I will swap them.

PR #607, opened 2015-10-22
SSTU. new dock, decoupler, interstage, mat bay configs

FFCJoseEduardo:


FFCJoseEduardo:
added the other docking port without parachutes, told RO that the decoupler and the interstage do works with RO (to avoid the non-RO message) since they work straight out of the box with RO, and also added, unaltered, a radial materials bay

PR #608, opened 2015-10-22
SSTU docking update to NASAdock, RO tags

FFCJoseEduardo:


FFCJoseEduardo:
changed the dockingnode config for the docking port to be compatible with NASAdock from RO and added two ROconfig lines because the parts are already working fine with RO

PR #610, opened 2015-10-24
Create 1st_Satellite.craft

Schnobs:
A Redstone/Explorer knock-off built from stock parts. Works just like the original, but is by no means an exact replica.
(File created by copy/paste, hope this works)


stratochief66:
I think this should be in the RP-0 repo, although we don't have a folder for RP-0 Stock craft there yet. @NathanKell ?

NathanKell:
@stratochief66 concur. @Schnobs sorry I didn't think of it at the time in IRC.

PR #620, opened 2015-10-27
Update RO_FASA_Atlas.cfg

ctiberious:
Clean up graphic defect on SLV-3C tank due to overlap of LFT_Gemini_Short and Mercury_Atlas_LFT_Long components.
Adjusted the dry weight of most of the Atlas tanks.  Although most references seem to contradict each other relating to the dry/wet mass of various Atlas tanks, I think these corrections come close.
Added the Atlas G/H/I tank.  I'm having trouble finding reliable specifications for this tank but most references seem to imply that its dry mass was slightly heavier, and that it was slightly taller, than the SLV-3C.  Most references also seem to imply that it's fuel mass was similar to the SLV-3A.  I've tried to reflect this in the provided stats.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -414,6 +414,37 @@
 		}
 		CONFIG
 		{
+			name = LR105-NA-5
+			minThrust = 366.1
+			maxThrust = 366.1
+			heatProduction = 100
+			massMult = 0.9443 // 413kg engine, source http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm for -5
+			PROPELLANT
+			{
+				name = Kerosene
+				ratio = 0.382
+				DrawGauge = True
+			}
+			PROPELLANT
+			{
+				name = LqdOxygen
+				ratio = 0.618
+			}
+			atmosphereCurve
+			{
+				key = 0 311
+				key = 1 215
+			}
+			cost = 100
+			entryCost = 2000
+			entryCostSubtractors
+			{
+				LR89-NA-5 = 1000
+			}
+			techRequired = advRocketry
+		}
+		CONFIG
+		{
 			name = LR105-NA-6
 			minThrust = 366.1
 			maxThrust = 366.1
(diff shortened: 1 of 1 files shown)


stratochief66:
I would prefer if this were done as a few separate PRs, if possible. Some changes are easily accepted (tweaking the tank positions, although I would have to look at this in game to confirm it before merging) while others are a bit more controversial and will need more thorough in-game testing.

At least one thing appears to be blatantly wrong, but of course I could be wrong to.
Line 77:
https://github.com/ctiberious/RealismOverhaul/blob/c3f6e2a5901294713d6bac272b7b77fec6cb8d9c/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg#L77

Atlas E/F tank ends up with a mass of 4.206 tons, 99.9kL volume. In comparison, at line 24 the Atlas D tank has a mass of 1.455 tons and 97kL capacity. I don't believe there is any way you end up with nearly 3 tons more dry weight to gain 2.9kL of volume.

The way the FASA parts are, converting researched dry masses to FASA mass values can be a big pain, and confusing. I also used bn14643.de to estimate the dry masses and capacities of a few of these Atlas tanks.

The LR-105 sustainer engine in FASA has some tank volume associated as well because of the 'tank butt' it has, which traditionally has been filled with LOX. Personally, I used b14643.de burn times to estimate, test, and refine proper FASA tank volumes for a few of the Atlas tanks.

ctiberious:
I wasn't sure the process I should be using to submit changes but I can certainly submit each change separately if that's the process you want me to use.  Just let me know if you'd like me to resubmit them as multiple pull requests.

As far as the Atlas E/F dry weight, what I can find is that the Atlas D had a dry weight of under 2400kg while the Atlas E/F had a dry weight closer to 5000kg.  I've got no idea why the E/F tank weighed nearly twice as much when it's fuel capacity was only slightly larger.  I can only assume that the E/F had more avionics/electronics built into it.
In game, I've got the Atlas D (Mercury - Atlas tank) with basic LR-105 and dual LR-101 weighing in at 2.347t compared to the Atlas E/F (with same engines) weighing in at 5.098t.  So it's possible the mass change I'm proposing here is off by 0.404t or so, but as you've probably noticed, finding references with good dry mass data for the older Atlas is pretty difficult.
As much as I hate using them, the most comprehensive source I've found so far is astronautix.com which we know is not completely reliable (though they do usually seem to jive with designation-systems.net).  But even if we can't trust the exact numbers given there, I have to believe (at least until I can find a better source) that if it lists one tank weighing twice as much as another, then there was some kind of significant weight difference.

NathanKell:
Is your source confusing the E/F sustainer dry weight with the sustainer+booster dry weight, maybe?

Here is a good source: http://forum.nasaspaceflight.com/index.php?topic=26915.80

stratochief66:
@ctiberious 

If you could turn this into a few PRs, that would be great, because I can merge some parts of this right away

1. Moving the tank to make a composite tank look better.

2. The tweaks to tank masses, excluding the E/F change.

3. The Atlas E/F tank change

4. The additional engine configs

If you could break this PR into a few separate PRs along these lines that would be awesome. We are planning for a release very soon, so this would allow me to review most of the changes in your PRs in time for them to make the next release.

ctiberious:
@NathanKell : The MA-3 booster is showing a weight of 3,174kg.  If I take that out of the 4,926kg the reference shows for the dry mass of the Atlas E/F, we'd end up with a dry mass of only 1,752kg.  That's only about 75% of the mass of the Atlas D.  Considering the Atlas E/F has larger internal tanks and improved pumping & electronics, I wouldn't expect it to be lighter.  Though considering how little reliable material there appears to be on the older Atlas missiles and rockets, I suppose it's possible.  That said, from what I can tell, all future versions of the Atlas (SLV-3, SLV-3A, LV-3C, SLV-3C and G/H/I) appear to be based on the Atlas D missile.  Not the Atlas E or Atlas F missiles (though a few Atlas-Agena and Atlas-Able rockets were built with the Atlas E/F fist stage).  You'd think that if the E/F models were better and lighter, they'd have been used as the base for all future rockets instead of the older version.

@stratochief66 : I'm working on resubmitting as multiple pull requests now.

NathanKell:
Ok. Then I'd say perhaps the best course of action is to devour that thread, and ask the question on it if there's no answer there already.

PR #623, opened 2015-10-28
Update RO_FASA_Atlas.cfg

ctiberious:
Clean up Atlas SLV-3A composite tank by adjusting size and placement of the "LFT_Gemini_Short" tank.  New model, with LR-105 engine results in a 21.4m height.


stratochief66:
Thanks for providing a metre length comparison. :)
It looks good in game to me.

PR #624, opened 2015-10-28
Update RO_FASA_Atlas.cfg

ctiberious:
Added composite model for Atlas G/H/I first stage tank.  I noticed that there is a reference to a "FASAAtlasH" already in RP-0 even though no corresponding tank exists in RO so I've re-used that name for this new part.  The resulting part has the same basic fuel capacity as the Atlas SLV-3A with a height (when LR-105 engine is attached) of 23.5m.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -268,6 +268,48 @@
 		}
 	}
 }
++PART[FASAMercuryAtlasLFTLong]:AFTER[RealismOverhaul]
+{
+	@name = FASAAtlasH
+	@MODEL,0
+	{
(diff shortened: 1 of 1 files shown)


stratochief66:
The mass of this tank appears just a little wonky (a bit off from some of the bnumz source data I looked at), I would suggest looking at bnumz source data and perhaps see if you find good data reasons to tweak this mass a bit sometime.

http://www.b14643.de/Spacerockets_2/United_States_4/Atlas/Description/Frame.htm

@ctiberious

ctiberious:
I think the issue probably lays in the fact that currently the engine weights aren't being changed.  I'm still researching the weights of the different LR105 and LR89 engines.  Once I have those, I'll do another pass over all the Atlas tanks so their their wet/dry mass goes along with the specific engine they were most used with.

btw, the link you mentioned.  I don't think I see any mass data although the page is all in German which I don't speak or read. :)

stratochief66:
@ctiberious On the point of bnumz, try the Google Auto-Translate page function, it works perfectly for that site because it doesn't have to parse sentences or grammar :)

PR #625, opened 2015-10-28
Update RO_FASA_Atlas.cfg

ctiberious:
Dry and Wet mass adjustments.  These new masses are taken from astronautix.com which I know it's wholly reliable but it is the only source I can find with even marginally accurate data on the older Atlas rockets.  The proposed mass and fuel alterations will result in matching wet and dry masses taken from that website assuming two LR-101 Verner engines and the correct LR-105 sustainer engine are included in the final craft.


stratochief66:
http://www.b14643.de/ is also a decent source, for example:
http://www.b14643.de/Spacerockets_2/United_States_4/Atlas-Agena/Description/Frame.htm

Most of the proposed changes are performance neutral, and the more major changes (ie. SLV-3C) appear pretty reasonable on a kL/dry mass ratio basis, and most importantly more reasonable in appearance than the values I had in there before / wherever they came from originally.

NathanKell:
(Just commenting on one of these, though it goes for all -- @ctiberious thanks for all this! :)   )

PR #626, opened 2015-10-28
Update RO_FASA_Atlas.cfg

ctiberious:
Based on the astronautix site (where I got all the masses for all the old Atlas missiles/rockets) the Atlas E/F dry weight is about 210% the dry weight of the Atlas D.  I can't find any information as to why the Atlas E/F is +2,579kg heavier.  It has been suggested that maybe the source has included the weight of the boosters as well, but if that's happened the Atlas E/F would only weigh in at 1,752kg which is only 75% of the dry weight of the Atlas D.  Presumably if the Atlas E/F was a more advanced design (which it was) that also weighed less, you'd expect that all future models of the Atlas would have been based on the E/F design.  Instead, all future Atlas models (SLV-3, SLV-3A, LV-3C, SLV-3C, G, H & I) appear to be based on the older Atlas D.  Considering how important weight is in a launch vehicle, if the E/F were heavier (as astronautix claims it is) that would certainly explain why the older version was ultimately used as the design template.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -74,7 +74,7 @@
 	@title = Atlas-E/F Fuel Tank
 	@description = The fuel tank for the Atlas-E/F series.
 	@attachRules = 1,0,1,1,0
-	@mass = 1.506
+	@mass = 4.081
 	!RESOURCE[LiquidFuel]
 	{
 	}


NathanKell:
Resumed discussion from old thread: Ok. Then I'd say perhaps the best course of action is to devour that Nasaspaceflight thread, and ask the question on it if there's no answer there already.

stratochief66:
with the Atlas tanks, I would expect dry weight and fuel volume to scale fairly linearily, or at least move in the same direction as each other. We would have to see data from a few sources to set the RO tank mass so out of line with the rest of the family of tanks, I think. I understand solid data sources can be hard to find for this hardware, but that is the only way forward I see regarding such a change.

Try the Google Auto-translate feature on the bnumz link I posted in the other Pull if you want to try that as a source @ctiberious

ctiberious:
So, as mentioned, on Astronautix I'm seeing empty mass of the Atlas E/F listed as 4,926kg, gross mass of 117,826kg and mass of the booster at 3,174kg.  So stage 0 + 1 shows with an empty mass of 8,100kg and gross mass of 121,000kg.
Now if I look at b14643.de (specifically http://www.b14643.de/Spacerockets_2/United_States_4/Atlas/Description/Frame.htm looking at the "Atlas (2)" page, specs for "Atlas F/OV") we see a stage 1 liftoff weight of 121.00t and a propellant weight of 112.9t.  Since the liftoff and propellant weight of the stage 0 section are both "--" I'm assuming that the mass of the boosters is included in the liftoff weight for stage 1.

Unfortunately none of the Atlas D (and LV-3) data on b14643.de (look at the above link, "Atlas (1)" page) lines up exactly with what I'm finding on astronautix.  Astronautix lists the Atlas D with dry mass of 2,347kg, gross mass of 113,050kg and booster mass of 3,050 kg.  That would give a total liftoff weight of 116,100kg, a total dry weight of 5,397kg and a propellant weight of 110,703kg.  Closest match to that on b14643.de is the LV-3 with liftoff weight listed at 115.7t and propellant weight listed at 110.3 (resulting in dry weight of 5.4t).  Though there are versions of the Atlas D listed with dry weights of up to 7.25t.

ctiberious:
Here's a bit more data showing the difference between the Atlas D (LV-3) from http://www.b14643.de/Spacerockets_2/United_States_4/Atlas-Agena/Description/Frame.htm "Vehicle Design" link versus astronautix.  b14643.de shows the lv-3 with liftoff weight at 120.2t and propellant weight at 114.76t resulting in 5.44t dry weight (including stage 0).
Astronautix shows gross mass of stage 1 + 0 at 116,100kg, empty mass at 5,397kp which give a propellant mass of 110,703kg.  The designs on b14643.de have more propellant which results in a higher liftoff weight, but in both cases the dry weight lines up.

Compared to the 8.1t dry weight of the Atlas E/F shown in my previous post.  So that's two sources that both seem to agree that the Atlas E/F had a significantly heavier dry weight compared to the Atlas D.

And since we know that later versions of the Atlas (SLV-3, SLV-3A, LV-3C, SLV-3C and G/H/I) were all based on the lighter Atlas D, that would explain why those all have a lower empty mass than the Atlas E/F.  I still don't know why the E/F was so much heavier than the D line, but it certainly seems to be the case.

PR #627, opened 2015-10-28
Update RO_FASA_Atlas.cfg

ctiberious:
Update engine configurations for the Atlas LR-89 and LR-105 engines.  All engine data was pulled from b14643.de site.  I considered altering the configuration names to also include the "MA-x" designation but I wasn't sure if that would be confusing or not.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -619,6 +619,7 @@
 		name = ModuleEngineConfigs
 		type = ModuleEngines
 		modded = false
+		origMass = 0.720
 		configuration = LR89-NA-3
 		CONFIG
 		{
@@ -626,6 +627,7 @@
 			minThrust = 748.7
 			maxThrust = 748.7
 			heatProduction = 100
+			massMult = 1.0
 			PROPELLANT
 			{
 				name = Kerosene
@@ -642,7 +644,6 @@
 				key = 0 282
 				key = 1 248
 			}
-			massMult = 0.89
 		}
 		CONFIG
 		{
@@ -650,6 +651,7 @@
 			minThrust = 758.7
 			maxThrust = 758.7
 			heatProduction = 100
+			massMult = 1.0
 			PROPELLANT
 			{
 				name = Kerosene
@@ -680,6 +682,7 @@
 			minThrust = 831.4
 			maxThrust = 831.4
 			heatProduction = 100
+			massMult = 1.086
 			PROPELLANT
 			{
 				name = Kerosene
@@ -710,6 +713,7 @@
 			minThrust = 931.7
 			maxThrust = 931.7
 			heatProduction = 100
+			massMult = 1.414
 			PROPELLANT
 			{
 				name = Kerosene
@@ -742,6 +746,7 @@
 			minThrust = 950.8
 			maxThrust = 950.8
 			heatProduction = 100
+			massMult = 1.414
 			PROPELLANT
 			{
 				name = Kerosene
@@ -774,6 +779,7 @@
 			minThrust = 1077.6
 			maxThrust = 1077.6
 			heatProduction = 100
+			massMult = 1.783
 			PROPELLANT
 			{
 				name = Kerosene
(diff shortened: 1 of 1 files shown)


stratochief66:
@NathanKell the LR89 7.1 & 7.2 have ~20kN thrust difference, shall we keep the new config in?

@ctiberious LR105 5&6 are identical, so please remove the 5 config you added. perhaps rename the 6 to 5/6 ? @NathanKell ?

The LR105 7.1 & 7.2 configs only have a 1kN difference, please remove the one you added.

The LR89 3 & 5 have a 10kN difference, sufficient to add the config @NathanKell ?

NathanKell:
I would tend to not bother with 7.1 at all, not least because in RP-0 terms it's in the same tech node anyway. As to 3 vs 5 vs 6, I would standardize on 3 and 6 across the board so we don't have mismatched numbers for LR105 and LR89.

NathanKell:
That reminds me, at some point I need to add 'notes' functionality to CONFIGs, so there can be note tooltips describing this sort of thing.

ctiberious:
Actually, 3 was never really flown on a launch vehicle.  The MA-1 engine block was used primarily on the early Atlas A, B & C which were just missile systems.  So if you want to prune out some of the configs, 3 is probably the most expendable.  It's kind of nice having it for people that want to recreate the earlist suborbital Atlas flights, but that's about all it was ever really used for.

As far as 5 is concerned, I'm still researching but I think that I'm going to find that the weight of the LR105-NA-5 is different from the weight of the LR105-NA-6.  That's really the only reason I can see that they'd relabel the engine.  Sure, the NA-5 was part of the MA-2 engine block, and the NA-6 was part of the MA-3 block.  And the MA-3 had more powerful boosters (as seen in the 9.6% thrust increase on the LR89-NA6), but I have to believe that something on the LR105-NA6 also got changed.  Let me keep researching and if I can't find anything that would have an effect in game, we can always combine the two configs into one at that point.

As for the 7.1 and 7.2, there is little difference between the two, but the difference adds up.  For one, the 7.1 was used exclusively (from what I'm seeing) on the Atlas-Agena while the 7.2 was only used on the Atlas-Centaur.  7.1 has slightly less thrust which results in better fuel efficiency and longer burns (i.e., more overall dV).  With the lighter upper stages, the Atlas-Agena needed less initial thrust to get off the pad but would have benefited from the marginally improved dV.  While the heavier upper stages of the Atlas-Centaur would have needed every bit of initial thrust they could get and wouldn't have suffered as greatly from the marginally reduced dV.  Based on my numbers, the MA-5.1 block used 11.0041kg/s (or 10.8055L/s) less fuel than the MA-5.2 block.  But the MA-5.2 had about 1.75% more thrust.

stratochief66:
The NA-5 & 6 could also differ in manufacturing process, or cost, or reliability. These are some other aspects that I think might lead them to relabel it while not impacting the features we can plug into RO.

ctiberious:
That's certainly a possibility.  I'm still trying to find weights specifically for the LR105 engines.  I can find information on the LR89 (more or less) since it was part of a separate stage but most mass data I find include the fuel tank.  I will keep looking, though.  If it turns out there really is no in game difference between the 5 and 6, then I can see the point of merging them into a single config.  Assuming you and Nathan don't think it would confuse anyone.

ctiberious:
I found a source that implies there might have been a difference in the amount of thrust on some version of the LR105-NA-5.  The site is http://www.designation-systems.net/dusrm/app3/b-2.html which lists the Atlas B, C and D with 363,231N thrust LR105-NA-5 engine.  This site is not 100% reliable since it then goes on to say that the Atlas E and F had a 386,395N thrust LR105-NA-5 engine and that the Atlas G and H had a 363,395N thrust LR105-NA-7 engine.  But it's one of the first sources that I've come across with a thrust for an older LR105-NA-5 that doesn't line up with b14643.de

ctiberious:
I added in mass changes for the various LR89 configurations.  Now the mass of two boosters plus the skirt should match what I'm finding for the stage 0 mass on astronautix.  I am still looking for a stage 0 mass for the NA-3 config so for now I have it set with the same mass as the NA-5.

stratochief66:
@ctiberious can you post those changes as a separate PR from this?

Also, please look at data from multiple sources, and cite them if/when you can.

ctiberious:
I can, but the mass data is for all of the expanded LR89 configs I submitted.  Not just the ones that are already part of RO.  So if I create a separate PR, it would have all the other config changes in it as well.

As for sources, I used astronautix for the weights since that is the site I used for the bulk of the Atlas related weight changes I've been submitting.  But b14643.de mostly agree with the values.

PR #628, opened 2015-10-28
Update RO_FASA_SRBs.cfg

ctiberious:
I noticed that when attaching anything to the sides of the UA1207, the parts appears to be "inside" the SRB.  This adjustment should correct that so it looks more like parts are attached to the outside of the SRB.


stratochief66:
Nicely done, I should note this down. I didn't know you could fix that problem this way before.

PR #629, opened 2015-10-28
Update RO_FASA_ApolloCSM.cfg

ctiberious:
For the CM change, as it stood the CO2 Scrubber only scrubbed about 95% of the CO2 out of the air.  That meant you were constantly building up CO2 with no way to actually remove it.  And if the scrubber every stopped working for a short time, you'd never be able to recover from that.  We know from historical events (Apollo 13) that the scrubbers were actually able to reduce the amount of CO2 in the air.  The new value will convert 105% of CO2 while generating the same amount of O2.

For the SM change, the old fuel cell did not produce enough electricity to power the CSM plus the high gain antennae.  The new values should be enough for both.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -576,7 +576,7 @@
 		converterName = Fuel Cell
 		conversionRate = 1.0
 		inputResources = LqdHydrogen, 0.000296379, LqdOxygen, 0.000210317
-		outputResources = Water, 0.000261, true, ElectricCharge, 2.8, true
+		outputResources = Water, 0.000261, true, ElectricCharge, 2.2, true
 	}
 	MODULE
 	{
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -314,7 +314,7 @@
 		name = TacGenericConverter
 		converterName = CO2 Scrubber
 		conversionRate = 3.0	// # of people - Figures based on per/person
-		inputResources = CarbonDioxide, 0.0058912100, ElectricCharge, 0.010, LithiumHydroxide, 0.00001189
+		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010, LithiumHydroxide, 0.00001189
 		outputResources = Waste, 0.00003847, false
 	}
 	@MODULE[ModuleRCS]
@@ -576,7 +576,7 @@
 		converterName = Fuel Cell
 		conversionRate = 1.0
 		inputResources = LqdHydrogen, 0.000296379, LqdOxygen, 0.000210317
-		outputResources = Water, 0.000261, true, ElectricCharge, 2.2, true
+		outputResources = Water, 0.000261, true, ElectricCharge, 2.8, true
 	}
 	MODULE
 	{


stratochief66:
the high gain antenna isn't running at all times, which is why it isn't factored into power consumption.

does it consume power at all times with certain mods, like RT, or just during data transmission?

stratochief66:
can you source what you mean by 'only remove 95%' ?

the scrubber type, IIRC from my research, only absorbs CO2 but doesn't emit O2. the potassium superoxide used by the soviets for some hardware, iirc, both absorbed co2 and emitted o2.

ctiberious:
The high gain antenna is always using power when you turn it on.  I suppose in a manned vehicle you don't need to turn it on except when transmitting data (which happens automatically).

As for the CO2, firstly, I should have said 99.5%.  Sorry.  Anyway, I don't have a "source".  That's just math.  When using TACLS, one kerbal produces 0.00592 CO2/sec.  The CO2 scrubber on the CSM currently can only scrub 0.0058912100 CO2/sec (with a conversion rate of 3 to account for the 3 kerbal crew).  So by default if you have a 3 man crew, you're going to be generating 0.00008637 CO2/sec above what the current scrubber is able to clean.  Thus, you are always going to be increasing the amount of CO2 in the air.  Admittedly, I arbitrarily select 105% scrub rate, mostly because the scrubbers on the Apollo 13 were so quickly able to recover from near lethal levels.  But we could change the input rate to 0.0059496 which would be a 100.5% scrub rate.  That's basically the same rate at which O2 is created with the LOX-O2 generator.

stratochief66:
In this case, I suppose by sourcing I meant 'show your math', yeah.

My original aim at the scrubbing rates was to be pretty close to 100% [+/- 0.5%, woot :) ]

My concern with a rate of 105% then would be that I don't exactly know how TACLS works under the metal, so that change might cause it...

I may as well leave that in, so you can see which wrong path I was going down. I was concerned that a 5% would also consume 5% more LithH, which you clearly didn't alter, so I suppose I'm fine with the change. I took the CO2 depletion numbers from some calculations from TACLS data, and the final mass of the LithH is from a source on the mass of the air cleaners used on the Apollo.

@NathanKell, any comments or inputs? Merge?

ctiberious:
About the only reason I would advocate for the higher fuel cell output is that in game we have time warp.  The higher fuel cell output means you can turn the antenna on and not think about it again.  Or, possibly more importantly, you can turn the antenna on and not worry that when you time warp the 3 day trip to the moon you've used up the bulk of your power because you accidentally left it running.  :)

stratochief66:
That would make things easier in the case of user error, but it doesn't serve to make the part more realistic. If someone wants to hedge against that chance they can also throw solar panels on the SM, which I would argue is a cooler looking way to do it. :P

If a more capable LS tool comes along, or TACLS adds some functionality, the fuel cells could be handled a bit more realistically. In reality, the SM had 3 fuel cells and generally ran on 2, not quite at full capacity IIRC. So, if TACLS let us simulate a "fuel cell" capable of simulating 0,1,2, or 3 cells running at a time, battery charging and depletion in a more realistic fashion could be achieved. As the mods we rely on stand for simulation I think we will leave the numbers as they stand.

Thanks for the help!

ctiberious:
Technically, you could.  Or at least you should be able to though I've never tried it.  You'd just have 3 generators instead of 1.  Then you'd have the ability to turn them on individually.  I'd have to test that.

Should I resubmit the CO2 scrubber changes as a separate PR?

stratochief66:
Yep!

ctiberious:
Instead of submitted a whole new PR, can you just reopen this one and I can edit it?  Or is there an issue with editing an open PR?

ctiberious:
2nd commit only has the CO2 scrubber change.  Fuel Cell change has been removed.

PR #630, opened 2015-10-28
Update RO_FASA_ApolloCSM.cfg

ctiberious:
We know from historical records that the Apollo CSM had a camera system.  My guess is they had a lot of other scientific equipment as well but most of that can be added into the equipment bays.  We don't have a camera (or material lab in stock) that will properly fit into the equipment bay so I think adding the experiment to the SM directly is a valid way to correct this issue.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -651,6 +651,25 @@
 			key = 1 100
 		}
 	}
+	MODULE
+	{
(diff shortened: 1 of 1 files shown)


stratochief66:
@NathanKell, does this seem reasonable from an RP-0 perspective?

NathanKell:
The camera the CSM had, as I discussed with @schnobs , is currently under the metric of 'Crew Report' -- the film return camera is a rather different beast. We do intent at some point to add a transmission camera (for Luna 3, the Rangers, etc)--and that could be added to the CSM, as that's more the level of camera it had.

The film return camera would be for the unflown plan to fly a Keyhole-dervied camera as replacement payload if an LM wasn't ready in time.

PR #634, opened 2015-10-31
configs for SSTU landers, only needs balancing

FFCJoseEduardo:
First pass config for SSTU landers, it works but might need balancing for fuel and masses

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Landers
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Landers
@@ -1,0 +1,1269 @@
+@PART[SSTU_LanderCore_LC2-DESE]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	!MODULE[TweakScale]
+	{
+	}
+	@rescaleFactor = 1.25
+	@mass = 0.076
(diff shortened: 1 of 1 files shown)


stratochief66:
@FFCJoseEduardo 

This downloaded as .txt for me, did you name it as *.cfg like the rest of the files, or is it currently with no extension or .txt?

stratochief66:
The LC2 variant that doesn't show external fuel tanks probably shouldn't report having MMH & NTO. That is possible with the switchy-tanks, correct?

FFCJoseEduardo:
I probably forgot to add an extension when I created the file... my bad

as for the LC-2, it's the same with the other 2 pods, the version without tanks has NTO and MMH for the RCS, they have small amounts of it (in stock it has 50 mono, I added 50 of each for RO, same thing goes for the tank versions, they have what the rescaled volume can handle+50 each for the monoprop it had in stock)

stratochief66:
Have the ascent (ASCE) engines been scaled up? they appear a bit small.

FFCJoseEduardo:
they all have, same scale for all lander parts, 1.25, same as Orion

it's just that they do look really small

stratochief66:
Perhaps you should get on #RO, we could be talking live :P

PR #640, opened 2015-11-02
Remove KAS folder installation and add more mods

Olympic1:
RO didn't update on CKAN because the `KAS` folder was removed from the download.
I also added some more mods to the list.


NathanKell:
Merging. Will netkan rerun or do I have to tag another release?

Olympic1:
It is added to CKAN

NathanKell:
Great! :)

PR #641, opened 2015-11-02
Update RO_FASA_Atlas.cfg

ctiberious:
It appears someone tried to have the mass of the LR89-NA3 be different from the mass of the other LR89 engines (they included a massMult = 0.89) but massMult will apparently not do anything if "origMass" isn't included in the ModuleEngineConfigs so I've added the line.
I moved the massMult for the LR89-NA-3 configuration to just below heatProduction just to be consistent with where I see this tag being used elsewhere.  I should point out that I can find no weight data on the LR89-NA-3 so I've simply left the massMult as it already was.
I've based the weights on the MA-3, MA-5 and Atlas II information from the astronautix website.  When combined with the correct LR-105 sustainer, two LR-101 verner engines, and the correct tank, the take off weights line up with what I see on both the astronautix.com and b14643.de websites.
I should note that I used the Atlas II site from astronautix instead of the MA-5A site because it appears that the MA-5A site is using the weight of each individual booster instead of the engine and skirt as a whole.  Since I can find no weight specifically for the skirt, I made the assumption that the 1.61t we currently have was correct and uniform across all versions of the Atlas (which is likely not correct but I doubt we want to have multiple parts for the skirt).  I therefore adjusted the weights of the boosters to cover the difference.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -599,6 +599,7 @@
 		name = ModuleEngineConfigs
 		type = ModuleEngines
 		modded = false
+		origMass = 0.720
 		configuration = LR89-NA-3
 		CONFIG
 		{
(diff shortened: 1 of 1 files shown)


stratochief66:
Thanks for the PR, and the detailed explanation. Still waiting on NathanKell to weigh on for some of the other ones.

PR #642, opened 2015-11-02
Add Atlas D (MA-2) engine setup

ctiberious:
Added a configuration for the LA89-NA-5 which is part of the Atlas MA-2 engine.  While the booster is only slightly more powerful than the NA-3 version (the MA-1 engine), it is the correct values for recreating the Atlas D rocket.
I've also relabeled the LR105-NA-6 to (hopefully) show that it was basically used in both the MA-2 and MA-3 engines.  There is no game difference between the LR105-NA-5 and LR105-NA-6 (the NA6 had a slightly higher chamber pressure) so no reason to actually include a separate configuration.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -651,7 +651,7 @@
 			entryCost = 4000
 			entryCostSubtractors
 			{
-				LR105-NA-5 (-6) = 1000
+				LR105-NA-5/6 = 1000
 			}
 			techRequired = advRocketry
 		}
@@ -677,11 +677,12 @@
 				key = 0 290
 				key = 1 256
 			}
-			cost = 200
-			entryCost = 4000
+			cost = 300
+			entryCost = 7500
 			entryCostSubtractors
 			{
 				LR105-NA-5/6 = 1000
+				LR89-NA-5 = 3000
 			}
 			techRequired = heavyRocketry
 		}
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -420,7 +420,7 @@
 		}
 		CONFIG
 		{
-			name = LR105-NA-6
+			name = LR105-NA-5 (-6)
 			minThrust = 366.1
 			maxThrust = 366.1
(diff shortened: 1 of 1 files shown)


stratochief66:
@NathanKell 

yay or nay to the slightly different config, and/or the proposed labelling to show that the 5 & 6 variants are the same?

NathanKell:
Ok. It's going to be a bit of a headache over in RP-0, but I can see the reasoning for including _the_ Atlas-D configs, for Mercury reenactment.

stratochief66:
@ctiberious 

If you feel up to trying, we would more than appreciate you giving a go at another PR to align RP-0 to work with the Atlas changes you've been making. :)

stratochief66:
@NathanKell so, good to merge, or was there some aspect of this PR I would prefer changed/removed? How do you like the labeling for the 5/6 version?

https://github.com/ctiberious/RealismOverhaul/commit/da75431c6535f27c94e6de6f70ab715bad45ff5a#diff-3f074ad242f14d80315ab91db53bfbbcR423

ctiberious:
Just let me know if there are adjustments that need to be made and I'd be happy to make them.  Either in a new PR or by simply updating this one.

NathanKell:
Yep, as strato says, it'd be great if you could do the RP-0 PR (or is it in RO itself? wherever) to set techRequired and entryCost (and the entry cost subtractors) for the new config.
5/6 will work fine for the other.

Good to go when the above is done :)

ctiberious:
I've relabeled the LR105-NA-5/6

As far as techRequired, I've updated it so we get the -3 at Basic Orbital Rocketry (unchanged), the -5 at Mature Orbital Rocketry (where we used to get the -6), the -6 at Heavy orbital Rocketry (new), the -7 at Refined Rocketry (unchanged) and the RS at "experimentalRocketry" (unchanged).  Presumably there is something done in RP-0 to move the RS to Advanced Rocketry (veryHeavyRocketry) but I don't think that exists on the tree in RO.  Fortunately, since the RS does not need to be moved, it should be fine as is.

As far as cost and entryCost, I made some adjustments so that the -5 and -6 have different values but since I'm not sure where you got those values originally (and I'm only guessing at how some of the fields work), the best I can do is guess at values based on the progression that already exists.  Hopefully the changes I'm proposing are acceptable.

Oh, and with the exception of the RS engines (which I didn't change anyway), it looks like techRequired, cost and entryCost are all part of the RO file.

NathanKell:
It does not matter whether a node exists or not before RP-0, since adding the techRequireds in RO is only for the purpose of RP-0. and AFAIK veryHeavyRocketry _does_ exist even stock, since that's where the stock ARM/NASA/3.75m parts go.

PR #643, opened 2015-11-02
Separate MA-5 Atlas engine into Agena and Centaur varients

ctiberious:
Based on what I'm finding on astronautix.com and b14643.de, there were two difference versions of the MA-5 engines: LR89/105-NA7.1 and LR89/105-NA7.2.  The 7.1 version appears to have been used exclusively on Atlas-Agena rockets while the 7.2 version was used on Atlas-Centaur.  The 7.1 version had less thrust and isp but resulted in slightly higher dV (about +2% based on in game mechjeb data) than the 7.2 version.  Presumably the lighter, less powerful Agena needed less SLT to get off the pad and needed the MA-5.1 engines to do a bit more of the work.  While the heavier, higher dV Centaur needed more initial SLT for launch but didn't need the MA-5.2 engines to do all the work of getting something into orbit.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Atlas.cfg
@@ -452,6 +452,7 @@
 		CONFIG
 		{
 			name = LR105-NA-7.1
+			description = MA-5.1 engine for Atlas-Agena launches
 			minThrust = 385.2
 			maxThrust = 385.2
 			heatProduction = 100
@@ -484,6 +485,7 @@
 		CONFIG
 		{
 			name = LR105-NA-7.2
+			description = MA-5.2 engine for Atlas-Centaur launches
 			minThrust = 386.4
 			maxThrust = 386.4
 			heatProduction = 100
@@ -689,6 +691,7 @@
 		CONFIG
 		{
 			name = LR89-NA-7.1
+			description = MA-5.1 engine for Atlas-Agena launches
 			minThrust = 931.7
 			maxThrust = 931.7
 			heatProduction = 100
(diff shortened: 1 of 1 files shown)


Olympic1:
You can commit changes to the same branch, you don't need to make a new PR every time

stratochief66:
@Olympic1 I had asked him to peel a few changes off to separate branches for some of the past PRs he made, since some of the changes were likely to be rejected and others likely to be merged.

He his also including a fairly verbose description of the changes with each of these PRs, which to me  is another valid reason to do them as separate PRs.

@ctiberious I probably won't get a chance to look at these until we get the next RP-0 release out, so in roughly a week or so.

Olympic1:
@stratochief66 Sorry, didn't know

stratochief66:
@NathanKell 
yay or nay to the new, slightly different config?

NathanKell:
I believe my original reason for combining them was for RP-0 reasons, since they'd be in the same node. But presuming RP-0 changes get made too, that's fine.

I (or @Starwaster) needs to add tooltips on CONFIG buttons that show what the stats are, so you can know before you buy.

ctiberious:
Since these are basically two versions of the same engine, I don't think I'd change the techRequired, the cost or the entryCost.  They should become available at the same tech node and players should have to purchase them (or not purchase them if that's their choice) separately.  That said, it would be nice if there was a way to denote that one is for the Atlas-Agena and the other is for the Atlas-Centaur.  But short of changing the engine name, I don't know how to do that.

NathanKell:
As of RF for 1.0.5, you can now add a
`description = Engine for Atlas-Agena launches.`
or whatever, to a CONFIG node.

ctiberious:
Ok.  I've added the description field for the 7.1 and 7.2 engine configurations.

PR #649, opened 2015-11-05
Update RO_FASA_ApolloCSM.cfg

ctiberious:
Instead of a single fuel cell that produces 2.2ec/s, this change sets up 3 separate fuel cells which matches what was on the actual CSM.  I've set each fuel cell so that they create 1ec/s while using 2.2x less LOX/LH2 and generating 2.2x less Water.  This means each fuel cell should have the same efficiency that the original.
Only 2 fuel cells are needed to maintain ElectricCharge for the entire duration of a flight which, with the LOX-O2 generator running, gives the default CSM about 15.3 days worth of operating fuel.
I did set the power generation of each fuel cell to be less than half of what the original fuel cell could produce (1.0 vs 2.2).  This means that while 2 fuel cells can produce enough EC to cover the per second usage, they don't generate anything extra so if you use additional EC (like running the high gain antenna, transmitting data or running power hungry science experiments), you cut into the crafts battery.
You can also turn on the 3rd fuel cell which increases power generation enough to have a small surplus even while running the high gain antenna but at a cost of longevity.  Based on the default craft, running all three fuel cells (plus the LOX-O2 generator) means you only have enough fuel for a 10.3 day flight.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -573,10 +573,26 @@
 	MODULE
 	{
 		name = TacGenericConverter
-		converterName = Fuel Cell
+		converterName = Fuel Cell 1
 		conversionRate = 1.0
-		inputResources = LqdHydrogen, 0.000296379, LqdOxygen, 0.000210317
-		outputResources = Water, 0.000261, true, ElectricCharge, 2.2, true
+		inputResources = LqdHydrogen, 0.0001347177, LqdOxygen, 0.0000955986
+		outputResources = Water, 0.0001186364, true, ElectricCharge, 1.0, true
+	}
+	MODULE
+	{
+		name = TacGenericConverter
+		converterName = Fuel Cell 2
+		conversionRate = 1.0
+		inputResources = LqdHydrogen, 0.0001347177, LqdOxygen, 0.0000955986
+		outputResources = Water, 0.0001186364, true, ElectricCharge, 1.0, true
+	}
+	MODULE
+	{
+		name = TacGenericConverter
+		converterName = Fuel Cell 3
+		conversionRate = 1.0
+		inputResources = LqdHydrogen, 0.0001347177, LqdOxygen, 0.0000955986
+		outputResources = Water, 0.0001186364, true, ElectricCharge, 1.0, true
 	}
 	MODULE
 	{


stratochief66:
NIce, good idea. 

1. Did you test this in game? In flight?

2. Does the mass of the craft remain constant?

ctiberious:
2 is the easy answer.  Yes, the craft's mass remains constant.  "FASA Saturn V Apollo Lunar" shows 2975.734t in VAB, "FASA Saturn iB Apollo" shows 594.662t and self made CSM (CM, decoupler, SM and engine) show 30.116t.

As for 1, I have completed one "short" test were I launched the default "FASA Saturn V Apollo Lunar" and then just advanced time with 2 and 3 fuel cells running.  I'm in the process of running a longer (i.e., more controlled time advancing) test now.  It's using the same "FASA Saturn V Apollo Lunar" craft file though I added a bunch of science parts to both the CSM and LM.
So far I have completed the 5 day flight to the moon (can't figure out why but my lunar transit time is usually around 4 days with close to another day in Lunar SOI before the LOI burn) with all three fuel cells running, reached lunar orbit and just landed the LM.  It looks like that craft file doesn't include a fully charged LM (the descent stage has 172.8k of 216.0k), plus I transmitted close to 20 science packages so I ran on all three fuel cells during the 5 day lunar transit in order to recharge all the batteries (most of the trip with the high gain running).  By the time I reached LOI burn, all the batteries were basically fully charged.  Now that I'm in lunar orbit, I've turned one fuel cell off and transmitted probably 30 more science packages so my CSM is down to about 79% battery power.  A couple of the science parts I added use a tiny bit of power when left running but with them off the 2 cells look like they are generating about 0.02ec/s more then the CSM is currently using so it seems like having each cell generate 1ec/s works well.
I had the CSM+LM in lunar orbit for around 6 hours (basically 3 orbit) before having them separate.  At that point the LM descended to a 30km circular orbit in preparation for landing.  The LM has been on the lunar surface for about 30 minutes and I plan on having it remain for 3 days.  I will continue to monitor the CSM for those 3 days, plus the expected 4-5 day transit back to earth, and will add more comments once the trip completes.

stratochief66:
Thanks! The details of your remaining testing would be appreciated.

I believe the default LEM isn't fully charged to represent the difference between the short stay LEM and the later, J class longer stay LEM. Any users should feel free to fill any parts they want in the VAB, the extra capacity wouldn't be there if it wasn't meant to be used :)

My guess would be that a shorter lunar transit time was/is achieved because of the extra velocity the  CSM/LM were given to achieve an initial free return trajectory. Only part way through the flight there do they reduce their speed to target a close lunar flyby (like Apollo 13 did, before the accident), which is again later modified into a lunar orbit.

stratochief66:
@ctiberious poke?

ctiberious:
Sorry.  I was without a computer for the last couple days so testing was stalled.  I expect to have the flight completed today so I'll get more notes out in a few hours.

ctiberious:
I spent 3 full days on the lunar surface then launched the ascent module and completed a rendezvous with the CSM.  I spent most of that 3 days focused on the LEM but I did make it a point to check on the CSM every 12 hours or so.  No issues jumped out and LOX, LH2, LithiumHydroxide, CO2, O2 and Water usage/generation was right in line with what my math said it should be (I have "Max Delta Time" in TACLS set to 4000 so that advancing time accurately uses generator resources).  My EC during this time (with 2 fuel cells running) stayed basically constant though I'm not sure if that was because the two cells produce basically just enough to counter what the CSM uses, or if EC isn't used/generated when not focused on a craft.
After rendezvous, I transferred the crew, science and remaining resources over to the CSM, then jettisoned the LEM and did a burn to put me on course for a return to Earth.
From here I transmitted some of the science I had collected while on the moon (picked a spot close to a few biomes and had my kerbals walk to collect more science) so for the next 6h or so I ran on all three fuel cells.  This allowed me to completely recharge the CSM's batteries (they were down to about 80% from before the LEM landed).  From there I went back to running on 2 fuel cells.
As mentioned before, it seems like 2 fuel cells is pretty close to being right on the money.  I advanced time to 10000x for around 3.5 days and EC reported as (-17.90) during that time.  Not sure how accurate EC usages/generation is at that time advance but even with that small draw, there's plenty of battery available.
I did, by the end of the flight, notice I had made an earlier math error.  The LOX-O2 generator uses 3 times as much LOX as I had originally calculated for (I left out the "conversionRate" value).  So with the default 265LOX/360LH2 (the craft file has a SM with less LH2 than the part normally gets so I actually ran our of LH2 a bit early) you can run on 2 fuel cells for about 14d3.8h (excluding boil off) and on 3 fuel cells for around 9d19.76h.  It may be necessary to slightly reduce the rate of LH2 and LOX consumed by the generators even though two fuel cells are using slightly less then the current single cell setup.  Either that or maybe add about 10% more LOX/LH2.

After 12d,13:51:13, I finally ran out of LH2.  This effectively shut down my fuel cells though my LOX-O2 generator has a few more hours since I ran low on LH2.  I have about 14.5h worth of battery power left but I'm still 17h47m from my 50km Periapsis.  Admittedly, I did run on 3 cells for about 1/3rd of the flight and had I done the math correctly from the start, I could have managed my power a bit better.

stratochief66:
Alright, so the changes in your PR should be sufficient for now. Part of the issue with LH2 resources is that they evaporate a bit over 14 days, so that further complicates estimations on how much you should bring.

I will review things like LH2 consumption after everything is up and running for 1.0.5, because boiloff will probably be a bit different then, and possibly need adjustment again.

ctiberious:
I should add, I did manage to land.  I just turned off the battery on the CM for about 4 hours, then turned it back on.  This brought me up to around 330 CO2 but it allowed me to stretch my available EC so I could finish reentry.  Ended up with a bit over 18h of O2, nearly 2 days of Food and over 13 days of Water.  All in all, a good flight.  :)

PR #657, opened 2015-11-08
stock decoupler

winged7:


stratochief66:
@winged7 

Could you please provide at least a short note to go with each PR?

For this one, do you just judge that the current state of the stock decoupler is a reasonable scale and mass to be RSSRO compatible?

winged7:
Yes, stats of the stock decoupler are comparable to the other decouplers already supported by RO.

PR #662, opened 2015-11-10
Space Shuttle  - Mike NZ

winged7:
SRB should be surface attachable exactly like stock SRB but I don't know how to make it, someone could help me? Now it is very hard, almost impossible to attach.

Also RS-25 and OMS require Real Plume config.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CSS/CSS_Mike_NZ
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CSS/CSS_Mike_NZ
@@ -440,7 +440,7 @@
 	@rescaleFactor = 2.4
 	@maxTemp = 1973.15
 	%manufacturer = Aerojet
-	@mass = 0.2
+	@mass = 0.118
 	@description = Aerojet AJ10-190 as found on the SpaceShuttle Orbital Manoeuvring System (OMS)
  	@MODULE[ModuleEngines*]
 	{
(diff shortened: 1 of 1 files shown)


FFCJoseEduardo:
you need to find this line attachRules = 1,0,1,1,0 and change it to 1,1,1,1,0

however I would suggest to increase the node size instead since it needs to be attached to the decoupler, and if you try surface attaching you either will need to use a stock decoupler or will attach to the tank, as it will be even harder to attach to the decoupler from the pack

stratochief66:
@winged7 

Do you want to work on the SRB attachment and engine Plumes in this PR, or will you open another one for that?

winged7:
Wait. I need to change attachRules and check how it works in my game.

Regarding Real Plume: I have seriously no idea how to make proper config so someone else should do it.

winged7:
@FFCJoseEduardo  I think that it's better to make them workable with stock decouplers.

stratochief66:
@winged7 It isn't too hard, come by the IRC sometime and I can walk you through it.

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RealPlume_Configs/SSTU/SSTU_ShipCore_B_ENG4.cfg#L1-L36

For the RS-25 you can basically just copy the code from this, throw it into the proper place and file name, and overwrite with the proper part name and replace the transformName, and that is more than good enough.

winged7:
Ok, now it is ready to merge.

As to Real Plume I will try to make configs, but later, not today.

stratochief66:
Gotcha. Thanks again!

PR #665, opened 2015-11-10
Update to scaling code for ISS and Mir

FFCJoseEduardo:
Changed the code to avoid the rescaling bug, please do not merge as this is yet to be tested

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/BobCat/RO_Mir.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/BobCat/RO_Mir.cfg
@@ -1,7 +1,14 @@
 @PART[Mir_Kristall]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	@rescaleFactor = 1.4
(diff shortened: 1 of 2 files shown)


Niemand303:
Yay! I wanted to do that today, but you beat me. :)
Thanks!

FFCJoseEduardo:
If you still want to work on this, Rassvet/Mir Docking Module feels too chubby for me, so I could leave that to you :P

FFCJoseEduardo:
@Niemand303 I got full Mir Complex working with this patch, plus FGB, Node 2 and MPLM, Unity had the nodes working regarding orientation, but Unity and Destiny modules are messed up with the scaling, the others I couldn't test because I removed some modules from my setup

also worth noting: Pirs/Poisk SM decoupler and HTV doesn't seem to be RO compatible

and one idea, if you want of course, Destiny and Columbus doesn't have modules in them, so adding a lab to them through a MM patch could work, but JEM has an animation module from what I remember, so adding a lab to it might be more complicated

PR #666, opened 2015-11-11
Update RO_FASA_Saturn.cfg

ctiberious:
Historically, once the Apollo CSM and LEM have left the Saturn IVB stage behind (at least on the mid and late Apollo flights), Houston was able to remotely control the stage to alter it's trajectory so that it could impact the lunar surface.  This change adds remote control capability and antenna (the same Omni that the CM has) to the Saturn Instrument Unit to allow for this capability.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
@@ -107,6 +107,28 @@
 		@maxAmount = 36450
 	}
 }
+@PART[FASAApolloIU]:NEEDS[RemoteTech]:FOR[RealismOverhaul]
+{
+    MODULE
+    {
+        name = ModuleSPU
+    }
+
+    MODULE
+    {
+        name = ModuleRTAntennaPassive
+        TechRequired = start
+
+        OmniRange = 2000000
+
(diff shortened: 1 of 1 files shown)


stratochief66:
@ctiberious 

That seems reasonable. I'm sure I thought about that, but never got around to it.

Did you test this in game? How much power does controlling end up consuming/how much powered life does the spent S-IVb get after it is jettisoned?

ctiberious:
I tested a Saturn 1B flight and was able to deorbit the SIVB stage as expected.

The Omni antenna config I used is directly from the FASAApollo_CM config which only uses electricity when transmitting data.  Presumably the power cost for the antenna is assumed to be built into the cost of the command module.  Since the Instrument Unit isn't actually using any extra power, you should have enough electricity (just counting the IU and S-IVB parts) for just under 48 hours of flight time.  Can't really guess as to how much power you might have left after jettisoning the spent stage since that will depend on when you turn your fuel cells on and whether you pull any charge out of the stage before you jettison it.  But unless you intentionally take all the charge before staging, you should have at least enough power to make a single course correction (for lunar impact or whatever else you wanted to do with the spent stage).

NathanKell:
@stratochief66 @ctiberious dang, I thought someone went through and added SPU to all command parts. Apparently this got missed. That should definitely be done. @batuhantasli IIRC that was you? Or am I thinking of someone else?

PR #668, opened 2015-11-11
Mk2LightningCockpit

winged7:
Fictional cockpit inspired by F-22 and F-35


NathanKell:
If this is supposed to be--as its name and art implies--an F-22 cockpit, then it probably shouldn't have spaceplane level thermal shielding. Otherwise looks good to me!

winged7:
So how much max temp? 800?

NathanKell:
I'd say 800 is reasonable, yeah. Especially since it has those intakes built in, no way it'll survive a reentry or even X-15 style flight. 800 means short M=3 sprints and long M=2 flight so that should be fine.

PR #672, opened 2015-11-12
SSTU Lander pods rescale fix

FFCJoseEduardo:
ATTN: needs to be tested and probably improved, do not merge yet


FFCJoseEduardo:
They did not show any undesired behaviour during testing, these are ready to be merged

stratochief66:
Awesome, I will give them a quick test and merge soon.

stratochief66:
One thing I noticed that should be an easy fix for us, the two docking ports are slightly different sizes, perhaps the smaller one (with chutes) still has my old smaller scale.

stratochief66:
The current SSTU decoupler behaviour is pretty laughable, I'm curious what we can do to make it work properly though. 

For me, if nothing is attached to it, it spins away and upwards rapidly. If I have the RSRMV attached, they instantly explode on decoupling...

PR #688, opened 2015-11-19
Minor changes to SSTU

FFCJoseEduardo:
just changed the default radius for the tweakable decoupler to 5m wide (but it's still tweakable) and edited the SM node attach point


FFCJoseEduardo:
got this PR tested and working, Orion Service Module has a node for ICPS on by default, and the correct order to put parts is: OrionSM > Stack decoupler (5m one from SSTU, which with this update now has a default diameter of 5m, but still tweakable) > ICPS (up to the user if he wants to attach to the top most bottom or a lower one, lower one recommended)

If someone wants to make Block 1B he needs to do the following: Orion SM > right click on it and activate EUS node (recommended to deactivate the ICPS node) > put the SSTU interstage under it (couldn't make it to come configured as Block 1B by default, so it is up to the user to tweak it to look like it) > attach the EUS to the bottom of the interstage (decoupler not needed, as interstage works just as Apollo)

PR #674, opened 2015-11-13
Fuel crossfeed, CMES patches & Lionhead solar panels

PhineasFreak:
* [RO] Import CrossfeedEnabler patches into RO.
* [CMES] Add requirement for Helium pressurization for Altair Ascent module & AJ10-190 engine.
* [LH] Add RO support for the Lionhead Aerospace circular solar panels.

PR #689, opened 2015-11-19
Fuel cross feed

PhineasFreak:
Originally from pull request #674.

* Add fuel cross feed capability for radial tanks.
* Structural pylon can toggle it's cross feed capability on will.

### DIFF
--- a/GameData/RealismOverhaul/RO_FuelCrossFeed.cfg
+++ b/GameData/RealismOverhaul/RO_FuelCrossFeed.cfg
@@ -1,0 +1,231 @@
+//  ==================================================
+//  Cross feed enabler.
+//  Originally part of the CrossfeedEnabler mod by NathanKell.
+//  Powered by ialdabaoth's ModuleManager.
+//  ==================================================
+
+
+//  ==================================================
+//  Squad fuel tanks.
+//  ==================================================
+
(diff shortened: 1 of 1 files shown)


NathanKell:
um, not really seeing the point of this? You're adding the crossfeed toggle to tanks, rather than decouplers--the tanks should always be crossfeeding (and do, as of 1.0.5 and our Physics.cfg tweak). Also, the radial decouplers already have the module added in stock 1.0.5, so you don't need to worry about changing them.

What would be good is if you applied the crossfeed module to other mods' radial decouplers (use the stock radial decoupler cfg as a guide).

NathanKell:
Ah, the structural bit. Yes, that bit should be kept, missed that.

PhineasFreak:
Sorry, i missed the physics patch change...

I could write a patch for all decouplers, not only the radials, and place it under the RO "global config". Then, all decouplers will be "cross feed disabled" until the user changes it from the GUI.

If you agree, close this PR. I will make the necessary changes to include the above changes in the "RealismOverhaul_Global_Config" file.

NathanKell:
Hmm. I worry that simply adding it to all things with decouplers may lead to some unexpected bad situations, but I guess we can see how it works, and it's a good thing to try and see, rather than not try. So yeah. :)

PR #694, opened 2015-11-20
Changes to Centaur

ctiberious:
Adjusted the dimensions and weight on the Centaur D/D1, D2 and D3 tanks.  I've used values as presented in astronautix  for the sake of consistency.  I did compare values to what is presented in b14643.de and my results are comparable.  But I did notice that b14643.de uses the same basic statistics for similar, but not identical, stages.  For instance, there are five different versions of the Atlas II presented, each with their own slightly different Centaur stage (each uses a different RL10A engine set) but the life off weights and propellant weights never change.  Seems unlikely that the RL10A-3-3A had the exact same weight as the RL10A-4-1N.  In the end, I tried to be consistent but also went with lower weight values on the assumption that minor changes could be dealt with by including a small amount of leadbalast.
The final tank height includes the tank and engines.
The final tank weight includes the tank, engines, payload fairing base and 0.03t worth of RCS thrusters.

I've also added a configuration for the Centaur D5 (used with the Atlas V) and the Centaur T (used with the Titan IV).  Information is based on the already existing Centaur tanks with weights and dimensions coming from astronautix and b14643.de

Finally, I've included a RCS thrusters for centaur roll control.  From various wiki sites, it seems that a standard Centaur setup included 8 thrusters for roll and attitude control and 4 slightly more powerful thrusters for ullage.  The ullage thrusters can easily be recreated using existing Attitude Jets but I've always run into problems properly positioning single jets for roll/attitude control.  To resolve this, I've used the FASA_Gemini_RCS_Thrusters as a model for small, Hydrazine powered attitude thrusters for the Centaur.  I've reduced the weight of the original (0.015 down to 0.006) since the centaur versions are weaker but I have no documentation for how much they should actually weigh.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Titan.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Titan.cfg
@@ -184,7 +184,7 @@
 	@title = Centaur A/B/C/D/D1
 	@description = Tank and thrust structure for the Centaur A/B/C/D/D1 models. Add two RL10A series engines to complete the stage.
 	@attachRules = 1,0,1,1,0
-	@mass = 1.726
+	@mass = 1.125	// From Astronautix, based on the Centaur I or Centaur D-1B.  Add leadbalast to duplicate older versions of this model Centaur
 	!RESOURCE[LiquidFuel]
 	{
 	}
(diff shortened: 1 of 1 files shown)


stratochief66:
Looks alright to me.
Can you add a *.craft file for something like the Titan - Centaur using FASA (no payload or Procedural weight payload) with your RCS jets so I can see exactly how you are thinking of using them? Since they are more or less necessary to properly ullage the stage I am also considering welding them onto the part.

ctiberious:
I've got a pair of craft files I can upload.  One for the Titan 3E and another for the Titan IVA.  Both are based on craft files I use in my live game.  However, I've edited both.  The primary, relevant, changes are that in my live game I use the "Interstage - w/ Decoupler [Procedural]" and "Payload Faiting - Base (Extended) [Procedural]", both from procedural fairings.  In the two craft files I've setup here, I've replaced those with "non RP-0 - Payload Launch Fairing Expanded Base [168"]" (FASA - both), "FB - FF3 Airstream protective Shell" (Stock - both), "Titan Decoupler [120"]" (FASA - Titan 3E) and "non RP-0 - FASA Dark Decoupler [Large]" (FASA - Titan IVA).  I've tested both craft files but not extensibly.  They should work as intended, however.

What I'm not sure of is how to upload them.  Is there some way to add the files to this PR?  Or should I create a new PR for each craft file and just have each refer back to this PR?

stratochief66:
If you use the desktop version of Github, it is fairly easy to add the files. However, if you don't, you could just create a new PR with them. Or, you could just add them to a branch (or branches) of your own Repo and link them here so I can take a look at them

ctiberious:
Unfortunately I don't use Github desktop (work on too many different computers) so I just added each craft file as a separate PR (#708 and #709)

stratochief66:
Thanks, craft files for testing. Much appreciated.

Those work pretty slickly! I might end up using them on non-centaur upper stages that need ullage as well.

PR #700, opened 2015-11-21
Create RO_InfernalRobotics.cfg

winged7:
Support for IR parts. They don't require any changes.

### DIFF
--- a/GameData/RealismOverhaul/REWORK/RO_InfernalRobotics.cfg
+++ b/GameData/RealismOverhaul/REWORK/RO_InfernalRobotics.cfg
@@ -1,0 +1,56 @@
+@PART[AdjustableRailScaleable]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = True
+}
+@PART[IRPistonScaleable]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = True
+}
+@PART[GantryLargeScaleable]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = True
+}
+@PART[GantryLargeScaleableVariant]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = True
+}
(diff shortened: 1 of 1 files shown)


NathanKell:
They don't? Aren't they rather heavy for what they do?

winged7:
I think, it's ok.

stratochief66:
@winged7 

Could you quote some of the masses of the IR parts? I haven't personally used the mod. That would give us an idea if we should suggested the mass staying the same, or being halved, or a guideline somewhere along those lines.

winged7:
AdjustableRailScaleable  mass = 0.2
Docking Washer mass = 0.025
Docking Washer Standard (Free Moving) mass = 0.025
Gantry Rail  mass = 0.8 ( huge part, 10meter long)
Gantry Rail Variant mass = 0.8
Closed Powered Hinge mass = 0.1
Open Powered Hinge mass = 0.1
Powered Hinge mass = 0.1
Powered Hinge 90 degrees mass = 0.1
Hydraulic Cylinder mass = 0.1
IR Rotatron mass = 0.1
IR Rotatron Mk2 mass = 0.1
VTOL Rotor Mk1 mass = 0.1
Telescopic Piston mass = 0.2

NathanKell:
Yeah, those sound way too high for probe use...100kg is like half the mass of Ranger, where the hinges would be used for the foldout solar panels...

winged7:
So how much they need to be lighter? 2x , 3x or 5x? :>

NathanKell:
Depends on their size I guess. Certainly I wouldn't expect hinges sized for the RO Ranger cores to be heavier than about 5kg each, but it's worth a half-hour's google into hinge weights, probably, and also depends on how large the hinges default to (since IIRC Infernal Robotics uses TweakScale to allow resizing the hinges?)

winged7:
yep, it uses tweakscale.

winged7:
I will back to this topic later, closed for now.

PR #703, opened 2015-11-22
Squad fixes & updates

PhineasFreak:
* Update RTG and NERVA to use the ModuleResourceConverter module
(operate even if vessel is on rails).
* Fix the ASL ISP of the Ion engine.
* Fix the model path for the Toroidal Aerospike engine.
* Add RO support for the radiator fin and overhaul all the other
radiator configurations (some WIP).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Electrical.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Electrical.cfg
@@ -124,25 +124,25 @@
 
     @PART[rtg]:FOR[RealismOverhaul]
     {
-	    %RSSROConfig = true
+        %RSSROConfig = true
 
-	    MODEL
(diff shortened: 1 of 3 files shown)


stratochief66:
https://github.com/PhineasFreak/RealismOverhaul/commit/e4f23f7d014faa6e0d7b7e834fcc0086ad9ebbe9#diff-52a3305ac0b3fa2a1b798703f5b7439bL621

The old version is how it would look if the stock part didn't have a key 1 xxx value. Does it now have one?

PhineasFreak:
Yes. The following is taken from "ionEngine.cfg" file:

>atmosphereCurve
>{
>	key = 0 4200
>	key = 1 100
>	key = 1.2 0.001
}

stratochief66:
Just as with 704, please do a formatting run when you can.

https://github.com/KSP-RO/RealismOverhaul/pull/704

@PhineasFreak

PR #706, opened 2015-11-22
Update BD Atlas V parts

PhineasFreak:
* Add support for the new engine thermal changes.
* Fix AJ-60A and RD-180 categories.
* Fix  AJ-60A and RD-180 ASL and VAC ISP values.
* Fix  RD-180 minimum thrust value.
* Fix RD-180 dry mass value.
* Remove redundant IGNITOR_RESOURCE modules.
* Add external reference sources.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Atlas_V.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Atlas_V.cfg
@@ -50,8 +50,8 @@
 
 		@MODULE[ModuleEngines*]
 		{
-			%minThrust	    = 1270.00
-			%maxThrust	    = 1270.00
+			%minThrust	    = 1270
+			%maxThrust	    = 1270
 			%heatProduction = 100
 			!fxOffset	    = NULL
 
@@ -64,8 +64,8 @@
 
 			@atmosphereCurve
 			{
-				@key,0 = 0 279.00
-				@key,1 = 1 245.00
+				@key,0 = 0 279
+				@key,1 = 1 245
 			}
 		}
 
@@ -87,8 +87,8 @@
 			CONFIG
 			{
 				name		   = AJ-60A
-				minThrust	   = 1270.00
-				maxThrust	   = 1270.00
+				minThrust	   = 1270
+				maxThrust	   = 1270
 				heatProduction = 100
 
 				PROPELLANT
@@ -100,8 +100,8 @@
 
 				atmosphereCurve
 				{
-					key = 0 279.00
-					key = 1 245.00
+					key = 0 279
+					key = 1 245
 				}
 			}
 		}
@@ -163,7 +163,7 @@
 		@MODULE[ModuleEngines*]
 		{
 			@minThrust		= 1951.44
-			@maxThrust		= 4152.00
+			@maxThrust		= 4152
 			@heatProduction	= 100
 			%ullage		 	= true
 			%pressureFed 	= false
@@ -209,8 +209,8 @@
 
             @atmosphereCurve
             {
-                @key,0 = 337.80
-                @key,1 = 311.30
+                @key,0 = 337.8
+                @key,1 = 311.3
             }
 		}
 
@@ -231,7 +231,7 @@
 			{
 				name		   = RD-180
 				minThrust	   = 1951.44
-				maxThrust	   = 4152.00
+				maxThrust	   = 4152
 				heatProduction = 100
 
 				PROPELLANT
@@ -250,8 +250,8 @@
 
 				atmosphereCurve
 				{
-					key = 0 337.80
-					key = 1 311.30
+					key = 0 337.8
+					key = 1 311.3
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


stratochief66:
Closed for the same reasons as 711, please resubmit with new formatting. 

PhineasFreak:
I think that this will do the job. No formatting yet, it will be done after @stratochief66 has assessed the changes.

stratochief66:
Some of the changes in formatting don't seem to be necessary (ie. adding two post-decimal values for ISP Keys, changing the case of True/true)

Your RD-180 Gross Mass and @mass don't agree.

@NathanKell ?

PhineasFreak:
That is because some of these are automatically written by a small script. I will remove it now.

The gross mass vs the actual part gross mass is because the sources only mention the engine mass outside the Atlas V engine compartment. The BahamutoD model already includes the engine fairing and the LOX line. Probably 1 ton is a bit too heavy but it is a good ballpark (i took into account not only replicas but the possibility to used on something else).

stratochief66:
@PhineasFreak script generated, gotcha.

please let me know when you've removed the excess figs

PhineasFreak:
Ok, i removed them. Tell me if i missed something or anything else that you want me to change!

stratochief66:
@PhineasFreak 

Line 5 still seems to have some meaningless figs (I doubt the mass of the part is known precisely to the hundredth of a kilogram) Look through the PR yourself for similar examples.

I believe both true and True work in cases like line 62 ( @NathanKell confirm? ) Please don't change it for change sake.

Do you want to do the formatting fixes now as part of this PR?

PhineasFreak:
These figures are a standard template that i have for commenting the part (meaning that it gets copy - pasted and adjusted accordingly to the part function). Does the formatting really matter?

stratochief66:
@PhineasFreak 

I don't consider false precision to be a 'formatting' issue, but no, it isn't mission critical that you change it.

Changing the case of between true/True actually appears to have an impact here, as NK stated elsewhere.

If you want a template, most FASA or Squad RO configs are well formatted. Example:
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg

Highlights of style/formatting things that are still wrong here:

* Please revert back to %DrawGauge = True
* Change "%minThrust	    = 1270" to '%minThrust = 1270" , This is the formatting standard.
* All lines in a block should be at the same level of indentation, for example please fix "        %heatConductivity = 0.060"

And please review your other open PRs to be consistent with this formatting.

PhineasFreak:
@stratochief66 As i mentioned, the only thing that wasn't corrected is the level of indentation, since that will make GitHub freak out. I will do it after the final review so that you can easily see the changes.

stratochief66:
@PhineasFreak 

Assessment complete, excellent fixes. :)

You can safely change the indentation now, then it should be ready to merge.

PhineasFreak:
I found another ability of Notepad++: to automatically convert on - the - fly a tab press into spaces. That means no more tab --> space PR corrections!

stratochief66:
I sent you an example to follow, note how all the new part lines (ie.   @PART[BahaRd180]:FOR[RealismOverhaul]:NEEDS[!RftS,!RealFuels_StockEngines])

have no spaces or tabs before them. A new part block should start with no indentation.

PhineasFreak:
If you have reviewed the changes, tell me so that i can fix the spacing and trailing zeros.

stratochief66:
yep... go ahead

PhineasFreak:
Done!

stratochief66:
See the suggestions on the other PR.

Also, note the proper way to format the changing of the ISP data:
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg#L173-L174

PhineasFreak:
What do you mean about the ISP curves? I do not understand it.

stratochief66:
If you don't understand it, then I strongly suggest you copy the syntax of the example I posted.

The ISP is a Key/Value set, and the Module will LERP (linear interpolation) between the two values as the atmosphere goes from sea level (1) to a vacuum (0)


You have the right syntax here:
https://github.com/KSP-RO/RealismOverhaul/pull/706/files#diff-9e09d274695f74cf7b39245dda4ab6aaR63

But wrong here where you are replacing a key/value set:
https://github.com/KSP-RO/RealismOverhaul/pull/706/files#diff-9e09d274695f74cf7b39245dda4ab6aaR205

PhineasFreak:
Oh, you mean the actual ISP value. Isn't this a valid value?

stratochief66:
No. I explicitly state that the syntax is the issue. Please look at both examples I linked and see the difference.

PhineasFreak:
Oh, DAT typo....sorry...

stratochief66:
Thanks. I hope all the back and forth here will help us both with PRs in the future :)

Ready to merge? It all looks good to me, both form and function.

PhineasFreak:
The problem was that i was fixing as i was learning. If you see later PRs, we had no problem there. And working from a cellphone certainly does not help the things!

Yes, if you have reviewed it you can merge it!

PhineasFreak:
Thank you!

PR #707, opened 2015-11-22
SNAP-19 update

PhineasFreak:
* Replace ModuleGenerator with ModuleResourceConverter (operate even if
vessel is on rails).
* Add core heat simulation.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/AIES/RO_AIES_SolarPanels.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/AIES/RO_AIES_SolarPanels.cfg
@@ -1,27 +1,109 @@
+//  ==================================================
+//  SNAP-19 Radioisotope Thermoelectric Generator (RTG).
+
+//  Power generated is the average of the upper (42.5 W) and lower (38 W) limits.
+
+//  Dimensions: 0.168 m x 0.282 m
+//  Gross Mass: 13.60 Kg
+
+//  Source 1: http://www.osti.gov/accomplishments/documents/fullText/ACC0166.pdf
+//  Source 2: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19680022615.pdf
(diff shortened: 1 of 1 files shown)


stratochief66:
Closed for the same reasons as 711, please resubmit with new formatting. 

PhineasFreak:
Ok, this also should be good.

stratochief66:
@NathanKell 

"%RSSROConfig = True" if fine, I'm not 100% sure if it works both ways. (NK?)

In my experience I always see the "!mesh" before the scaling lines, is there a reason for this NK or just tradition?

NK also will have a better idea of how this ModuleResourceConverter and ModuleCoreHeat will/should behave, so I will leave the rest of this PR to him.

PhineasFreak:
As i mentioned in the BD_Atlas PR, some of these little changes are made automatically. For the mesh & model references, it is a preference to keep them close together so that they are easy to modify while making a part configuration. I mostly follow the config structure of the stock KSP part configurations.

For the converter & core heat modules i was searching for available references on the forums and the Wiki but i did not find something. I hope that @NathanKell will be able to provide some examples for these modules.

NathanKell:
It's best to use True when assigning a boolean field. However, RSSROConfig _requires_ capital-T True, IIRC, since it is used in a (case-sensitive) MM HAS block, rather than parsed by KSP itself.

As to ModuleResourceConverter: If at all possible this should be **avoided**. It is way, way, way slower than ModuleGenerator, and I believe it is also slower than the TAC converters.

NathanKell:
Further, while scripts are awesome (really! They help lots), please don't add 'default' position/rotation lines to MODEL patches. That can mess up later MODEL patches which expect the lines to not exist (and then add a duplicate line, since one already existed).

PhineasFreak:
A lot of good points here! And i did not know that the boolean flag for RSSROConfig was case - sensitive.

Now, since the major change is not considered as critical, feel free to lock the PR again.

PR #708, opened 2015-11-22
Create FASA Titan 3E.craft

ctiberious:
This craft file is a representation of the Titan 3E using FASA parts.  It's based on the changes proposed in PR #694.

I've used the "FASAExplorerProbe" as the payload so hopefully it would be a simple matter for players to swap that out with whatever satellite they have created.

The Centaur stage is based on the Centaur D1 (or D/E depending on source).  It includes hydrazine powered RCS and Ullage thrusters.  I've only set the thrusters to tech 0 to closely match the thrust ratings I was able to find for the actual Centaur.  Including the fairing base, I've set the fuels (including some leadbalast) to match the wet and dry weight of that same Centaur as seen on the astronautix website.  Engine is the RL10A-3-3.  This is effectively the same Centaur used on the Atlas SLV-3C and early Atlas G.

### DIFF
--- a/Ships/VAB/FASA Titan 3E.craft
+++ b/Ships/VAB/FASA Titan 3E.craft
@@ -1,0 +1,12698 @@
+ship = FASA Titan 3E
+version = 1.0.4
+description = 
+type = VAB
+size = 14.77783,47.43388,7.66025
+PART
+{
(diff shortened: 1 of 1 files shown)


stratochief66:
Done using for testing, thanks!

PR #709, opened 2015-11-22
Create FASA Titan 4A.craft

ctiberious:
This craft file is a representation of the Titan IVA using FASA parts.  It's based on the changes proposed in PR #694.

I've used the "FASAExplorerProbe" as the payload so hopefully it would be a simple matter for players to swap that out with whatever satellite they have created.

The Centaur stage is based on the Centaur D1-T (or G depending on source).  It includes hydrazine powered RCS and Ullage thrusters.  I've only set the thrusters to tech 0 to closely match the thrust ratings I was able to find for the actual Centaur.  Including the fairing base, I've set the fuels to match the dry weight of that same Centaur as seen on the astronautix website.  The wet weight is about 100kg heavier then it should be but the only way I could match the wet and dry weights, and still have the intended burn time, was to only include about 20 units of Hydrazine which didn't feel like enough so I took a little creative license and included a bit more.  Engine is the RL10A-3-3A.  There was also a version (identical name) that used the more powerful RL10A-4 engine but it was used on the Titan IVB which I could not fully recreate because I don't see an option for the more powerful Hercules USRM SRBs.

### DIFF
--- a/Ships/VAB/FASA Titan 4A.craft
+++ b/Ships/VAB/FASA Titan 4A.craft
@@ -1,0 +1,12688 @@
+ship = FASA Titan 4A
+version = 1.0.4
+description = 
+type = VAB
+size = 14.816,47.46601,7.668359
+PART
+{
+	part = FASAExplorerProbe_4293735882
+	partName = Part
+	pos = 3.099907E-06,45.01504,7.543713E-06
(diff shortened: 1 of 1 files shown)


stratochief66:
Done using for testing, thanks!

PR #710, opened 2015-11-22
Blackleg industries

PhineasFreak:
* Support for the engine thermal changes in KSP 1.0.5.
* Fix wrong model reference name for the BCP-2000 service module.
* Add plume for the R-4D-11 HiPAT engine.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/BlacklegIndustries/RO_Blackind_Engine.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/BlacklegIndustries/RO_Blackind_Engine.cfg
@@ -1,124 +1,120 @@
-//	==================================================
-//	Remove extra parts.
-//	==================================================
+//  ==================================================
+//  Remove extra parts.
+//  ==================================================
 
-	!PART[satbusengineb]:FOR[RealismOverhaul]{}
+!PART[satbusengineb]:FOR[RealismOverhaul]{}
 
-//	==================================================
-//	R-4D engine.
+//  ==================================================
+//  R-4D engine.
 
-//	Dimensions: 0.130 x 0.220 m
-//	Gross Mass: 5.20 Kg
-//	Throttle Range: N/A
-//	Burn Time: 3600 s
-//	O/F Ratio: 1.65
+//  Dimensions: 0.130 x 0.220 m
+//  Gross Mass: 5.2 Kg
+//  Throttle Range: N/A
+//  Burn Time: 3600 s
+//  O/F Ratio: 1.65
 
-//	Source: https://www.rocket.com/files/aerojet/documents/Capabilities/PDFs/Bipropellant%20Data%20Sheets.pdf
-//	==================================================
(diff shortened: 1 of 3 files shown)


stratochief66:
closes for the same reasons as 711, please resubmit with new formatting. 

PhineasFreak:
As with the other PRs, this has only the base changes for now. I will fix the wrong formatting later.

The plume config is a new file that conforms with the indentation standards.

PhineasFreak:
Also @stratochief66 , could you review it? It is the same situation as the Atlas V parts (meaning "fix configs now, formatting after").

stratochief66:
Have you tested the plume in game?

PhineasFreak:
Yes, in fact every plume configuration that i make must always pass from my "Quality Assurance", that is, to test them in - game. Otherwise, i do not include them.

stratochief66:
Alright looks fine. To the formatting now (try to do proper formatting on your first PR going forward please) 

* No spaces before @PART
* consistent indentation within a block
* Please try to avoid false precision, even in a comment line. ex. Gross Mass: 20.00 Kg

PhineasFreak:
Got it!

Since that you have reviewed the changes, should i fix the remaining formatting issues now, before merging the changes?

stratochief66:
Yep, please do.

PhineasFreak:
Done!

stratochief66:
Getting there.

https://github.com/PhineasFreak/RealismOverhaul/commit/3ce105273c07b1005149847f847d0df8db6917dd#diff-15dd6d6fdf137457e7bce971261e1eb0R33

extra spaces around the = , false precision in the numbers. Node size (last value) has to have a number.

Again, review the suggested templated; https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg

If you do something differently than the the other RO files do, you should have a good functional reason for doing it, so please follow the standards.

PhineasFreak:
I missed the first two but for the node size what do you mean? The last value is a number.

stratochief66:
The last value is the 'size', or strength of an attachment node. Heavier parts will have a larger number, therefore a 'larger' node, or a stronger attachment strength (more resistance to wobble). I don't know if I've seen a node in another config with a 0, so I'm not sure what exactly that would mean for the attachment strength, but a value of at least 1 is needed.

PhineasFreak:
No, size0 is a valid node size: http://wiki.kerbalspaceprogram.com/wiki/Parts/Engine/liquidEngine48-7S/liquidEngine48-7S.cfg

stratochief66:
Hmm, fair enough. I'm curious what effect that has on the strength of the connection between the parts.

@PhineasFreak Please make the other suggested fixes.

@NathanKell what effect does a size 0 node have compared to a size 1?

PhineasFreak:
For the specific part, smaller node size: since it is a very small engine, larger nodes tend to obscure the view when attaching the part.

stratochief66:
Hey, sorry for the delay. I had been waiting for a response from NK about node size specifics. It all looks good.

PR #711, opened 2015-11-22
Update LR-87 LH2 engine

PhineasFreak:
* Remove engine FX (it is taken care by the RealPlume configuration).
* Add support for the engine thermal changes in KSP 1.0.5.
* Remove deprecated EngineIgnitor modules and replace them with the
RealFuels - compatible configuration.
* Remove alternator module and ElectricCharge storage (leftovers by the template engine config).

### DIFF
--- a/GameData/RealismOverhaul/Parts/Engines/LR87LH2.cfg
+++ b/GameData/RealismOverhaul/Parts/Engines/LR87LH2.cfg
@@ -1,306 +1,256 @@
-+PART[liquidEngine2]:FIRST
-{
-	@name = LR87LH2Vac
-}
-@PART[LR87LH2Vac]:FOR[RealismOverhaul]
-{
-	!mesh = DEL
-	MODEL
-	{
-		model = RealismOverhaul/Models/LR-91eng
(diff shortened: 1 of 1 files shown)


stratochief66:
@PhineasFreak 

Apologies, but the formatting of this PR requires us to reject it.

Could you try using a different text or code editor to make your changes? The formatting of parts you haven't altered shouldn't be changed, and preferably code including your changes should abide by the style of the rest of the file.

The changes themselves appear promising, but because the PR nukes the formatting of the entire file I can't zero in on all the changes you have made to properly review them. Additionally, the nuking of the file's formatting removes one of the serious benefits of Github, which is code tracking which allows us to see when a line was last changed, by whom and in which PR.

@NathanKell et al, can you more precisely define the desired formatting. I see it as pretty similar to Python's PEP 0008.

PhineasFreak:
Well, i rewrote every single one of the configurations that are included in these PRs from scratch, so i was expecting it to be different that the original files. I just copied over the values.

Also, i cannot fix the formatting of the "broken" old files without GitHub complaining. So, what can i do?

PR #713, opened 2015-11-23
Squad Fixes Formatting

PhineasFreak:
@stratochief66 Will this work? Although i noticed that in some places GitHub decided to again replace code blocks instead of single lines (the only change here is the formatting, no other alterations are present).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Electrical.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Electrical.cfg
@@ -132,19 +132,19 @@
             scale = 1.000, 0.777228, 1.000
         }
 
-        !mesh          = DELETE
-        @scale         = 0.777228
+        !mesh = DELETE
+        @scale = 0.777228
         %rescaleFactor = 1.000
 
-        @node_stack_top    = 0.000,  0.3268025, 0.000, 0.000,  1.000, 0.000, 0
+        @node_stack_top = 0.000,  0.3268025, 0.000, 0.000,  1.000, 0.000, 0
         @node_stack_bottom = 0.000, -0.3268025, 0.000, 0.000, -1.000, 0.000, 0
-        @node_attach       = 0.000, -0.3268025, 0.000, 0.000, -1.000, 0.000
+        @node_attach = 0.000, -0.3268025, 0.000, 0.000, -1.000, 0.000
 
-        @title        = Multi-Hundred Watt RTG
+        @title = Multi-Hundred Watt RTG
         @manufacturer = NASA JPL & DOE
-        @description  = The Multihundred-Watt radioisotope thermoelectric generators (MHW-RTG) as found on the Voyager spacecraft and two Lincoln Experimental Satellites.
+        @description = The Multihundred-Watt radioisotope thermoelectric generators (MHW-RTG) as found on the Voyager spacecraft and two Lincoln Experimental Satellites.
 
-        @mass          = 0.0332
+        @mass = 0.0332
         %fuelCrossFeed = false
 
         !MODULE[TweakScale]{}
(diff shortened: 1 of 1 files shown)


stratochief66:
Yes! The formatting here is nearly perfect. :)

The only remaining issue I see there is that from lines 125 to 193 have a four space indent that should be removed.

PhineasFreak:
Oh, i missed that one...damn!

Ok, but there is a bit that it's worrying me: as you see from lines 156 to 164 the code block was "replaced", even if it is exactly the same content. If i try to update the other files again, i think that it will think that it is new content and not updated (like the LR-87 file, i did not change almost nothing there, only a rewrite). What do you think?

stratochief66:
With the new changes, it looks good.

Yeah, I wish Github didn't consider a line with slightly different formatting as an entirely new line of code, but that is how it is. That is why it is important that we try our best to keep code formatted in a consistent style, both for ease of readability and to keep the GIthub beasts on our side.

Minor inconsistencies in formatting are fine (in my opinion). For instance, I use actual tabs while others use 4 spaces.

Github will act like you have replaced your code when you bring the formatting in line with the repo, but that is preferable to leaving the formatting as it is. That is one reason that a well chosen PR name or PR comment is important. If you name a format fixing PR similarly to this one it is clear in the history that the PR was just for format fixing. It gets -really- confusing when a PR includes both functional changes and formatting changes (like some of your previous PRs)

Thanks for the effort, are you ready for me to merge this PR?

PhineasFreak:
I also got this with the LR-87 file that i am just updating (rewriting...) and i even if i can understand why GitHub does that: a space can be literary code - breaking if it is inserted or removed.

From now on i will use this formatting (zero starting spacing, 4 space indentation and a comment block at the top for summaries & sources). For the files that were not formatted or they were formatted badly i will follow the plan that you mentioned: a PR that will have all the part module changes followed by a PR clearly indicating that it will change the formatting (if applicable).

Give me some time to update the rest of the Squad files. These will be only for formatting of course. Oh, and if you can, could you reopen the closed PRs so that i can push the changes there? Less clutter on the pull request list...

stratochief66:
Apologies about the LR-87 if I misread it.

Previous PRs open so that you can alter them there.

I believe Github is smart enough to recognize a line deletion as just a line deletion, and recognize that the rest of the file below there is the same, just shifted.

If you mean moving blocks of MM code around (like, moving a part that was mid file to the end) without changing the code itself, I would consider that formatting, yeah.

PhineasFreak:
Yes, i can see that the GitHub Desktop can "see" the modified lines as you mentioned.

What i meant is if it is valid to move a line like "%RSSROConfig = true" from say line 10 to line 2. Always in the constrains of the specific part configuration of course.

stratochief66:
About an example like %RSSROConfig = true

We don't have an ironclad template for where a line like that would be (similar examples for me include the precise ordering of early elements in a MM patch, such as the order of @mass , @maxTemp , etc.) If it is at the end of the file or somewhere silly and you want to move it to the beginning portion, then fair enough. 

But just shuffling MODULE blocks or line for shuffle sake, I would say 'don't do that'.

@NathanKell might be able to point you to a file that represents our preferred template?

PhineasFreak:
Ok. Thank you very much for your time that you spent reviewing these!

stratochief66:
No worry! I knew I would pay the karmic price of my formatting learning curve one day, and on the balance RO gets two more capable RO contributers for NathanKell et al's patience in teaching me.

PhineasFreak:
@stratochief66 Should i discard the changes since @NathanKell warned against using the ModuleResourceConverter?

stratochief66:
From my read of these commits, you are mostly fixing the formatting, not changing functionality.

I agree with NathanKell that other superior Modules exist for this sort of thing, but until TACLS is updated so that you/we can use it, I think this is an ok stand-in. Unless I missed something, this looks ready to merge.

PhineasFreak:
I was thinking about reverting the changes but since it's OK for now then merge it if you can.

PR #704, opened 2015-11-22
Lionhead Aerospace solar arrays

PhineasFreak:
* Add RO support for the Lionhead Aerospace circular solar arrays.

NOTE: The small arrays are modeled after the Phoenix & Mars InSight robotic exploration
platforms and the large arrays are modeled after the Orion CEV & 606 designs.

PR #714, opened 2015-11-23
LH_Solar_Panel text formatting

PhineasFreak:
Only for fixing the text formatting. No changes were made. As requested by @stratochief66 in #704.

Edit: a small change, i added a second meter indicator at the comment section.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/LionHeadAerospace/RO_LH_Solar_Panels.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/LionHeadAerospace/RO_LH_Solar_Panels.cfg
@@ -19,7 +19,7 @@
 
 @PART[LH_CircularPanel1]:FOR[RealismOverhaul]
 {
-    %RSSROConfig = true
+    %RSSROConfig = True
 
     !mesh = NULL
 
@@ -63,7 +63,7 @@
 
 @PART[LH_CircularPanel2]:FOR[RealismOverhaul]
 {
-    %RSSROConfig = true
+    %RSSROConfig = True
 
     !mesh = NULL
 
@@ -107,7 +107,7 @@
 
 @PART[LH_CircularPanel4]:FOR[RealismOverhaul]
 {
-    %RSSROConfig = true
+    %RSSROConfig = True
 
     !mesh = NULL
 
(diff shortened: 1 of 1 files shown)


stratochief66:
Looks good. Can you move the !mesh to before the MODEL portion though? I'm unclear on why/when it is necessary and @NathanKell hasn't yet had a chance to weigh in if it matters where that line sits.

PhineasFreak:
A !mesh is needed when the original part module was defined as "mesh = model.mu". These meshes are replaced with MODEL{} definitions (the only way to rescale the part with precision) and so we need to remove them from the part (i think if they are not removed it looks like a model inside a model). As with the part module parameters that you mentioned in #713, KSP doesn't care where it is placed. I have tested different placements and all behave the same.

Generally, none of the part configuration placement matters. Take a look at the "ModuleManager.ConfigCache" file that ModuleManager generates: it places the one - liners at the top and every other code block (such as MODEL{}, MODULE{} and RESOURCE{}) are placed at the bottom at a random way.

Personally, i put the mesh definitions there because it looks odd in other places.

NathanKell:
We generally do !mesh first, for clarity in reading the cfg--it signals early to the maintainer that the patch was for a part that used mesh. Then when the part updates to use MODEL itself (as happened to a few parts in 1.0.5) it's easy to see our patches are out of date.

PhineasFreak:
Since that i have been in a not - as - good position as i hoped with the formatting of the config lines, could you suggest a template that we can follow for future reference? @stratochief66 made a good point about using the Python PEP - 0008 guidelines, as mentioned in PR #711.

PhineasFreak:
Anything here that you want to be corrected?

stratochief66:
Doesn't this have to be capital T from what NathanKell said?

%RSSROConfig = true

From NathanKell:
It's best to use True when assigning a boolean field. However, RSSROConfig requires capital-T True, IIRC, since it is used in a (case-sensitive) MM HAS block, rather than parsed by KSP itself.

PhineasFreak:
Frak, i corrected it for the Atlas V parts and missed it here! Yes, give me 2 minutes...

PhineasFreak:
And done!

stratochief66:
Thanks!

PhineasFreak:
Thank you for your help & support for all these!

PR #718, opened 2015-11-27
VSR reaction wheel

winged7:
The same stats as for the smallest stock reaction wheel.


stratochief66:
That seems awfully.. boiler-plate-y. And, you gave it Tweakscale. Tweakscale isn't on the list of compatible mods with RO. I would double-check the forum, but it is currently down.

Having a reaction wheel in RO that was actually consistent with the properties of a real reaction wheel would be interesting, but I'm not entirely sure of the purpose since without Persistant Rotation a vessel loses any rotation on timewarp anyway.

PR #722, opened 2015-11-28
Universal storage update

PhineasFreak:
* Fix part module names.
* Remove decoupler configs (parts do not exist anymore).
* Add RO support for the radial water tank.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/UniversalStorage/RO_UniversalStorage.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/UniversalStorage/RO_UniversalStorage.cfg
@@ -10,7 +10,7 @@
 		note = (pressurized)
 	}
 }
-@PART[US_c_Hub_Quadcore]:FOR[RealismOverhaul]
+@PART[US_1C10_Wedge_Quadcore]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
 	!MODULE[TweakScale]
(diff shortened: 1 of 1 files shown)


stratochief66:
Looks pretty clear, thanks for the bullet points describing what is done :)

For the US_Radial_R30_Water part, what does it weigh? I noticed some of the other parts seemed a little heavy for what they do. For instance, US_1R360_Wedge_Water is a 50kg contained that holds 50-100kg of water.


<You may not have input on the below, but since I've never used this mod, I thought you might know more than me>
In reviewing the code for this, I see both MODULE[TacGenericConverter] and MODULE[REGO_ModuleResourceConverter], do you know why both are used, and where REGO_ModuleResourceConverter comes from (what mod?)

I'm ready to merge this as is, but any answers to my questions would be much appreciated. :)

PhineasFreak:
If you want the original weight as defined by the US mod, it is 26 Kg dry, 36 wet. The same dry weight is used for the other radial tanks. Judging from the fact that it keeps 75 Kg of water (consistent with the volume of the other tanks), we might need to increase the dry weight a bit for these tanks.

I think the REGO_ModuleResourceConverter are from RoverDude's Regolith mod (now part of the stock KSP in some way). It was another "on-the-rails" resource converter. ModuleResourceConverter should be the closest descendant to that module. I have no idea why these use both TACLS and that specific module.

stratochief66:
Thanks for the info, I will dig into it more when NathanKell is around.

PR #723, opened 2015-11-28
RO LR-87 LH2 update

PhineasFreak:
* Remove redundant FX (already taken care by the RealPlume
configuration) and duplicated EFFECTS definitions.
* Add support for the KSP 1.0.5 engine thermal changes.
* Fix a null reference exception caused by a wrong default engine config name.
* Replace the deprecated EngineIgnitor modules with the RealFuels fuel & ullage functionality.
* Remove the leftover engine alternator module and ElectricCharge
resource.

### DIFF
--- a/GameData/RealismOverhaul/Parts/Engines/LR87LH2.cfg
+++ b/GameData/RealismOverhaul/Parts/Engines/LR87LH2.cfg
@@ -18,33 +18,47 @@
 	@node_stack_bottom = 0.0, -1.629, 0.0, 0.0, -1.0, 0.0, 1
 	%node_attach = 0.0, 1.0101, 0.0, 0.0, 1.0, 0.0, 1
 	@attachRules = 1,1,1,0,0
-	!fx_exhaustFlame_blue = f
-	!fx_gasJet_white =f
-	!fx_exhaustFlames_yellow_tiny =f
-	!fx_exhaustFlames_white_tiny =f
-	!fx_exhaustLight_blue =f
-	!fx_smokeTrail_light =f
-	!fx_exhaustSparks_flameout = f
-	!sound_vent_medium =f
-	!sound_rocket_hard =f 
-	!sound_vent_soft =f
-	!sound_explosion_low =f
 	
 	@mass = 0.74
 	@maxTemp = 1973.15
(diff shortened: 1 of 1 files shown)


stratochief66:
The fx block (with ! before each line) appears to be deleting the effects

I assume %powerEffectName = powerflame was setting/re-enforcing the FX name? 
Why did you remove '@name = ModuleEnginesFX" ?

How does the engine behave differently before and after the %heatConductivity = 0.060 block is added?

Some of the changes are clear and make sense (fixing the ignitor stuff, renaming/fixing the engine default config)

Did you test this part in game before and after making these changes?

PhineasFreak:
Yes, RealPlume already has a patch to delete all FX and EFFECTS{} nodes from any engine that it finds. You may return these lines there, i removed them because of the above reason.

The %powerEffectName was already defined as such (exactly the same line) in the RealPlume config for that engine, so it added two different effect names. The "name" was deleted accidentally (ugh!) and thank you for bringing that into my attention! Also, i made another mistake withe the "EngineType": it should be "LiquidFuel" and not "LiquidEngine"...

For the engine thermal changes: it increases a bit the overall thermal radiation capability of the engine (cools faster after shutdown) and lowers the thermal conduction (reduces a bit the heat conducted to the tank that is mounted). Maybe @NathanKell could provide some info of how these values further affect the parts that include them. For now, these values are taken from the stock engine configs.

NathanKell:
The stock engine values are fine; we've never had a chance to try to figure out appropriate values for RO engines so we accept the stock ones for now. However, I had thought that as of 1.0.5 stock engines no longer modified conduction values (it's commented out in stock, I thought)...?

PhineasFreak:
No, all stock engines have them. The NERVA has even higher thermal radiation capability (0.85) than the rest.

NathanKell:
As I said, I was speaking of conduction values. I'm well aware of the emissivity changes. :)

PhineasFreak:
The conduction values reported are:

* heatConductivity = 0.06
* skinInternalConductionMult = 4.0

But in the end it doesn't matter because the LR-87 config is a copy of the stock LV-T45 engine that already includes these lines. So, there is no reason for me to add them. I will remove them.

PR #729, opened 2015-12-03
Reset J-2 with 3 restarts

ctiberious:
I know this was only recently changed.  I'm also aware that the Saturn V traditionally only fired the J-2 on the SIV-B stage twice.  But according to Wikipedia and other references, the only limitation on how many times the J-2 could be restarted was fuel supply and the amount of Helium available for tank pressurization.
The main difference between the S-IVB-200 and S-IVB-500 (besides the flared fairing) was that the -200 series did not include as much Helium since it would not need to be restarted in flight.
According to http://www.nasa.gov/centers/marshall/pdf/499245main_J2_Engine_fs.pdf, page 6-7, "Refill of the gaseous helium tank was not required because the original ground-fill supply was sufficient for three starts".

Furthermore, the reference document states (http://www.nasa.gov/centers/marshall/pdf/499245main_J2_Engine_fs.pdf page 6-2) "The ASI (Augmented Spark Igniter) operates continuously during entire engine firing, is uncooled, and is capable of multiple reignitions under all environmental conditions."

Plus, the only in-game way to simulate the fuel expelled that allowed the S-IVB to be redirected for lunar impact is with an additional burn.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
@@ -854,7 +854,7 @@
 		}
 		%ullage = True
 		%pressureFed = False
-		%ignitions = 2
+		%ignitions = 3
 		!IGNITOR_RESOURCE,* {}
 		IGNITOR_RESOURCE
 		{


stratochief66:
I will wait to see other opinions on this (@NathanKell , @raidernick) but I know that you can also use the Saturn APS units to re-direct the spent S-IVb into collision with the Moon.

I have not thoroughly testing the S-IVb with the new boil-off code in 1.0.5 RealFuels, but previously the boiloff of fuels during timewarp would be experienced as a forward burst of acceleration when timewarp ends. If you point the S-IVb before timewarping a bit, that should also produce some thrust in the desired direction.

A1Ch1:
I agree with this change. The J-2 was limited only by helium supply, it theoretically supported unlimited restarts using the ASI and start tanks.  The Saturn stage only used two ignitions because there was absolutely no need for a third one. The in-game version doesn't need to be limited to two ignitions. If anyone can find documentation that clearly limits the J-2 to one restart that would be evidence otherwise, but everything I've seen indicates there was no such limit on the engine itself.

RedAV8R:
I only support this change if there is also a requirement for Helium for a restart. That way realism configs for say Saturn can set the 'limit' to 1 restart while others can have however many their helium supply can handle.

raidernick:
This is actually the 3rd time it's been rejected. I agree with @RedAV8R for the 3rd time for the same reasons, let's stop making these.

A1Ch1:
Show me any documentation that indicates the J-2 was limited to one restart. Again, everything I've read indicates it was designed for multiple restarts.

raidernick:
NOT the damn j2, it has nothing to do with the engine capabilities but the system as a whole.

NathanKell:
What were the mass differences between the Saturn IB S-IVB's J-2 and the Saturn V S-IVB's J-2, other than helium?
I ask because, as @RedAV8R mentioned, this seems the best approach:

1. Add a helium RESOURCE block to the J-2 engine
2. Default it to the Saturn V S-IVB load, but
3. Allow it to be set to 0 to yield Saturn IB level mass
4. Allow it to be increased to yield 3 or more ignites.

However, IIRC (but I am not sure, hence my asking) there were other mass-increasing changes besides loading more helium, which would need to be modeled. There are various ways of handling that however.

A1Ch1:
I agree that adding a He requirement for start seems to be the best way of modeling restarts. I'd like to see any documentation that indicates what limitations the "system as a whole" puts on re-ignitions. There was no need for the S-IVB to restart more than once, but the J-2 didn't need anything more than ullage and helium.

A1Ch1:
The other changes to the J-2 between the S-II and S-IVB other than restart helium was elimination of the plumbing to refill the start tank, since they were pre-pressurized on the ground. The restartable version vaporized a small amount of H2 to refill the tank. This is also what limited restarts - the H2 had to heat up in the start tank to supply enough pressure to start the turbines, but after a certain amount of time escaping H2 prevented any further restart attempts. This is why re-ignition took place a certain number of orbits after insertion.

NathanKell:
Ok. And that was also done on the S-IVB when used on Saturn IB, right? IIRC that's a fair-sized savings in mass.

Schnobs:
Relevant sources:
http://www.nasa.gov/centers/marshall/pdf/499245main_J2_Engine_fs.pdf (p.7f)
http://history.nasa.gov/SP-4206/sp4206.htm (Stages to Saturn; p. 177ff (original pages) / p. 200ff (PDF pages))

Seems as if any extra use of Helium beyond a single ignition was mostly/only in order to purge residuals.

This issue ties in with the realism/reenactment issue I opened a while ago. For many engines there is no hard reason why they should not have more ignitions. In this particular case, I'm not even convinced that extra Helium would need to be modelled; weight-wise, wasting a few seconds worth of LH and LOX seems more important. If you insist on using Helium, just allow the player to add more in high-pres tanks.

ctiberious:
I'm not against modeling He as a resource for restarting pressurized engines.  But my guess is that's a bit more involved than the simple "ignitions = x" setup we currently have.  You also then have to figure out exactly how much He is used under various conditions (S-IVB-500 likely uses more He to pressurize it's tanks then the Surveyor probes did, for instance) and how much He (mass) was included in various rockets.
As an example, I've found plenty of references that state the S-IVB used on the Saturn V was pre-loaded on the ground with enough He for 3 starts of the J-2 engine but I can't actually find any documentation stating exactly how much mass that He was.  Or how much was used/lost during each start.
Also, I should point out that the He tank on the S-IVB stage did NOT supply pressure to the J2 engines on the S-II stage.  That stage likely had it's own tank, also fitted inside the LH2 tank, though I'd have to research to be sure of it's location.

I had always assumed that for many engines the ignition limit was to represent the limited amount of resources available as part of the restart system.  For instance, the Surveyor probes engines have basically no ignition limit because so little He is needed to keep the lines pressurized so you're likely to expend all your fuel before you use up all the He.  But the Centaur requires more He to keep pressure up so it's RL-10 engines have fewer restarts.  And the S-IVB-500 has even fewer because it requires even more He.  If it's actually supposed to represent the engines mechanical limitations, then the J2 would actually have a much higher limit then the 3 I am suggesting.  After all, the documentation on the engine clearly states that it "is capable of multiple reignitions under all environmental conditions".  

As for the mass of the J2 used on the S-IVB-200 vs the S-IVB-500, from what I can tell the engines were identical (though later J-2 used had greater thrust so not sure if there was a mass difference there).  The changes were mainly in the size and shape of the fairing, the fact that less He was included on the S-IVB-200 because no restarts were needed, the number of solid rocket separation motors (3 on the -200 vs 2 on the -500) and the APS motors which were not included on the -200 model.  With the exception of the He mass, all the rest is already counted since those are separate parts.

Finally, regardless of whether the dev wants to change the ignition system, my point remains.  The document I stated in my op clearly states that there was enough He included in the tanks to allow the J2 used on the S-IVB-500 to be started 3 times.  Why did they include enough He for 3 starts when they only ever needed 2?  Who knows.  Maybe it was for emergencies or maybe some post-Apollo plan would have needed more then 2 restarts and they didn't want to have to redesign the stage again.  Whatever the reason, the J2 on the S-IVB-500 could be started at least 3 times so that's what we should have the engine setup with.

RedAV8R:
Maybe there was enough for 3 because if all conditions were perfect 3 ignitions were possible, maybe that was just a safety factor. Regardless. Simply increasing ignitions is not the way to do it.

I agree, as long as J-2 was physically capable, heating/cooling wise, that we can increase ignitions. Add He as an ignition resource and add either He to the engine itself, or in the case of replica craft to the parts above it. Problem solved. If someone wants a custom/fictional craft with J-2, then they can add their own tanks.

Schnobs:
http://history.nasa.gov/ap12fj/pdf/a12_sa507-flightmanual.pdf, p. 119ff

There was no hard technical limit on the number of ignitions. Open the right valves at the right time and have resources flow through them.

Resources consumed per ignition: LH and LOX flowing through the engine before ignition (at least seven seconds, at tank pressure); one "start tank" worth of high-pressurized H2 to spin up the pumps; Helium to purge the engine after shutdown (~1 second).

There's lots of arcana about how it was done in case of the S-IVB stage; I'll elaborate if requested, but here's the nutshell: 
Presumably the heaviest part of it was the "Oxygen/Hydrogen Burner" to provide heat for repressurization; during operation, the main engine itself provides heat enough. Said burner was effectively a small rocket motor (cf p. 127) with intricate plumbing; other than that, few extra parts were needed.
The burner was ignited ~9Minutes before the second ignition.
The "start tank" was refilled with LH2 during operation (and it took ~60sec to do so); during coast, it would pressurize by itself. Hence the ~80 minutes wait before ignition could occur.

ctiberious:
@RedAV8R : Presumably the J2 is not the only restartable rocket that relies on Helium to pressurize the fuel tanks.  I'm fairly certain that the RL10 works basically the same way, and I'm sure there are others.  So if you could add He as a requirement for the J2 to restart, you'd have to add it (or another appropriate resource) to every other restartable engine.  I'm guessing the existing ignition system can do this because I've seen a few first stage engines that have a strange "fuel" (TEATEB or something like that).  If you or someone else wants to go through every engine and add in an "ignition fuel", cool beans.  But that's a bit beyond the scope of the change I'm proposing.  My PR is requesting a change based on the way the engine is currently setup and it's a change to match documentation I have found and cited.

@raidernick : You say "it has nothing to do with the engine capabilities but the system as a whole".  I've provided a resource that states the S-IVB-500 had enough resources to be started 3 times.  So the "system as a whole" should have at least a 3 start limit.  That's what we had until the resent change and that's what this PR is proposing we return to.  If you've got a source that contradicts the one I've provided, that's great.  Please post it here.

@NathanKell : Even if there is a mass difference due to changes in plumbing between the S-IVB-200 and S-IVB-500 because one never needed to be restarted and the other did, that's a little beyond the intent of this PR.  For that to be modeled properly, you'd need the ignition limit to be based on the fuel tank and not the engine.  Maybe we could show that by completely removing the J-2 engine's ignition limit but adding an "ignition fuel" requirement that was then included in the attached fuel tank.  I wouldn't be against that kind of setup but you'd have to do it to more then just this one engine.  That's a pretty significant undertaking considering the number of engines we have available.  Definitely beyond the intent of this PR.  If that's ultimately the direction you guys want to go, I'm all for it and would likely even volunteer to help.  At least with the engines I'm familiar with (mostly I've research the Atlas, Delta, Titan, Saturn, Agena and Centaur engine families).  But I'm guessing you'd want a branch dedicated to such a large set of changes.  And you'd need people that were familiar with the engines I've not really messed around with.  And none of that would change this PR which is proposed based on the system we currently have in place.

RedAV8R:
So therefore, when it's changed, it's going to be changed for everything. I've got enough on my plate right now, so you are free to start changing all of them.

I propose this PR be rejected, yet again, until it is implemented in such a way as proposed in this discussion with He as a required ignition resource.

Schnobs:
Why Helium of all things? All Helium used in the restart cycle was for tank pressurization. That is already reflected in the dead weight of tanks, even though the resource is never displayed or used. (I also wonder how much pressure may credibly be lost while coasting -- the tanks weren't leaking, were they?)

Yes, there was also a Helium purge after shutdown, but how much is that? If you think that the user has to pay for extra ignitions in terms of added weight, I'd advise you to look at the heater unit first, and the propellant flow during pre-start second. That was liquids after all, while the purge happened with gaseous helium.

More generally, I see overly correct models as a problem. Much hardware was tailored to a specific mission, and recreating it as accurately as possible means that the player has no choice but to reenact historical mission profiles.

RedAV8R:
Kinda the point of REALISM Overhaul. You are free to fork and release whatever you want.

@stratochief66 @NathanKell PLEASE PLEASE PLEASE just reject this PR and end this circus.

stratochief66:
I think all the necessary information to make a decision on this PR are in place. I'm going to lock the conversation for now.

stratochief66:
@NathanKell I think we should close this PR and tackle new standards for reasonable restart-ability after this release.

Had we previously discussed a reasonable number of seconds of fuel to consume for a restart, for game balance?

ctiberious:
Only question I have is, if it's agreed that the J-2 engine itself technically had no restart limit, and the S-IVB-500 launched with the ability to preform 3 starts, why are people so adamantly against setting the engine back to where it was in v10.4?  Not saying I'm against a more realistic restart system.  Just wondering why, while we're waiting on such a system to be introduced, we're restricted to fewer starts than the Saturn V was actually capable of.

A1Ch1:
I'm all for figuring out a system for resource consumption upon restarts (propellant and helium, helium bottles could be a standard restart requirement), but in the meantime the J-2 should be restored to three ignitions. This was a demonstrated capability.

NathanKell:
@stratochief66  I concur. :)

NathanKell:
@A1Ch1 We hope to release an RO and an RP-0 today, and I'd rather we work over the J-2 and get it right, than do a quick change here and then another one.

PR #732, opened 2015-12-05
SSTU Update

FFCJoseEduardo:
added new modular tank
removed RSSRO flag from old tanks as a warning (they will be removed soon)
fixed node toggle module for Orion SM
fixed layout definitions for 3x RS-25 mount


stratochief66:
there appears to be the originally scalled and configured Orion command pod? (in the non-RO menu, near the end of the list)

stratochief66:
I'm still seeing the 2 RealFuel tanks on the Common Bulkhead upper stage, with the dominant one sticking at 2kL. The other one appears good. :)

stratochief66:
Have you configured the RCS on the new upper stages tanks to use real fuels? If so, it might be worth noting in the description which fuels it uses in what ratios, or possibly just adding a little to the tank by default. I might be able to help with that.

PR #734, opened 2015-12-07
SXT J2 update

PhineasFreak:
* Add the requirement for liquid Hydrogen & Oxygen for ignition (might extend it later to also require Helium).
* Remove leftover ElectricCharge resource from ModuleAlternator.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Engines.cfg
@@ -395,6 +395,18 @@
 			name = ElectricCharge
 			amount = 0.5
 		}
+
+        IGNITOR_RESOURCE
+        {
+            name = LqdHydrogen
+            amount = 0.7454
+        }
+
+        IGNITOR_RESOURCE
+        {
+            name = LqdOxygen
+            amount = 0.2545
+        }
 	}
 	MODULE
 	{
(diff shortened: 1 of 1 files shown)


A1Ch1:
@stratochief66 

The thread regarding J-2 restarts was locked, but I've come across some interesting information and no obvious place to put it. According to the [Saturn V Payload Planner's Guide](https://www.dropbox.com/s/a6c9lnwj216d0ue/46779714-Saturn-V-Payload-Planner-s-Guide.pdf?dl=0) page 48, the S-IVB could be modified to include a second restart at the cost of some extra weight. The guide does not specify the exact weight needed to perform 3 S-IVB ignitions, but it would presumably not require any significant redesign of the stage since the guide includes several payload figures to orbits that require three S-IVB starts. Indeed, one of the major advertised advantages of the Saturn V is "large synchronous orbit payload capability (...about 62,000 lbs to an equatorial synchronous orbit)".

It's likely the extra mass a combination of ACS propellant for ullage and more helium for tank repressurization (different from pressurization, repressurization prior to restart is what required large amounts of helium. This helium came from external bottles, not the helium bottle in the J-2 itself [Pages 6-16 to 6-17](https://www.dropbox.com/s/6in2ue37w0rvzzg/sa507-flightmanual.pdf?dl=0), which was used for valve timing and contained enough for [three starts](http://www.nasa.gov/centers/marshall/pdf/499245main_J2_Engine_fs.pdf).) Regardless, it appears the S-IVB could be modified for three starts with minimal changes and Douglas was prepared to do so should the need arise.

stratochief66:
Any idea how many seconds of fuel consumption that ends up being? @PhineasFreak

NathanKell:
That's about 0.3kg of fuel, or 1/1000th of a second, by my calculations.

PR #739, opened 2015-12-09
Correct Gemini SM dry mass

stratochief66:
I ran the numbers for a very light and very heavy mission (Gemini 3 and Gemini 11 respectively) and found the craft was over mass about the same amount in both cases, by approximately 500kg.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Gemini_Pod.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Gemini_Pod.cfg
@@ -589,7 +589,7 @@
 	!node_stack_bottom1 = DELETE
 	@title = Gemini Adapter Equipment Section	
 	@description = The Gemini Adapter Equipment Section.  It contains O2 and H2 for fuel cell consumption and life support.  It also contains fuel for manoeuvring and attitude control thrusters.  This is the configuration as found on Gemini 10+. RCS O/F Ratio 1.3:1.
-	@mass = 0.67688
+	@mass = 0.177
 	@MODULE[ModuleDecouple]
 	{
 		@ejectionForce = 10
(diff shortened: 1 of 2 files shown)


RedAV8R:
Source? Not saying you're wrong, just asking where.

stratochief66:
@RedAV8R 

Entirely fair. the only source I have yet to find with masses for each Gemini mission individually is wikipedia, ie. https://en.wikipedia.org/wiki/Gemini_3

If there is a Gemini equivalent to "Apollo by the numbers" I would love to see it :)

I found that my FASA Gemini with full fuel for the fuel cells and LS supplies was substantially overweight for the capabilities of a Titan II, so i started digging and I found the failures of past-me.

RedAV8R:
Now I brought this up before when H2/O2 numbers were being changed for something or another. It was my understanding that these amounts were being fudged to deal with boil-off, rather than their historical numbers. This true?

stratochief66:
I recall some fine-tuning at some point to deal with boil-off (like, adding an extra L or something of LH2?) I have found boil-off in a ServiceModule tank to be similarly mild with the new builds of RealFuels, but I haven't rechecked them yet with the Gemini (they looked just fine with the Apollo SM and fuel cell though)

RedAV8R:
My point is. This shouldn't be happening. We need to fill tanks to their historical levels and THEN change boil-off to match usage. Not the other way around.

Because for most of FASA dealing with Apollo, Gemini, Mercury, I made tankage historical. Using the Familiarization Manuals as much I could. Yet Gemini has seen a 10x increase in O2/H2 from these values.

stratochief66:
Citation?

Perhaps we should move this conversation to an ISSUE, as this PR is purely about dry mass. That is, the dry mass of the Gemini SM with no fuel cell supplies, only LOX to provide breathing gas.

RedAV8R:
As I said before. Gemini Project Familiarization Manual.

Vol 1 p5-30. Vol 2 p5-33.

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/issues/745

RedAV8R:
#745 works...I strongly recommend this be rejected/closed then.

stratochief66:
Currently, a craft kitted out (dry empty SM) like Molly Brown (Gemini 3) can barely get into orbit on a Titan II GLV, while one kitted out as Gemini 11 or 8 has no chance to reach even 130kmx130km.

This is a near term fix to correct that, while more precise values are researched, sourced, and implemented.

RedAV8R:
Whatever. I'm done. Do whatever you guys want.

PR #742, opened 2015-12-09
RO Mk1 fuselage scaling fix

PhineasFreak:
* Fix incorrect RO-Mk1Fuselage scaling factors.


NathanKell:
Is this correct for both stock and for VSR 1.9 ?

PhineasFreak:
It is correct for VSR but now that you mention it, i will check it for stock.

PhineasFreak:
I tested both VSR & stock and size is correct on both cases. But it seems like VSR doesn't touch the MK1 fuselages?!?! I thought it did...

PR #752, opened 2015-12-11
RS-25 stock & VSR scaling update

PhineasFreak:
* Uses different scale factors and node placements for stock & VSR.


NathanKell:
I don't believe you can :NEEDS a value. Instead you should set the nodes normally, then do a second @PART, i.e. @PART[foo]:FOR[RealismOverhaul]:NEEDS[VenStockRevamp] and change the nodes there.

PhineasFreak:
No, it is absolutely possible. In fact, i was surprised when i first discovered it some months ago.

https://github.com/sarbian/ModuleManager/wiki/Module-Manager-Syntax

You can also test it by exchanging temporarily the "!" for the node positions. It works beautifully!

NathanKell:
Huh! Awesome!

PR #753, opened 2015-12-11
VSR radial fuel tanks RO support

PhineasFreak:
* Add RO support for the radial fuel tanks.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_FuelTank.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_FuelTank.cfg
@@ -48,7 +48,7 @@
 {
     %RSSROConfig = True
 
-    @title = Stratus-VI Roundified Tank (Pressurized)
+    @title = Stratus-VI Roundified Tank
     @description = A short radially - mounted tank that can be configured to hold a wide variety of resources.
     @mass = 0.0035
     @maxTemp = 1073.15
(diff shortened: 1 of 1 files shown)


NathanKell:
It might be worth doing what we do for the Mk3 parts, i.e. have the LFO and LF parts both ingame, but have one kind set to one tank type, and the other to another. In this case, since radial monoprop tanks already take care of ServiceModule, maybe Default and Cryogenic?

PhineasFreak:
@NathanKell i envisioned the tanks as storage for non - cryogenic type of fuels and/or for life support purposes but now that you mention it, you are right.

Question: should i change the tank type from "Default" to "ServiceModule"?

NathanKell:
I'd suggest leaving the LFO types as "default" and the LF-only types as Cryogenic

PhineasFreak:
Perfect! I will have them ready in a few hours (because i have been awake for too long).

NathanKell:
Awesome! And do please sleep, sleep's important. :)

PhineasFreak:
Batteries full, let's do some modding!

PhineasFreak:
Dang it, i deleted the wrong branch! Stupid me...

NathanKell:
Whew! Glad it's back :)

One last thing--I'd suggest removing (Pressurized) from the type=Default tanks. While it's technically true, we generally reserve talking about a tank as pressurized for when it's _Highly_ Pressurized, i.e. type ServiceModule or Fuselage.

PhineasFreak:
OK, on it!

NathanKell:
Thanks! :)

PhineasFreak:
My pleasure working with you guys & gals!

PR #755, opened 2015-12-11
VSR engines update

PhineasFreak:
* Remove the deprecated VSR RS-25 engine RO config (replaced by stock).
* Add RO support for the VSR RE-D7 "Bollard" engine (modeled after the
RD-107/RD-108 engines - https://github.com/VenVen/Stock-Revamp/issues/46).

NOTE: RD-108 part dry mass might be not balanced and the gimbal range is WIP. If anyone has a source for the vernier gimbal range please let me know.

Also, do we need to create a copy for the RD-107 series?

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Engines.cfg
@@ -669,99 +669,6 @@
 	!nodeScaleZ = DEL
 }
 
-@PART[Ks25]:FOR[RealismOverhaul]:NEEDS[!RftS,!RealFuels_StockEngines] // RS-25D/E
-{
-	%RSSROConfig = True
-	!MODULE[TweakScale]
-	{
-	}
-	@MODEL
-	{
-		// Size3 cluster has 2.2352 but the real engine should only scale up this much.
-		%scale = 1.947, 1.947, 1.947
-	}
-	%scale = 1.0
-	%rescaleFactor = 1.0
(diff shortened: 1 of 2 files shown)


stratochief66:
Do just the verniers gimbal on this model?

The description looks like a stock copy-paste, can you remove or alter it to reflect the actual engine? There might be an existing one you can borrow.

Did you source the numbers from another RD-108 config? can you cite it?

Your mass in your cited block doesn't match the number you use.


Does the Ks25 no longer exist in the new VSR then?

PhineasFreak:
* Yes, verniers do gimbal.
* I missed the description replacement. If it's OK, i will reuse the one from the SXT RD-108.
* Take a look at the part comment block. There i mention the reason for the part mass and the links from where engine data were sourced.
* VSR version 1.0.5 deprecated it's version of the RS-25 and now replaces the stock one. Specifically, if you take a look at the part description, Ven states that:

>KS-25-X "Vector" Rocket Engine -please remove-

Since there is no reason for him to add a duplicate engine. He probably keeps it for backwards compatibility.

stratochief66:
Of course you can definitely re-use RO's own 108 description :P

Perhaps ! out the Ks25 part then in RO, so it doesn't show up?

PhineasFreak:
I will ping @NathanKell for the Ks25 issue so he can also take a look. As for the description, i am fixing it right now.

NathanKell:
Yep, ! is better here. I do that for some other VSR engines, see the top of the VSR engines file. :)

stratochief66:
Merging, I will do the ! for the Rs25 in a separate PR for expediency

PR #758, opened 2015-12-12
OLDD Saturn V

PhineasFreak:
* Update the S-IC retrorocket RO config to use a copy of the S-II retrorocket configuration.

Previously this part was created by "PART" and not by "+PART" tag. If the mod that supplied the part was not installed, then it spammed the KSP log with errors.

Part functionality was not affected.


stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/758/files#diff-a602d59da70009e7364b7899ed890f97R551

0.15 to -0.15 ? why?

NathanKell:
Either way it needs an orientation set (the second triplet of numbers). Otherwise it'll spam exceptions in the editor. If node is at 0, 0, 0.15, the whole line should be 0, 0, 0.15, 0, 0, 1 (and -1 if -0.15)

PhineasFreak:
Before i made the changes, i took a look the configs and they all need a "face-lift". This specific PR could be delayed as i was planning to update the Saturn V RO configs (and the new CMES update doesn't help either).

PhineasFreak:
@NathanKell @stratochief66 i have completely remade the Saturn V RO configs but i am stuck on a small issue: the J-2 should have 2 or 3 restarts?

stratochief66:
@PhineasFreak 

Set as 2 for now, it will shortly get modified to tie into the new centralized engine config system anyway as per this.
https://github.com/KSP-RO/RealismOverhaul/pull/778

NathanKell:
@stratochief66 I'd like to get an RO out before the new engine system, since there have been a fair few commits. Is this safe to go in now?

PhineasFreak:
@NathanKell should i replace the files here? As i mentioned earlier, this is not really applicable anymore since i remade the CFGs completely.

stratochief66:
@NathanKell @PhineasFreak 

The same problem on line 551 still exists, no orientation.

In the comments, it also seems the reason for going from 0.15 to -0.15 wasn't explained. I would say to leave this PR out, it needs work.

PhineasFreak:
@stratochief66 this specific PR is outdated now anyway. I was considering closing it and opening a new one where i have updated all the Saturn V parts. Am i clear to proceed?

PhineasFreak:
Now completely updated! Some notes though:

1. Requires the new engine config system.
2. When decoupling the S-II and the S-IVB interstages, the whole stack will explode (because of the colliders).
3. The fairing does not drop away when decoupled (the built - in engine modules did not work - i will use a nice workaround that RaiderNick did with the BobCat Ares V fairing).

stratochief66:
Now there are so many changes that it won't let me see the code differences, and there are merge conflicts. I would suggest closing this and making a new PR. Given the number of lines changes, it may make sense to break it up into multiple PRs.

PhineasFreak:
So i should take each part separately, edit it and make a PR?

PR #759, opened 2015-12-12
CryoEngines AIES and KW config consistency/improvements

SirKeplan:


A1Ch1:
It's a minor thing, but re: J-2X mass, I see you're changing 2.47 to 2.472. I gather this is from converting 5,450 lb to 2,472 kg This is [false precision](https://en.wikipedia.org/wiki/False_precision), and it crops up a lot, particularly with thrust and mass, where numbers have been converted from US customary units to SI. The problem is significant figures - 5,450 lb does not equal 2,472 kg, it equals 2,470 kg, unless we assume the final 0 is significant.

SirKeplan:
A1Ch1: yeah, that makes sense, there are other figures out there, but 5,450 lb is most reliable. gonna remove that false precision.

A1Ch1:
Regarding the RS-68, we have some rather messy numbers currently. From [this](http://web.archive.org/web/20140328201054/http://ulalaunch.com/site/docs/publications/ULA-Innovation-March-2010.pdf) we know that the original RS-68 throttled from 57%-102%, and the RS-69A throttles from 55%-108%. From [this](https://www.dropbox.com/s/e9pje1fv39ydfiu/EVOLVED_EXPENDABLE_LAUNCH_VEHICLE_SYSTEM__RS_68_M.pdf?dl=0) we see that the RS-68 generated 3341 kN at full thrust and 1922 kN at min thrust, with an Isp of 409/357, so the RS-68A should throttle from 3538 kN to 1800 kN. We know that the SL thrust of the RS-68A is 3137 kN and the vacuum Isp is 411.9 from [this](http://www.ulalaunch.com/uploads/docs/Launch_Vehicles/Delta_IV_Users_Guide_June_2013.pdf). Therefore, its SL Isp can be calculated to be 365.

So I get:
RS-68
maxThrust = 3341
minThrust = 1922
atmosphereCurve
{
key = 0 409
key = 1 357
}
RS-68A
maxThrust = 3538
minThrust = 1800
atmosphereCurve
{
key = 0 411.9
key = 1 365
}

SirKeplan:
Ah, thanks, my searching didn't find all those sources.
However your first and second link disagree with each other: if maxthrust is 3341 and throttle goes from .57 to 1.02, then minthrust should be 1867 not 1922.
So either full power in that chart is 100% instead of 102%(that gives a minthrust of 1942 still not quite right), or the thrust levels mean something subtly different to what I thought.

A1Ch1:
Ah yes, I forgot that [those](https://www.dropbox.com/s/cy6vtg32hp5yq4r/PropulsionForThe21stCentury-RS-68_Slides.ppt?dl=0) numbers are actually from when the engine was specified to throttle from 58%-101%. Substitute that and everything checks out.

A1Ch1:
So I get:
RS-68
maxThrust = 3374
minThrust = 1922
atmosphereCurve
{
key = 0 409
key = 1 357
}
RS-68A
maxThrust = 3573
minThrust = 1819
atmosphereCurve
{
key = 0 411.9
key = 1 361.7
}

And since it's likely the numbers in kN are simply converted from Klb-f (751 Klb-f converts to 3340.6 kN) the fourth digit isn't significant and is false precision (and regardless thrust is always + or - 1%-2%, so any more than 3 sigfigs is overkill) . So a more reasonable conversion would be:

RS-68
maxThrust = 3370
minThrust = 1920
atmosphereCurve
{
key = 0 409
key = 1 357
}
RS-68A
maxThrust = 3570
minThrust = 1820
atmosphereCurve
{
key = 0 412
key = 1 362
}

SirKeplan:
Thanks A1Ch1, realizing it's 101% really clears that up

PR #762, opened 2015-12-12
Dependency check

PhineasFreak:
* Just a small check for the Vanguard probe when RP-0 is not installed.


NathanKell:
Should be RP-0 not RP0

PhineasFreak:
Right now i am checking for the DLL. Is the internal name different than the DLL name?

NathanKell:
Nah, but we use the folder name (RP-0) so RP-0 patches run on a later pass. Best to use RP-0 in NEEDS as well.

PhineasFreak:
Done!

PR #769, opened 2015-12-14
Modify VSR BE-4 Thrust level

Wetmelon:
Thrust value from
http://www.ulalaunch.com/uploads/docs/BE-4_Fact_Sheet_Web_Final_2.pdf

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Engines.cfg
@@ -1281,8 +1281,8 @@
 		CONFIG
 		{
 			name = BE-4
-			maxThrust = 5295
-			minThrust = 2120 // Guess 40% throttle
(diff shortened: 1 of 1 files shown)


PhineasFreak:
You must take into consideration that the ULA brochure values are valid for a single engine but the VSR engine part model has two of these engines mounted.

SirKeplan:
If we're going to have a BE-4 in RO, wouldn't it make sense to have a single engine version, rather than model the dual engine one?

Wetmelon:
Ah! Good point. Then perhaps it should be made clear in the name / description?

PhineasFreak:
@SirKeplan the BE-4 is somewhat of a template right now (since the engine doesn't really exists outside of the testing benches). I am sure that we will find a better engine model in the future.

stratochief66:
So, the VSR part has two nozzles, so the original value is alright standing, correct?

I agree that a single nozzle model would be better, until/unless a solid design is in the pipeline that calls for them in this configuration.

Until someone changes over to those, the values are fine staying as is, right?

NathanKell:
The ULA values should be doubled; that's still more accurate values than what we had, and the throttle percent is certainly different.

SirKeplan:
it should be left unchanged so far, the current values are based off the 550,000 lbf thrust value(which is sea level thrust)

PR #771, opened 2015-12-14
B9AerospaceProceduralParts

Olympic1:
`B9AerospaceProceduralParts` is only compatible till 1.0.4. Crzyrndm forked and updated it to 1.0.5, so we need to make a new netkan file. The new id is `B9-PWings-Fork`.

Merge after KSP-CKAN/NetKAN/pull/2827


Olympic1:
KSP-CKAN/NetKAN#2827 is merged, ready to merge.

PR #777, opened 2015-12-16
AJ10-118F/K Delta configurations

ctiberious:
Add engine configurations for the AJ10-118F and AJ10-118K.  These engines were used on the "Delta F" and "Delta K" second stages for Delta I and Delta II rockets.  I've added these configurations to the existing SXTAJ10Mid engine because they were advances on the older AJ10-118E.  That may be accurate for the 118F since the Delta F second stage was similar in size to the older Delta E second stage.  But it's possible the 118K should be a completely separate engine since the Delta K was larger.  Trouble is I can't find documentation that shows the dimensions of the engine itself.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Engines.cfg
@@ -1330,6 +1330,78 @@
 			cost = 30
 			entryCost = 600
 		}
+		CONFIG
+		{
+			name = AJ10-118F	// Delta F 2nd stage
+			minThrust = 42.3
+			maxThrust = 42.3
+			heatProduction = 100
+			PROPELLANT
+			{
+				name = Aerozine50
+				ratio = 0.502
+				DrawGauge = True
+			}
+			PROPELLANT
+			{
+				name = NTO
+				ratio = 0.498
+			}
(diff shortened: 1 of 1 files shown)


stratochief66:
@NathanKell 

is this a decent way to host more AJ10 configs? Will there be an AJ10 advanced model at some point?

NathanKell:
I would rather see F and K as a separate engine, since it was converted over to NTO / Aerozine, ran at much higher chamber pressures, and had a different nozzle design and to some extent engine design itself. (The path went more like 118E -> Transtage and SPS -> 118F -> 118K)

stratochief66:
In practice, what would be the best way to go about that, given we don't have an additional model for an AJ10 (Adv) ?

+PART[SXTAJ10Mid]:FOR[RealismOverhaul]
{
  @name = SXTAJ10Adv
}

then, deleting the current configs and put in the 118F and 118K?

or should @ctiberious make a PR to SXT to add the + PART-ing there?

stratochief66:
@NathanKell  nudge

ctiberious:
I believe we decided to create a new AJ10Adv engine to include these configs on.  It was one of the things we talked about in the Later American placement, ctiberious #417 over on RP-0.  So I'll close this PR for now until I'm ready to submit all the changes from that discussion.

NathanKell:
Yep!

PR #778, opened 2015-12-17
Proposal: separate different engine variant configs from part cfg files.

SirKeplan:
Is this a good idea? using the RL10 here as an example, as it has so many variants.
I'm basically using a similar method to what is used for RCS.


NathanKell:
Not only is this a good idea, it's a GREAT idea. I've actually been meaning to do this--and talked about it rather extensively a few times in the channel--but haven't had time. If you want to take it on, do so with all and more of my blessings. :)

NathanKell:
(it would make a bunch of other things, like RP-0 configs, entry cost modifiers, and the like--much easier to deal with too)

SirKeplan:
OK great! what i'll do then is at some point i'll go through and everything i think needs it to this PR.
Are there any kind of conventions/naming  i should  be following for this?

SirKeplan:
Hmmm, we could probably put more than just `ModuleEngineConfigs` in here actually, how much of a  config file is it worth separating?

stratochief66:
What remains to be done in this? Just curious to know when this is good to go so I can do a thorough review.

NathanKell:
@SirKeplan haven't gone through it all yet--how do you handle multiple engines? I'd suggest doing what RCS does, having thrustMult and costMult keys that then multiply a CONFIG's maxThrust and minThrust, and cost (but *not* entryCost).

SirKeplan:
@stratochief66 right, there are still some configs i am going to add, should be pushing more commits in soon.

@NathanKell ah, that sounds like an improvement on my current method, going to look into that now.

NathanKell:
The whole file is good, but here in particular may help:
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/RO_RCS_Config.cfg#L653-L662
Just add one for cost and one for thrust rather than just the one for cost as in the file (the RCS autoconfigger should use a thrust mult too instead of the manual way it currently works...moar fixes...)

SirKeplan:
I was thinking of having single engineTypeMult field, that would be a general multiplier that gets applied to thrust, cost, and maybe mass(origMass).

NathanKell:
Ah, yes, that makes sense.

SirKeplan:
@stratochief66 @NathanKell probably a good time to look through this now, i haven't done *every* single engine, but have done most duplicates and engines with various variants. more can be done later if found necessary.

NathanKell:
Mom's in the hospital, has been since the 24th (she's stable, but she fell down stairs and broke her ankle fairly badly). So I won't be on this for a good while yet.

SirKeplan:
Ok. Sorry to hear that, I hope she makes a good recovery.

PhineasFreak:
How we will handle engines that also include "supporting" hardware, like propellant lines, fairings, heat shields, etc?

SirKeplan:
@PhineasFreak so far i mostly haven't been, i've just left the awkward engines unmodified. definitely worth thinking about though, some engines appear to contain a lot more structure than others, example being the KW and Tantares Vulcain, vs the Cryoengines Vulcain.

PhineasFreak:
Hmm, we could do it with a patch that runs after the normal RO pass and modify the mass parameters of the engines (mass, origMass and massMult), since each engine will already have the "base" mass values.

PhineasFreak:
@SirKeplan Ok, i did some testing with the RL-10 setup and i think that we could achieve the mass changes using auxiliary configs.

* First, the main RO pass (scale, name, operation). 
* Second, the engine parameter pass. It will use a system like RCS, meaning that the "AFTER[RealismOverhaul]" check will be replaced by "FOR[RO-Engines]".
* Third, a custom patch for any engines that do not comfort to the base parameters using "AFTER[RO-Engines]".

NathanKell:
Offsets could also be built into the main pass itself, i.e. it could check for, say, a value massOffset = foo and then add foo to its configured mass/origMass.

SirKeplan:
@NathanKell that makes sense, i was thinking along those lines.
although some engines are more complicated, so it might make sense to do as @PhineasFreak said and have custom patches, possibly both is best?

@PhineasFreak are you testing of my branch? your points are roughly what i have done(though i have it named RO-EngineConfigs). the only patching of non conforming engines so far is with parts that are multiple engines, but as we say others with extra mass etc. could be done similarly.

PhineasFreak:
@SirKeplan no, i created a temporary file config to test only the RL10 but as i see now, newer engine passes already implement the RCS system. Sorry about that...

As @NathanKell says, we can also create new parameters to the main RO patch so that we can configure parts that include multiple engines. An engineCount = X parameter could be used and then at "minThrust" and "maxThrust" values we can do some math to multiply the thrust according to the number of the engines:

>@PART[x]:FOR[RealismOverhaul]
{
    %engineCount = 2
}

Then, patch the thrust values at the "RO-EngineConfigs" pass:

>@PART[x]:HAS[#engineCount]:FOR[RO-EngineConfigs]
{
    @MODULE[ModuleEngines*]
    {
         minThrust = valueA
         maxThrust = valueB
        @minThrust *= #$/engineCount$
        @maxThrust *= #$/engineCount$
    }
}

But this requires from the config creator to include the thrust values for one engine within the EngineConfig pass. If we can get the thrust values using MM variables beforehand it would be much better!

PhineasFreak:
Another question: how will we handle engines that are required to have only one engine config? Take a look at the following part for an example:

https://github.com/PhineasFreak/RealismOverhaul/blob/CMES_update/GameData/RealismOverhaul/RO_SuggestedMods/CMES-WIP/RO_CMES_Structural.cfg#L167

SirKeplan:
@PhineasFreak yeah, not sure. that one is an extreme case, not only is it a subset of the RL10 configs, it is also a whole stage rather than simply an engine.
My less than optimal solution was to leave parts like that alone, unless there's something smart in modulemanager that can help?

PhineasFreak:
@SirKeplan i was afraid of that. Apart from breaking down every engine config and creating separate sub - configs for each one, i also cannot see how this could be achieved.

Schnobs:
I don't see that last one as a huge problem. Point out which engine config ought to be used in the part description, and leave it to the user. Bonus points if it defaults to the correct engine.

stratochief66:
I would say that the 'less than optimal solution' is the best one to go with for now, leave it 'as is' (unless some module manager wizard has a clever idea.

I wouldn't expect this PR/implementation to allow us to use this trick on every engine, but hitting most engines is pretty damn good for now.

Sorry I haven't had time in the past few weeks to review this PR (or conversation), hopefully NathanKell or I will have a chance sometime in soon.

NathanKell:
Need to be able to set a flag to not add origMass, and need to do an After[RO-engines] pass to @MODULE[ModuleEngineConfigs] { !CONFIG,*:HAS[~name[CECE]] {} }
That should do it.

SirKeplan:
A wild module manager wizard appears!
that looks like it would work, is creating a flag to not use origMass necesarry, won't the massOffset work just as well if i set the right value?

what file is best to place the module manager line in? just next to the part definition?

NathanKell:
Heh. :)

The offset would work, but IMO it's bad practice, the point is that in this case ModuleEngineConfigs shouldn't be changing mass at all.

Yep, after the part definition on an AFTER.

NathanKell:
Plus, config masses could change later and then you'd have to change the offset. The whole point of this project is to make config changes independent of parts. :)

PhineasFreak:
A nice, elegant solution! @NathanKell you are truly a god!

SirKeplan:
@PhineasFreak what's your source on the mixture ratio for the CECE? my nmbers are coming out slighly diff to what you have.

PhineasFreak:
The CECE is (for now) a rebuilt RL10-A variant so an O/F ratio between 5.5 and 5.85 would be applicable. A better source: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20090028814.pdf

NathanKell:
@PhineasFreak Nah, just been doing this a looong time by now. :) (And this is very similar to the auto-RCS thingie)

SirKeplan:
@PhineasFreak odd, looks like you actually had 5.84 ish ratio for the Altair descent module. your source doesn't give clear mix ratio values, i'm staying at the 5.8 i saw elsewhere for the minute.

A1Ch1:
The mixture ratio (and specific impulse) of the actual engine would vary depending on throttle anyway so the exact mixture ratio is kind of arbitrary

SirKeplan:
@A1Ch1 yeah, unless we implement a variable mix ratio ingame, then these values have to be somewhat rough i feel.

PhineasFreak:
@SirKeplan the source was about the RL10 --> CECE conversion and that it should use an O/F ratio of 5.5 (since it is an RL10). The values that i used were taken from the RL10 engine configurations (back then i did not check or include any sources).

NathanKell:
Just dropping by to send kudos to y'all :)

stratochief66:
Trying to do all of this all in one PR is probably a bad plan, since in altering so many files, some of the files you alter will end up being changed between the creation of the PR and completion/merging.

I would suggest pealing off all of the Engine_Config/*.cfg file creations as one PR (reviewing all of that will be a decent feat in and of itself)

Then, the major mods/mods seeing a lot of work should also be done in individual PR's to avoid merge conflicts.

Does that make sense? @NathanKell @SirKeplan

PhineasFreak:
Taking a look at some of the engine configs, it might be possible to also include the RF propellant checks (ullage, pressure & ignitions) with the global engine configs. Then we won't have to worry about new RO engines that do not have them included.

SirKeplan:
@PhineasFreak i was wondering about that and other fields when i started. I thought it might be complicated and i didn't get any input, so left them out.

worth having a look though.

PhineasFreak:
It is a good chance now since all configs will be migrated to the new system.The only "problem" will be to copy and paste all the RF fields from the ModuleEngine{} block to each of the engine CONFIG{} blocks.

PhineasFreak:
BTW, @SirKeplan your work is amazing! Keep it up!

PhineasFreak:
Cluttering the conversation but after making some tests, i found that it is possible to completely ignore the original ModuleEngines{} block of an engine because it gets replaced by the contents of ModuleEngineConfigs{}.

We can also assign the "ModuleEnginesRF" type directly from the ModuleEngineConfigs{} and remove these from the RealPlume configs (example: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RealPlume_Configs/CryoEngines/cryoengine-125-1.cfg#L15).

SirKeplan:
@PhineasFreak thanks for testing these things.

is it worth getting rid of ModuleEngines{} or replacing them with stubs? just to tidy up the whole thing.

about "!fxOffset = DEL", maybe i'm blind, but i only see them already just in RealPlume files

NathanKell:
Not quite true: you do need the stub and the stub needs to set the MODULE name to ModuleEnginesRF. You can't change the module type once KSP loads.
The reason it's working without that is because the RealPlume configs *also* force the correct name IIRC.

PhineasFreak:
So, RO has a patch to convert all engines into the RF compatible type.

The quick test i was referring to was for the CST-100 engine pack. Easy to retrofit and test new things: https://github.com/PhineasFreak/RealismOverhaul/commit/98216b7e87a2e876491d9df7a48f987ed3e6272a

PhineasFreak:
After even more testing, i have compiled a list for the overall changes:

1. Keep the "@Module[ModuleEngines*]" patching.
2. Replace "name" with "ModuleEnginesRF"
3. Keep the "heatProduction" parameter.
4. Move the ullage requirements and simulation fields over to the respective global engine config so that all engines can have them by default.
5. Ignore the "PROPELLANT" and "atmosphereCurve" blocks since they get overriden by the global engine configs.

The nice thing is that ModuleEngines can be used to insert custom parameters into an engine config. For example, someone could add an extra ignition requirement to the Apollo SPS engine asking for a Nitrogen resource (pneumatically - operated valves).

PR #779, opened 2015-12-17
Added SSTU Apollo and AJ10-137

FFCJoseEduardo:
Needs to be tested

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU Orion.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU Orion.cfg
@@ -196,7 +196,7 @@
 
 	%skinMaxTemp = 3500 // 3200
 	%maxTemp = 2273.15
-	%emissiveConstant = 0.85 // pretty much black, but not perfect emitter
+	@emissiveConst = 0.85 // pretty much black, but not perfect emitter
 	%thermalMassModifier = 1.0
 	%skinMassPerArea = 0.6
 	
@@ -315,8 +315,10 @@
 		outputResources = Oxygen, 0.006883126, false
 	}
 	
-	@MODULE[ModuleEnginesFX],*	
+	@MODULE[ModuleEngines*]
 	{
+		@minThrust = 33.362
+		@maxThrust = 33.362
 		@PROPELLANT[MonoPropellant]
 		{
 			@name = MMH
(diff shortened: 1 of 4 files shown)


stratochief66:
RCS on the new Apollo capsule doesn't appear to work, or at least it didn't appear to consume fuel or generate thrust.

The custom SSTU chute module didn't appear to generate drag (with FAR) @FFCJoseEduardo did it work with stock atmo? If so, my guess is that he might need to do something to cause it to also reflect increased drag with FAR. I don't think we can override his chutes since they are built into the capsule.

stratochief66:
What are the advantages/difference between the SSTUHeatShield module and the ModuleAblator we use? If there isn't any advantage to SSTUHeatShield I would suggest a straight removal

!MODULE[SSTUHeatShield] {}

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R111

Please rephrase this as an @ . Also, @ inside the module since it already exists.

stratochief66:
The SM engine bell moves when I try to aim the thrust, but the thrust itself doesn't seem to gimbal. interesting

stratochief66:
Tested, worked!

https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R209

Remove this line, and the CM RCS works.

stratochief66:
Looks pretty good, I will have to do a quick test again before I can merge.

Please skim for indentation errors, like these:
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R96
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R96
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R302
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R302
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R302

stratochief66:
Any if anybody (@NathanKell ? ) has any idea why this part would visually gimbal, but the FX and actual through wouldn't gimbal, please let me know.

https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-58a231f424e726507f69400cd3afd542R563



@FFCJoseEduardo I need to check the assemblies of these engines, it could be that this one and the clusters don't share their thrustTransform name, but the clustered ones appear like this:
https://github.com/KSP-RO/RealismOverhaul/pull/779/files#diff-534c625c6ec30d529e704a72cff4b5a3R1341

NathanKell:
Presumably because the thrust and FX transform(s) aren't parented to the gimbal transform?

stratochief66:
@NathanKell 

Is that done at the mu/model level?

NathanKell:
Yes. I would bring it up with the author; we should mirror their original configs regarding what transforms to use for what, so if it's not working for us it's not working for them either and they should fix it.

PR #783, opened 2015-12-19
changes to radiator

stratochief66:
based on in-game experience and discussion with NathanKell

@NathanKell, please point out of any of my comments are incorrect. I figured it would be useful to have some documentation of these variables.

Also, according to NathanKell (if I interpreted this correctly) radiatorMax sets the priority of a part to be cooled, higher is better. above 0.01, below 0.99. I set some tanks to 0.9 (2 tanks on one craft) and the habitation module to 0.35 and it didn't seem to behave as expected, so I didn't commit that to this PR

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
@@ -917,12 +917,14 @@
     %breakingTorque = 250
     @maxTemp = 1073.15
     %fuelCrossFeed = False
-    @radiatorHeadroom = 0.25
+    @radiatorHeadroom = 0.29 	// 0.2702 sets the limit to 17C/290K which is the ISS, this relies on maxtemp as 1073.15
+    				// this means 'the max temp the radiator will rise to is MaxTemp * radiatorHeadroom'
 
     @MODULE[ModuleActiveRadiator]
     {
-        @maxEnergyTransfer = 1750
-        @overcoolFactor = 0
+        @maxEnergyTransfer = 50000 	// how many KJ per sec (or tick?) the part can transfer, NathanKell recommended this
+        @overcoolFactor = 0.0186367 	// minimum temp to lower parts to = maxTemp * overcoolFactor
+        				// here that is 20K, so cool enough to keep LH2 liquid
 
         RESOURCE[ElectricCharge]
         {


NathanKell:
Rather than priority, it sets what parts are put in the "cool us" group and which are not.
If a part's temperature is > radiatorMax * maxTemp, then it is cooled by radiators that tick, else it is not.

stratochief66:
ahhh, so that basically lets you set the temperature target!

I will have to revisit those values then

NathanKell:
Yep, you're defining the maximum permissible temperature of the part, basically.

stratochief66:
Alright, some testing has led to unexpected results, so I want to ensure I have my terms straight first.

Retractable Radiator parts have:
radiatorHeadroom: cooling is provided if maxTemp * radiatorHeadroom > temp
(temp here is the radiator's own current temperature)

overcoolFactor: cooling will only be provided to parts where: temp > overcoolFactor * maxTemp
(temp here is the potentially cooled part's current temperature)

Cooled parts have:
radiatorMax: if maxTemp * radiatorMax < temp , cooling is received
(temp here is the part's own current temperature)

stratochief66:
@NathanKell 

So, radiatorHeadroom is applying as described above, no trouble.

Either overcoolFactor or radiatorMax are not acting as I would expect. Currently, I have tanks being cooled down towards 0K with no regard for the values I set for these.

I have maxTemp for both the radiators and the tanks set to 1073.15 all around.

What is a retractable radiator supposed to do with a potential cooling tick, if it finds nothing needs cooling? I would expect it to just not cool, but perhaps it is applying cooling anyway?

NathanKell:
1.A radiator will pull heat if temp * headroom < maxtemp.
2. A radiator will then pull heat *from* parts where target part temp > target part radiatorMax * target part maxTemp *and* where target part temp > radiator temp * radiator overcool factor.

Other than that, dunno what might be going wrong... :\

stratochief66:
I think that is more than sufficient to get 'unexpected' behaviour (since I appear to have the equations wrong.

I will change it up to match, and see how that goes.

stratochief66:
1. Isolating temp, that is temp < maxTemp/radiatorHeadroom , 1073.15/0.29 , temp < 3700K ?
Since that is the one thing working as I expect, I believe it is temp < maxTemp * radiatorHeadroom

NathanKell:
You're quite right, I flipped the sides. Oops.

stratochief66:
the second is an AND, my guess is that it is behaving as an OR

in my case, the targetTemp (20K or higher) targetRadMax = 0.0187 and targetmaxTemp = 1073

so, the equation stops being true when the targetPart temp falls to 20K

the second, the targetTemp (20K or higher) radTemp maxes at 211K, and radOvercool is 0.0186

That only becomes untrue when the part temp drops to 3.9K , which is fairly close to where my part stop cooling.

My brain may be failing. I can probably fix my problem by increasing the overcool factor, but that line of reasoning assumes this is an OR case, not an AND as you say

stratochief66:
With the changes, my tanks stay in the neighbourhood of 20K, but all excess cooling capacity goes towards cooling the Skylab drastically.

Currently partTemp 181.1K and falling. radTemp 311K, radovercoolFactor = 0.0948, partradMax = 0.292, partmaxTemp = 1073.15K

"target part temp > target part radiatorMax * target part maxTemp "

181.1K > 0.292 * 1073.15K is False, and only True if partTemp > 313K

"target part temp > radiator temp * radiator overcool factor"

181.1K > 311K * 0.0948 is True and will be True until the part reaches 29.5K

Currently cooling is occuring on the part (Skylab) while only one equation evaluates to True.

NathanKell:
Only thing I can think of is it's no longer the radiator sucking heat, it's the boiloff cooling the tank?

stratochief66:
Nope, that issue stopped when I assumed it was an OR, and adjusted the radovercool value to limit to 20K.

The part that keeps cooling the most is the Skylab, which contains no fuel to boiloff.

Starwaster:
Ok, stratochief66 just pointed me at this discussion. There could be interference from Real Fuel's code which *does* modify radiatorMax as follows:

```            part.radiatorMax = (minTemp * 0.9d) / part.maxTemp;
```
Where minTemp is the lowest tank.temperature defined in the module. The intent was only to apply it to cryogenic tanks but TANK.temperature defaults to 300 so every RF tank is going to have its part.radiatorMax set to 0.251596 (assuming previously cited maxTemp & rounded) or lower, regardless of how else that part might be configured. That can only account for temps over 300 though so I'm not sure why it would be cooling below that unless it's also interaction with overCoolFactor. (have to

I'll be scrutinizing RF's involvement here more closely but I won't be able to do much with it for the next two days.

NathanKell:
Ah, that might have something to do with it, yes.

Starwaster:
And... thinking about it further, the effect could be worse for parts that HAVE ModuleFuelTank but no actual tanks assigned.

Starwaster:
ugh found a worse bug in the RF code that modifies radiatorMax. Definitely looking like an RF issue. It's a two line fix. I'll have push an update to the main branch in a bit.

Starwaster:
Ok, this is fixed in Real Fuels (request sent and merged) for the next time an RF update is pushed to the public

NathanKell:
I promised @Saabstory88 I'd add a way to limit how many resource tanks one can add to an MFT part, so I'll do that tomorrow and then maybe we run a release?

Starwaster:
sounds good.

Schnobs:
To summarize our options:
> A radiator will pull heat if temp * headroom < maxtemp.

That means we can set a temp where the radiator shuts down. For current RO radiators, this happens at 1073*0.28=290K -- as soon as a radiator reaches that temp, it will no longer pull heat from anywhere.

> A radiator will then pull heat from parts where target part temp > target part radiatorMax * target part maxTemp

In other words, every part can "demand" to be cooled to a certain temperature. Until it reaches that temp, it will only radiate for itself.

> and where target part temp > radiator temp * radiator overcool factor.

This determines how good a fridge the radiator is. Set the overcool factor to 1 for a passive device that will at most become as hot as the part that it's cooling (actually should be slightly >1, if the code supports it. Heatpipes are good but not perfect).

With an overcool factor <1, the radiator will actively pump around heat, possibly cooling it's target parts to below ambient temperature.
--

How much a radiator can radiate is determined by physics: surface area, emissive constant, and (more than anything else) it's temperature. While we can tamper with all three in some way, we have no direct control over the outcome.

FWIW, some ballpark figures for blackbody radiators (we have emissive constant =0.9 on most parts, so subtract 10%)
- 290K: ~400W/m²
- 500K: ~3500W/m²
- 750K: ~17500W/m²

Radiators also pick up some radiation; all bodies radiate and the radiators only orient themselves away from the sun, if they can turn at all. For our ISS-style 290K radiators in LEO, this seems to be quite noticeable. 
--

Stock default maxTemp for parts is 2000K; RO resets this to 1073K. As most thresholds are defined as maxTemp * factor, changing maxTemp has side effects.

What's the default value for radiatorMax (adjusting the temp at which a part asks to be cooled)?

And, what happens if there is more demand for cooling than the radiators can give, how is their capacity distributed?

PR #785, opened 2015-12-20
Squad MMRTG update

PhineasFreak:
* Fix a major bug where KSP cannot output to a non - existent resource
(missing DepletedFuel resource) and the converter cannot operate.
* Adjust the consumption values so that the generator will last ~43
years (instead of several thousand), half the time of one half life of
Plutonium 238.


NathanKell:
Thanks! :)

PhineasFreak:
My pleasure!

PR #798, opened 2016-01-06
Stock fuel cell conversion rate similar to current FASA Gemini

Schnobs:
Output adjusted to be about Apollo-like by weight and size.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
@@ -586,7 +586,7 @@
 @PART[FuelCell]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	@description = A 750W fuel cell. Consumes about 41 liters LH2 / 29 liters LOX per week.
+	@description = A 750W fuel cell. Consumes about 41 liters LH2 / 29 liters LOX per week. These figures were determined by ground tests and include boiloff, but actual hydrogen boiloff may be considerably higher depending on situation and the surface:volume ratio of your tanks.
 	!MODULE[ModuleResourceConverter]
 	{
 	}
@@ -604,7 +604,7 @@
 @PART[FuelCellArray]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	@description = A 4.5kW fuel cell. Consumes about 245 liters LH2 / 175 liters LOX per week.
+	@description = A 4.5kW fuel cell. Consumes about 245 liters LH2 / 175 liters LOX per week. These figures were determined by ground tests and include boiloff, but actual hydrogen boiloff may be considerably higher depending on situation and the surface:volume ratio of your tanks.
 	!MODULE[ModuleResourceConverter]
 	{
 	}
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
@@ -571,10 +571,22 @@
 		@type = RealismOverhaulStackHollow
 	}
 }
+
+//fuel cells
+//Smithsonian exhibit: http://airandspace.si.edu/collections/artifact.cfm?object=nasm_A19730934000
+//"Normal power output for each power plant is 563 to 1420 watts, with a maximum of 2300 watts.
+// 111.8 x 55.9cm, 111.1kg (44 x 22 in., 245lb.)"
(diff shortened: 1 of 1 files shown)


stratochief66:
This looks like more of an unrealistic overhaul than what you had mentioned on IRC, although that could be the lack of coffee and lack of justifications notes talking.

On IRC, I said it made sense to change the LOX/LH2 ratio to match the FASA Gemini, since they were intended to be mirrors of each other.

Why all the other changes? Also, I don't think "making this more Apollo like" is a good thing to do to the Gemini hardware, as Gemini hardware is distinct from Apollo.

Schnobs:
This patch doesn't touch any Gemini parts, or Apollo for that matter. This is about the stock models.

I adjusted the conversion rate to be similar to that of the FASA Gemini; and adjusted their throughput to be about Apollo-like by weight and size. Hmm, I probably should have changed it to the Apollo conversion rate rather than Gemini: then again, do the two differ by much? You said the Gemini figures were based off Apollo.

NathanKell:
@stratochief66 poke :)

stratochief66:
@NathanKell 

The specs are changes, but the masses are not.

One advantage of the current FASA Apollo style as that it lets you dial up and down product as needs dictate (so, it can act as a 1/3 , 2/3, or full capacity 3 fuel cell unit. This doesn't have that, which might be nice to make it more widely usable for custom user payloads.

I would be more happy if they were reproductions of actual existing fuel cells, because then the specs can be straight-forwardly cited and reproduced.

Also, a more realistic generic Gemini tech level fuel cell would produce waste-waster, and the Apollo tech level would produce drinkable water.

These are probably good stand-in's, if you're ok with them not representing actual things.

stratochief66:
snb: if you feel the H2:O2 consumption ratio is off of real Apollo, but I don't see a spot where you cite the actual consumption ratios by volume or mass of Apollo. The values for both Gemini and Apollo need adjusting to match real consumption ratios, but that should be done with citations on hand.

NathanKell:
I'm ok with generics in this case; if someone wanted to make, say, the pre-moon Apollo proposals (D2, M410, etc), that would be useful.

Schnobs:
OK, let's recap: I noticed that these babies currently use H:O in a near-perfect 2:1 ratio and thought that this can't possibly be right. On top of that, the large module was 5x mass but only 3x output compared to the small one -- so why would anyone ever use the large module?

I changed them to the same ratio that can currently be found on FASA Gemini, under the assumption that this would be an improvement. At no time did I intend these to be replicas of the actual Gemini or Apollo fuel cells -- I just adopted the conversion rate, not questioning whether it was sourced somewhere (I assumed matter-of-factly that it was).

That the Gemini cells produce wastewater instead of water utterly escaped me. Then again, the FASA Apollo ones do make potable water -- and lo and behold, at just the same rate as what I pieced together here. What a coincidence! At any rate, the conversion rates in this PR are just as right or wrong as those currently found on FASA Apollo.

While I know nothing about conversion rates, the mass & power of real modules were easier to find. Seeing that the large one is about 2 Apollo fuel cells, I adjusted the power output accordingly. The small one was set to 1/6th that for reasons that should be obvious when looking at the models.

The PR doesn't give their mass (which is unchanged from the original stock models, so here you are: 50kg for the small model, 240kg for the big one). A strict application of the known ratio would lead to 830W / 5kW -- I chose numbers that are easier to work with, and also because I thought the PR would be more palatable if these parts were somewhat underpowered.

As to dividing the large model into several modules that could be toggled individually, the thought never occured to me. On the vessel that made me question these parts, I ended up using four of the small modules, and that's what I'd still recommend to anyone who can't quite use the output of the large one. (Also, until quite recently I was in the belief that fuel cells would auto-toggle.)

That said, the part descriptions are wrong. These numbers were found using timewarp on the pad, but when I used them on a lunar mission, I found that I needed considerably more hydrogen. Like, more than twice as much. What is a boy to do? I'd like to give users some guidance, but if that can't be done, it would be better to say nothing at all.

NathanKell:
Probably best to mention putting the LH2 in a servicemodule tank and even then adding some extra to account for boiloff?

stratochief66:
Good point. @Schnobs, can you add some of what Nathankell said to the description? The prefab FASA parts come with those resources, so the user doesn't have to think about all that before seeing it done for them, so it is less critical than here, IMO.

Also, thanks for the changes here Schnobs. Apologies if my pedantry earlier was bothersome.

NathanKell:
@Schnobs do let us know when you're ready to reopen this btw--we'd still very much like to merge it! :) It's just sat untouched for a while, so...

NathanKell:
@Schnobs can you maybe finish this off so I can put it in 10.9.1 ? :)

PR #803, opened 2016-01-06
KK_ATK updates from RP-0 issue 417

ctiberious:
The are changes as discussed in RP-0 issue 417.  The changes include:
1- New Castor 4 srb created from the existing Castor 4a


A1Ch1:
I notice the thrust curve here comes from the Altair config and the engine uses HTPB when the Castor 4 should use PBAN. I added a Castor 4 to KW with #784, could you use the stats from that config for mass, thrust, propellant, and thrust curve?

NathanKell:
@A1Ch1 as long as the density matches (and the specific heat is close) it should be good enough. :) Thanks for mentioning, @ctiberious that's a good place to draw from.

ctiberious:
I actually used the ATK version of the Castor 4A to generate the stats for the Castor 4, but I didn't do a whole lot of research beyond that.  I made the assumption (yeah, I know, bad thing to do) that it was basically just a weaker version of the 4A.  Didn't notice the new KW version of the Castor 4 until I was submitting everything.  Give me just a few minutes and I'll update the ATK version to match the KW version.  I'll also take a look at the ATK version of the 4A and make sure it matches the KW version as well.

ctiberious:
You know, at this point I don't know where the info for the Castor 4 came from.  I know the version I've been using in my game is based directly on the 4A, which is why it was using HTPB instead of PBAN (didn't realize they changed the fuel type).  But what I ultimately posted wasn't what I've got in game.  It is now and both the 4 & 4A match the KW Rocketry versions at this point.

NathanKell:
@A1Ch1 do these numbers match your data?

A1Ch1:
Yes, the Castor 4 matches. The changes to Castor 4A don't, but they'll both be replaced by RealismOverhaulEngines anyway once I've got the new thrust profiles for all the ATK motors generated.

ctiberious:
I didn't intend to make changes to the 4A and I know SirKeplan's changes will resolve the problem anyway but I've gone ahead and made sure the ATK version of the 4A matches the KW Rocketry version.  So now both the Castor 4 and 4A should match in both mods.

PR #804, opened 2016-01-06
Squad Engines updates from RP-0 issue 417

ctiberious:
The are changes as discussed in RP-0 issue 417.  The changes include:
1- Update the LR89 and LR105 engines to match the FASA versions
2- Adjust tech node placement for the LR101-11 engine config

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -915,8 +915,8 @@
 		CONFIG
 		{
 			name = LR79-NA-9
-			minThrust = 774	// 1961 NASA Launch Vehicle Handbook
-			maxThrust = 774	// 1961 NASA Launch Vehicle Handbook
+			minThrust = 774	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
+			maxThrust = 774	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
 			heatProduction = 100
 			PROPELLANT
 			{
@@ -931,8 +931,8 @@
 			}
 			atmosphereCurve
 			{
-				key = 0 284	// 1961 NASA Launch Vehicle Handbook
-				key = 1 245	// 1961 NASA Launch Vehicle Handbook
+				key = 0 284	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
+				key = 1 245	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
 			}
 			massMult = 1.03 // source: http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm (guesstimate between S-3D and LR79-NA-11)
 			cost = 100
@@ -944,8 +944,8 @@
 		CONFIG
 		{
 			name = LR79-NA-11
-			minThrust = 850	// 1961 NASA Launch Vehicle Handbook
-			maxThrust = 850	// 1961 NASA Launch Vehicle Handbook
+			minThrust = 850	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
+			maxThrust = 850	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
 			heatProduction = 100
 			PROPELLANT
 			{
@@ -960,8 +960,8 @@
 			}
 			atmosphereCurve
 			{
-				key = 0 286	// 1961 NASA Launch Vehicle Handbook
-				key = 1 248	// 1961 NASA Launch Vehicle Handbook
+				key = 0 286.2	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
+				key = 1 248.3	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
 			}
 			massMult = 1.08 // source:http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm
 			cost = 200
@@ -976,8 +976,8 @@
 		CONFIG
 		{
 			name = LR79-NA-13
-			minThrust = 868	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine
-			maxThrust = 868	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine
+			minThrust = 868	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine.  See RO/Github #804 for details
+			maxThrust = 868	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine.  See RO/Github #804 for details
 			heatProduction = 100
 			PROPELLANT
 			{
@@ -992,8 +992,8 @@
 			}
 			atmosphereCurve
 			{
-				key = 0 290	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine
-				key = 1 252	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine
+				key = 0 290	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine.  See RO/Github #804 for details
+				key = 1 252	// Guess based on b14643, astronautix, alternatewars and comparable LR89 Atlas engine.  See RO/Github #804 for details
 			}
 			massMult = 1.08 // source:http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm
 			cost = 500
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -640,7 +640,7 @@
 		name = ModuleEngineConfigs
 		type = ModuleEngines
 		modded = false
-		origMass = 0.844
+		origMass = 0.460
 		configuration = LR105-NA-3
 		CONFIG
 		{
@@ -673,7 +673,7 @@
 			minThrust = 366.1
 			maxThrust = 366.1
 			heatProduction = 100
-			massMult = 0.9443 // 413kg engine, source http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm for -5
+			massMult = 0.8978 // 413kg engine, source http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm for -5
 			PROPELLANT
 			{
 				name = Kerosene
@@ -705,7 +705,7 @@
 			minThrust = 385.2
 			maxThrust = 385.2
 			heatProduction = 100
-			massMult = 1.01185 // same source
+			massMult = 1.02174 // 470kg engine, same source
 			PROPELLANT
 			{
 				name = Kerosene
@@ -738,7 +738,7 @@
 			minThrust = 386.4
 			maxThrust = 386.4
 			heatProduction = 100
-			massMult = 1.01185 // same source
+			massMult = 1.02174 // 470kg engine, same source
 			PROPELLANT
 			{
 				name = Kerosene
@@ -915,8 +915,8 @@
 		CONFIG
 		{
 			name = LR79-NA-9
-			minThrust = 756.2
-			maxThrust = 756.2
+			minThrust = 774	// 1961 NASA Launch Vehicle Handbook
+			maxThrust = 774	// 1961 NASA Launch Vehicle Handbook
 			heatProduction = 100
 			PROPELLANT
 			{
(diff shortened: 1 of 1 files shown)


NathanKell:
@ctiberious why the doubling of mass?
Also, on the Isp: I think bnumbers has a mistake there, as I find it very, very hard to believe that despite large gains in Isp for the 89 the 79 did not change at all, despite increasing thrust massively. @ferram4  and I went around and around on this, and eventually figured it made the most sense to draw from comparable 89s. Do you have other sources on late (non-H-1) LR79s having the same Isp as the original S-3D? It seems so odd...

It's also just possible that the different gas generator arrangement on Thor vs Atlas had something to do with how any given chamber's Isp was recorded?

A1Ch1:
#### LR79-NA-9
For MB-3 Block 1 (LR79-NA-9),
**[Astronautix](http://www.astronautix.com/engines/mb31.htm) has:**
667.2 kN (150 klbf) SL/761.6 kN (171 klbf) Vac
250 s/285 s

**b14643 has:**
667.2 kN (150 klbf) SL/ 756.2 kN (170 klbf) Vac
250.4 s/283.9 s

**The 1961 NASA Launch Vehicle Handbook has:**
667 kN (150 klbf) SL/774 kN (174 klbf) Vac
245 s/284 s
mb3-b1 lr79na9 launch vehicle handbook1


#### LR79-NA-11
Astronautix doesn't have an entry for MB-3 Block 2 or LR79-NA-11.

**b14643 has:**
734 kN (165 klbf) SL/ 831.8 kN (187 klbf) Vac
250.4 s/283.9 s

**The 1961 NASA Launch Vehicle Handbook has:**
734 kN (165 klbf) SL/850 kN (191 klbf) Vac
248.3 s/286.2 s
mb3-b2 launch vehicle handbook

#### LR79-NA-13
For MB-3 Block 3 (LR79-NA-13),
**[Astronautix](http://www.astronautix.com/engines/mb33.htm) has:**
756.1 kN (170 klbf) SL/866.7 kN (195 klbf) Vac
253 s/290 s

This agrees with the **[Rocketdyne report from alternatewars](http://www.alternatewars.com/BBOW/Space_Engines/YLR-79-NA-13_Specs.pdf),** which has
756 kN (170 klbf) SL (+- 0.6%)
252.4 s SL (+- 0.3%)

These numbers include the vernier engines. Looking at just the thrust chamber and gas generator (thrust divided by fuel consumption of thrust chamber + gas generator)
754 kN (169.5 klbf)
251.7 s SL

**b14643 has**
756.2 kN (170 klbf) SL/ 858.5 kN (193 klbf) Vac
250.4 s/283.9 s

#### So
What b14643 has done is take the SL thrust, which is given (Rocketdyne referred to variants of their engines by klbf, so the MB-3 Block 1 would have been called the 150 klbf or 150K), estimated a vacuum thrust of 170 klbf (756.2 kN is exactly 170 klbf), then derived Isp (probably using burn time + prop mass of Thor-Delta). That Isp was then applied to the 165 klbf MB-3 Block 2 and the 170 klbf MB-3 Block 3 to derive thrust for those engines. But just by looking at the tables on b14643, it's nearly impossible to tell what is known information (SL thrust), what's pulled out of thin air (vacuum thrust) and what's derived using that data (Isp). Information from that site needs to be checked against other sources (both the values and especially the names - the entire MA and LR89/105-NA-XX series found there has barely any basis in actual designations, as I found while researching Atlas variants for Bluedog Design).

Astronautix has some of the same problems, but is usually better at citing sources and providing information as to how derived numbers were figured out (chamber pressure, expansion ratio).

Using the information above and preferring data from official sources:

**LR79-NA-9:**
774 kN Vac
245 s/284 s

**LR79-NA-11:**
846 kN Vac
248 s/286 s

**LR79-NA-13:**
868 kN Vac
252 s/290 s

The numbers have been rounded to three significant digits (accuracy to a tenth of a kN (20 lbf) is overprecision in hundred-klbf engines, additionally the Rocketdyne document specifies +- 0.6% which is 4.4 kN (1000 lbf)) and modified so that the most important given information (SL thrust = 150 klbf, 165 kibf, and 170 klbf) is accurate.

ctiberious:
@NathanKell I basically cut&paste the ModuleEngineConfigs data for the FASA LR105 over the Squad version.  I missed the fact that the origmass shouldn't have changed, though.  I'll correct that back and double check that the massMult gives the correct final engine weight.

As for the LR79, I'm fine with making the changes that A1Ch1 is listing.  I always thought the b14643 numbers were a little weird for this engine but since they seemed accurate for most others, and I've always been warned to take astronautix with a bag of salt, I figured what I was finding was accurate.  But A1Ch1's findings make a lot more sense.

ctiberious:
@A1Ch1 Only question I have about your numbers is the variation you're sometimes using.  For the LR79-9 you're using the numbers directly from the NASA handbook (wish I'd have found that weeks ago :) ).  But for the LR79-11 you're dropping the values from the handbook by about 4%.  If we're going to use the official source data for one config, shouldn't we use it for the other config as long as we have it?

A1Ch1:
Yeah the numbers are about 0.4% lower because I lopped off the decimal from the Isp and used that value to calculate vacuum thrust. Leaving the decimal in and using 850 kN instead of 846 kN is totally fine too :)

A1Ch1:
@NathanKell What engine is LR79-NA-3 referring to? The S-3D?

ctiberious:
I should point out that I did not include the LR79-NA-3 configuration on the MB3 engine in FASA.  That only includes the three actual MB3 configurations: LR79-NA-9, -11 & -13.  Should I add the -3 configuration to the FASA MB3?  Not sure if that configuration was ever used on a Thor launch vehicle.

NathanKell:
@A1Ch1 ah, that explains things! Since we're statting the LR79, though, not the MB-3 system, can you run the numbers on Isp and thrust with no verniers? We know the thrust of the verniers and their Isp ( http://heroicrelics.org/info/lr-101/lr-101.html ) and we know the total system thrust, so you should be able to calculate main-chamber Isp and thrust alone, right? And yes, the -3 is, I believe, the S-3D.

@ctiberious I thought the S-3D (perhaps as S-3? There's confusion about designation) was indeed used on the Thor missile, and perhaps some early Thor-Able flights.

ctiberious:
I haven't looked much at the thor missile but I hadn't noticed any Thor-Able flights with anything less than the -9.  Of course, documentation is so poor on some of this stuff, who knows.  :)  I'll work on adding the -3 as the base config for the FASA MB3 engine once I'm back at my main computer.

A1Ch1:
@NathanKell Yes, the numbers are already for the LR79 without the verniers - the NASA Launch Vehicle Handbook already broke the stats down by engine, and I subtracted the verniers from the Rocketdyne document on Alternatewars.

Regarding the S-3D, I can't find any documentation that refers to the engine as the LR79-NA-3 (a google search for that exact phrase gives results only here and in RP-0), so maybe we should just rename the config S-3D?

I haven't done much research into the S-3D specifically, but from what I understand it's the designation for the uprated [150 klbf Juno II/Jupter engine](https://books.google.com/books?id=hKYeBgAAQBAJ&pg=PA196#v=onepage&q&f=false), derived from the [Jupiter's 135 klbf S-3 engine](http://www.spacelaunchreport.com/jupiter4.html), and was [almost identical to the MB-3 Block 1 engine from Thor](https://books.google.com/books?id=hKYeBgAAQBAJ&lpg=PA196&pg=PA198#v=snippet&q=s-3&f=false). Its specifications match the LR79-NA-9:
rocketdyne large engine projects

It's likely that the major difference between the S-3 engines and MB-3 engines was that the [S-3 used vectored turbopump exhaust](http://www.spacelaunchreport.com/jupiter3.html). This makes sense since Thor had verniers while Jupiter didn't and would have needed some kind of roll control. The above image seems to indicate a major difference in the turbopump exhaust. Both were probably variants of what Rocketdyne referred to internally as the [135K](https://books.google.com/books?id=7m7Uq-B6RbYC&lpg=PA263&vq=135k&dq=atlas%20ma-5%20engine&pg=PA110#v=onepage&q&f=false), which is why they're referred to as Jupiter/Thor engines:
rocketdyne family tree


So my thinking is that the 75K (Navaho) was further developed into the 120 klbf Atlas sustainer engine, When the size of Atlas was reduced, a new 60 klbf sustainer (the 60K) was developed for Atlas and the 120 klbf engine continued on as three 150 klbf engines - the Atlas booster, the Thor booster, and the Jupiter booster. Early tests at 150 klbf found this thrust level unstable, so the engine was [derated to 135 klbf](http://www.spacelaunchreport.com/jupiter3.html) and its internal designation was thus 135K. Since the Atlas boosters were developed as a pair of engines, its heritage would diverge from the Jupiter/Thor engines, so its internal designation was the 150K.

When Von Braun was designing what would become Saturn, he used a cluster of Jupiter engines for the first stage (since he was involved with Juno). Rocketdyne developed these engines from the 135K, and therefore the H-1 was the descendant of the S-3 and M-3.

ctiberious:
I've added the LR79-NA-3 config onto the FASA engine over in #827  Assuming we're happy with the numbers for this engine, both of these PRs should be ready now.

NathanKell:
@A1Ch1 the Rocketdyne engines page at Alternate Wars has a link to a nice short monograph on the Jupiter LR79 variant if you haven't seen it, as well as the Navaho heritage (see below).

Per the PR, my understanding was that the initial variant of 150klbf engine on Thor (i.e. the LR79 component of the MB-x system) was indeed exactly comparable to S-3, excepting the exchange of the gimbaled exhaust duct with verniers (forming the MB-x system, rather than just an LR79 engine). Indeed, the MB-1 (per the above link) was likewise downrated to 135klbf, before (one presumes in MB-2) being rerated for a stable 150klbf, and then standardized for production as MB-3.

As to the development history, that largely jibes with my understanding. I'd add only that with the Navaho program Rocketdyne was already familiar with a booster-sustainer arrangement, and the development of the A-4 copy, the NAA75 as offshoot, and later the XLR43 (and then 79/89) were all linked. The advances of each team were presumably shared; it's no accident the final kero-burning Navaho engines are nearly indistinguishable in appearance and performance from the early Atlas/Thor/Jupiter engines.

Ah, one other thing: the H-1 was certainly a descendant, but it was also a quite extensive redesign for reduced cost and greater performance (and was, I believe, proceeding in parallel with, or perhaps forming the basis of?, the E-1 program, essentially a double-thrust H-1 which was the original choice for Juno V / Saturn).

Now, as to the performance numbers for the LR79-NA-9, 11, and 13: what then can explain their thrust uprating broadly matching advances in LR89 performance, but their specific impulse being essentially unchanged in comparison to the advancements in LR89 specific impulse? This is what had @ferram4 and me scratching our heads.

A1Ch1:
@NathanKell Re: LR79/89, part of the reason is that the "-NA-X" designations for the LR89 only tell half the story. These were designations given by the military, not Rocketdyne, and the engines did not get a designation bump each time they were uprated, so there were many different thrust and Isp ratings within each designation. For instance, the LR89-5 had versions rated at 150 klbf and 154 klbf while the LR89-7 had versions rated at 165 kilbf, 167.5 klbf, 168 kilbf, 185 klbf, and 188.5 klbf. Comparable thrust levels had comparable relative Isp gains: the 154 klbf was rated 248/288 while the 170 kilbf was rated at 251/290. It's not until 188 klbf or so that we see numbers like 259/294.

Re: E-1, that engine preceded H-1 development by several years, and while it was originally the preferred choice for what would become Saturn 1, it was decided it would be faster to uprate the 135K engines to 188 klbf than wait for the E-1.

Re: S-3, I only question the "-NA-3" designation. That's why I asked about it - I had never heard that specific designation, and a google search turned up nothing except RP-0 and RO, whereas I was familiar with the S-3 and S-3D, which have plenty of information available. If we mean S-3D, that's what it should be called. And then, since the config we have for it is exactly comparable to the MB-3 Block 1, being rated at 150 klbf, we might as well add a vernier for TP roll control and split it into a different engine with the 135 klbf S-3. If we want the early 135 klbf Thor engine without TP roll control, call the config MB-1:
thor mb1

NathanKell:
Huh, I was rather flipped around on the E-1 then!

Regarding the -3, it seems very odd to me that the joint designation system would skip 3; I can think of no other case where that was done. Perhaps a mooted model that wasn't produced? But meh, in the end I guess it really doesn't matter.
That said, regarding LRxx vs MB/MA/whatever, since in RO we generally have the engine apart from the verniers, and for commonality with other designated engines, we stick to the joint designation system where possible (as we probably should go back and do at some point for the X-405 [Navy-only rather than joint designation; I forget the joint designation--some even number one presumes!] and Bell 80xx [XLR81].) In this case I guess it would be -NA-1?

Regarding the Isp issues: it's true that there were all sorts of different subvariants, but still--they showed way more Isp gain over time, and earlier too. It seems very odd to me that the 89 branch was up to 290s for Atlas E/F (170klbf per chamber booster) in 1961 whereas that didn't come to the Thor side until years later (and was the final upgrade to the MB system!) Maybe Douglas simply didn't pay what Convair did?

ctiberious:
The reason the Atlas showed more ISP gains than Thor/Delta was because of the different engineering and launch goals for each LV.  Atlas was intended to have as much power as the technology could support.  It was basically always on the "cutting edge" for the times.  This meant as much thrust and as much fuel efficiency as possible, allowing for larger payloads and broader mission possibilities.  But it also meant less reliability.  The early Atlas (up to Atlas II) had about 88% reliability.
Thor/Delta, on the other hand, favored reliability and cost efficiency over power.  Thor/Delta took the proven advances of Atlas, but only the ones that were completely reliable.  That meant Thor/Delta engines tended to be less powerful and less fuel efficient, meaning they couldn't lift as large a payload and had more limited mission uses.  But it also meant that Delta had about 95% success rate.

A1Ch1:
Yes, and additionally Thor/Delta used solids for thrust augmentation and so wasn't as dependent on the power of its main engine. When they needed to lift more payload they were firmly in the "moar boosters!" camp. But Atlas performance gains were entirely tied to its engines, particularly the LR89 since it determined maximum liftoff weight.

NathanKell:
Bumping this since with the tree rework we are thinking of implementing S-3 and the early Atlas engines.

PR #808, opened 2016-01-11
Fixes for RZ.2 engine

Ravenchant:
-changed size from ~2.4 to 3.00m (3.05m in reality, but...)
-added gimbal
-fixed thrust (the historic static thrust figure was used for vacuum thrust)
-added config for uprated engine as used on Europa

It should now be usable, though the attach nodes are still weird.


NathanKell:
3.048m, and yes we use that not 3m :) Is the rest of the Tantares stuff not exact? @Niemand303

Ravenchant:
There, fixed:) The VA capsule used to have scaling problems, dunno if that's still the case, and IIRC Black arrow's stage 2 didn't gimbal. I'll take a look this weekend.

NathanKell:
Cool. Merging. @Niemand303 if other parts depend on 3m rather than 3.048m can you update those as well?

PR #810, opened 2016-01-14
Sstu updates 0.3.26

stratochief66:
Some (certainly not all) updates necessary to bring the RO up to date with the current version of SSTU.

A number of parts have changed names, and the deployable engine RL10's no longer use a custom module.

A number of issues still to be sorted, some I couldn't figure out.

The ICPS and HUS parts now seem to have the engine's separate, so they need to be moved from their current positions to sit proper and possibly be scaled.

There appear to be some duplicate engine clusters in the VAB, that might be conflicts between the code we have in RO to create the engine clusters and the code that is now in SSTU to generate clusters.

Many engine clusters now seem to be in a state of broken-ness. Most don't have the properly scaled thrust, many don't have the proper scaled mass, etc, etc. The J-2 x6 appears to be mostly sane, so I was studying that a bit, but to no avail.

@FFCJoseEduardo  ?


stratochief66:
#811

PR #817, opened 2016-01-16
Major overhaul of SSTU engine and cluster patches.

shadowmage45:
* Uses in-line patching on existing engine clusters rather than delete-and-recreate.
* Removes :FINAL block from most places (I don't think it is needed, and actually causes problems with other subsequent patches)
* Fixes real-plumes setup for those with configs.

### Known Issues:
* All engine clusters show up in the engines/propulsion tab even though their category is set to none in the cluster patches.  Something is patching -after- the clusters are created and patched and forcing them back into the engines/propulsion tab.  These can probably be fixed (if desired) by running a second patch later in the loading cycle and removing them from the tab again (e.g. @category = none).


stratochief66:
@shadowmage45 that might be this, which was a quick patch I threw together when the engine/propulsion category name changed around the 1.0 update. It can/should probably be modified to reduce/remove unexpected behaviour.

stratochief66:
I'll test this out shortly, much thanks again!

stratochief66:
First pass in game, looks excellent!

I believe there was previously a 6x and 7x J-2 cluster, but I could be mistaken. Either way, those should be easy to add if we want them.

PR #818, opened 2016-01-17
SSTU Touch-Up

stratochief66:
Fixed some outstanding bugs after @shadowmage45 's previous outstanding PR.

All previously configured engine clusters now tested as having the
correct masses, thrust, and burning RealFuels.


stratochief66:
Tested, working nicely.

http://imgur.com/a/5w7nZ

PR #809, opened 2016-01-14
RO Filter Extensions

PhineasFreak:
* Simplify the FilterExtensions patches.
* Prepare the ground for the RP-0 FilterExtensions categories.

PR #824, opened 2016-01-18
RO FilterExtensions

PhineasFreak:
* Simplify the FilterExtensions patches.
* Prepare the ground for the RP-0 FilterExtensions categories.


PhineasFreak:
Remade from #809.

PR #827, opened 2016-01-19
LR79/MB3 engine update

ctiberious:
Update engine data for the LR79 to match changes proposed to the Squad version of this engine in #804

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Other.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Other.cfg
@@ -85,13 +85,13 @@
 	@maxTemp = 1973.15
 	@MODULE[ModuleEngines*]
 	{
-		@minThrust = 774	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
-		@maxThrust = 774	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
+		@minThrust = 758.7
+		@maxThrust = 758.7
 		@heatProduction = 100
 		@atmosphereCurve
 		{
-			@key,0 = 0 284	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
-			@key,1 = 1 245	// 1961 NASA Launch Vehicle Handbook.  See RO/Github #804 for details
+			@key,0 = 0 282
+			@key,1 = 1 248
 		}
 		@PROPELLANT[LiquidFuel]
 		{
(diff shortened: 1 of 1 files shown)


A1Ch1:
@ctiberious This isn't related to your changes, but I noticed

**configuration = MB-3-1**

could you fix that while you're at it so it points to LR79-NA-9 instead?

ctiberious:
Done

ctiberious:
I've added the LR79-NA-3 config into the MB-3 and set it as the "configuration = " line.  Stats for this config are pulled directly from the Squad version.

A1Ch1:
The S-3D shouldn't be on this engine since Thor used verniers for roll control while the S-3 had vectored turbopump exhaust and was used on Jupiter.

NathanKell:
@A1Ch1 the designations seem all over the map last I checked; what seems clear is that the first Thor missiles did use an engine of that vintage. (I've seen it called S-3 instead of S-3D, or some sources claim both were S-3D, or...it's murky). Whatever its actual name, we should be representing the first-generation LR79.

A1Ch1:
@NathanKell We already have a 150 klbf config in the NA-9, If we want an earlier version rated at 135 klbf as a predecessor we should be able to find enough information to do that, but I'm not comfortable calling it the NA-3. Of course since names were all over the place I could be mistaken, but I've never seen any indication that particular designation was ever used. Since there are many sources linking the S-3 and S-3D specifically to Jupiter/Juno II, I propose:

* pasting on a vernier to act as TP exhaust, as with the CryoEngines RS-68 or the AIES LR91, and spinning off the S-3 for Jupiter/Juno. Configs at 135 klbf for S-3 and 150 klbf for S-3D.

* Thor engines at 135 klbf, 150 klbf, 165 klbf and 170 klbf to give a nice progression all the way from early tests of the IRBM, to the operational IRBM and early boosters, to the later Thor/Delta variants.

 * H-1 variants at 188 klbf, 200 klbf and 205 klbf

 * RS-27 and RS-27A.

ctiberious:
Everything I'm seeing shows the Thor PGM-17 (the Thor IRBM) using the LR79-NA-9 along with a pair of LR101-NA 4.5kn Vernier engines.  I can't find any references to the Thor IRBM using anything else.  I do find a couple instances where the main engine is listed as "Rocketdyne LR79-NA-9 (Model S-3D)" but the thrust is definitely the 150klbf version.  I'm thinking we should just limit this LR79 engine to have the NA-9, NA-11, NA-13, RS-27 and RS-27A configs.

NathanKell:
Given what @SirKeplan is doing with the configs, it probably even makes sense to not combine LR79 and H-1/RS-27 configs, really. That would leave us starting with the -9 and ending with the -13 I guess, unless we add a derated 135klbf version (I'm not exactly sure where to place that in the tree on the RP-0 side).

ctiberious:
That makes sense on the FASA side since it has a separate H-1 that already includes the RS-27 and RS-27A configs.  I'll pull those two configs out of the the FASA MB3 engine.  On the Squad side, though, we might need to add a second engine since we've got just a single engine with all the LR79, H-1 and RS-27 configs together.

SirKeplan:
@ctiberious already added a new engine in squad configs in my PRs
You should see it in #830

PR #829, opened 2016-01-20
RO-EngineConfigs new engine files.

SirKeplan:
checked out from main branch for easier review/merging.
First PR #778 
Squad engines - #830
CryoEngines  -   #831 
#832 #833 #834 #835 #837  #838 #839 #840 #846 #847 #848 #851 #852 #861 

EDIT:
Noted diffs:
TR-201 - RLA has diff thrust. mass ratio isp etc to the others.
RD-180 - very minor variation in isp, noted in Baha
RD-0210 - files disagreed on O:F mix ratio, 2.6 looks to be correct one.
RD-0213/4 - similar mix ratio disagreement,
RD-0214 -  isp different on, 293 or 298?. also check mass.
LMDE - mass looks to be 0.179, but had been set to 0.135 one place.
LR79-NA-13 - entry cost subtracter 3000 or 4000


PhineasFreak:
Just to be on the safe side: shouldn't the "ignoreMass" field better be applied as a HAS[#ignoreMass[True]] check instead of just checking for the parameter existence?

SirKeplan:
@PhineasFreak yeah, probably. it really doesn't make any difference to making the configs, but it's better practice.
will do that soon.

stratochief66:
Are these safe to merge now @SirKeplan, or do you plan to keep altering this?

Also, @NathanKell if you're ok with this being merged.

SirKeplan:
@stratochief66 @NathanKell, I'd say it's safe to merge now, I’ve noted a few irregularities I noticed at a few points above though.

PhineasFreak:
The only "issue" that i can find is that we need two different versions of the RS-88, one with the original propellants and one with hypergolics (MMH & MON3).

SirKeplan:
@PhineasFreak ah great, um, we can just have multiple part variants of course. do you have a source stating the mass of the RS-88, we currently 0.22 and 0.25, and I expect they should be the same.

PhineasFreak:
Unfortunately no, and i searched around a lot as i was making the configs for the CST-100. For it, i assumed a mass value of 50 Kg.

SirKeplan:
50 Kg? did you mean 250 Kg?

PhineasFreak:
Per engine. The CST-100 application uses 4 of these (and the CST-100 model for the RS-88 has 4 of them together).

SirKeplan:
Let me get this right, the RS-88(or LAE as i've seen it called on the CST-100) has a thrust(per engine) of ~220 kN. so if 1 engine weighs 50 kg, then it would have a TWR of  220/(0.050*g) = 449, that would be unprecedented, and uh, highly unlikely.

PhineasFreak:
I never did any math for them so you can go ahead and correct anything that you see out of place.

stratochief66:
Ahh, you've already made a change I was about to request on #851 :)

Could you freeze changes to this so we can merge it? Other configs and changes could be made to these configs in a later PR, of course.

PR #830, opened 2016-01-20
RO-EngineConfigs Squad

SirKeplan:


stratochief66:
Too many lines changed for Github to show me the differences. @NathanKell do you think this is worth a "merge and pray", or should this be broken into multiple PRs?

NathanKell:
I would suggest breaking it up, or at the very least re-applying fresh. I had to do a fair bit of surgery after the merge of @ctiberious's changes, which were themselves considerable.

SirKeplan:
in 3 parts now: #879 #880 #881

PR #836, opened 2016-01-21
SRM Configs: KW

A1Ch1:
Configs for SRMs found in KW. Separate PR for KW itself


SirKeplan:
ah, cool!

SirKeplan:
Eh, sorry @A1Ch1 your going to have to change `:FOR[RO-EngineConfigs]:BEFORE[RealPlume]` To `:FOR[RealismOverhaulEngines]` in your PRs
NathanKell showed me i was using ModuleManager slightly wrong, what i had before before worked fine, but this is safer/more robust.

A1Ch1:
Alright :) I finally got around to putting together a program to automatically extract thrust curve data, so I'm taking the opportunity to go through the thrust curves again anyway.

stratochief66:
Looks good, one things I've noticed.

Should we include " !MODULE[ModuleEngineConfigs] {} " near the beginning of the templates?

@SirKeplan as well.

SirKeplan:
@stratochief66 why does that need to be included for these configs? i mean i see no harm adding that line for the solids. unless you mean in general, then that would cause issues with verniers i guess.

NathanKell:
We need to be careful about our clones then, if it's not included.

NathanKell:
To clarify further, there are two cases we need to worry about:
Case 1: The engine config is being applied to a part that was (a) cloned from an existing RO part after it has been RO-patched, and (b) that existing part is not using the new engine config system. In this case, we need to remove any existing ModuleEngineConfigs.
Case 2: The engine part has multiple engines on it. This will necessitate *not* destroying any prior ModuleEngineConfigs.

I would suggest that the best approach is to make sure Case 1 never exists, and then we will never need to worry about !MODULEing.

SirKeplan:
@NathanKell agreed, should be pretty easy to ensure Case 1 doesn't happen i believe.

A1Ch1:
In this case, the clones are all cloned from parts with engineType, so %engineType was sufficient. I did test the parts with !ModuleEngineConfigs just in case, to verify nothing changed.

stratochief66:
What is the state of this PR, good to go?

It might be good to cite which RO file you took the thrust curve from, and if you have found any disagreements/discrepancies between different implementations of these parts.

A1Ch1:
These are good to go :) Added thrust curve sources where I've gone back and re-verified (the Castor and GEMs). There are no discrepancies yet since they're only for KW parts so far.

PR #838, opened 2016-01-21
RO-EngineConfigs SSTU

SirKeplan:


stratochief66:
I'll be damn impressed if this works. 

Will your code also take care of the part mass change?
https://github.com/KSP-RO/RealismOverhaul/pull/838/files#diff-41e892c02cedcca213b70857e4776cfcL309

stratochief66:
For this engine, was RL10A-4 the only config you didn't remove?
https://github.com/KSP-RO/RealismOverhaul/pull/838/files#diff-f9e9a8a296b479b65977b4899e58b577L48

stratochief66:
If this code works, it should definitely be the standard for keeping only one config
https://github.com/KSP-RO/RealismOverhaul/pull/838/files#diff-e5e07d36684f087d10803c3f024fb64dR80

SirKeplan:
@stratochief66 yes, mass is all handled correctly. For the RL10, I believe that engine is just the RL10A-4 variants with small nozzle extensions(RL10A-4 and RL10A-4-1/2), if any other variants have short extensions then they should be included too.

SirKeplan:
#904 makes this obsolete

PR #844, opened 2016-01-21
CO2 Scrubber for Skylab

ctiberious:
Adding a TACLS CO2 Scrubber to the Skylab since I have to assume it had one.  I've used the same scrubber that we have on the FASAApollo_CM part, though I've adjusted the conversaionRate up to 6 to account for the 6 crew capacity on the Skylab.  This will result in twice as much CO2 and LithiumHydroxide being used, with twice as much Waste being generated.

I've also reset the LS quantities so there should be 300 days worth of LS (1 day worth of CO2), including running the scrubber.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -79,8 +79,8 @@
 			TANK
 			{
 				name = ElectricCharge
-				amount = 30000
-				maxAmount = 30000
+				amount = 100000
+				maxAmount = 100000
 			}
 			TANK
 			{
@@ -97,54 +97,48 @@
 			TANK
 			{
 				name = Oxygen
-				amount = 1065312
-				maxAmount = 1065312
+				amount = 177552	//50 days for 6 crew
+				maxAmount = 177552
 			}
 			TANK
 			{
 				name = Water
-				amount = 6967.3
+				amount = 6967.3	//300 days for 6 crew
 				maxAmount = 6967.3
 			}
 			TANK
 			{
 				name = Food
-				amount = 10528.7
+				amount = 10528.7	//300 days for 6 crew
 				maxAmount = 10528.7
-			}
-			TANK
-			{
-				name = LithiumHydroxide
-				amount = 1849.1
-				maxAmount = 1849.1
 			}
 			TANK
 			{
 				name = CarbonDioxide
 				amount = 0
-				maxAmount = 3069
+				maxAmount = 3069	//1 day for 6 crew
 			}
 			TANK
 			{
 				name = Waste
 				amount = 0
-				maxAmount = 6940.9
+				maxAmount = 1380.3	//300 days for 6 crew including CO2-O2 Converter
 			}
 			TANK
 			{
 				name = WasteWater
 				amount = 0
-				maxAmount = 8864.6
+				maxAmount = 8864.6	//300 days for 6 crew
 			}
 	}
 
 	MODULE
 	{
 		name = TacGenericConverter
-		converterName = CO2 Scrubber
+		converterName = CO2-O2 Converter
 		conversionRate = 6.0	// # of people - Figures based on per/person
-		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010, LithiumHydroxide, 0.00001189
-		outputResources = Waste, 0.00003847, false
+		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010	// See RO Github #844 for explanation of values
+		outputResources = Oxygen, 0.0071565375, true, Waste, 0.0000027155975, true	// See RO Github #844 for explanation of values
 	}
 
 	@MODULE[ModuleRCS]
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -97,39 +97,54 @@
 			TANK
 			{
 				name = Oxygen
-				amount = 1070000
-				maxAmount = 1070000
+				amount = 1065312
+				maxAmount = 1065312
 			}
 			TANK
 			{
 				name = Water
-				amount = 7700
-				maxAmount = 7700
+				amount = 6967.3
+				maxAmount = 6967.3
 			}
 			TANK
 			{
 				name = Food
-				amount = 7384.10
-				maxAmount = 7384.10
+				amount = 10528.7
+				maxAmount = 10528.7
+			}
+			TANK
+			{
+				name = LithiumHydroxide
+				amount = 1849.1
+				maxAmount = 1849.1
 			}
 			TANK
 			{
 				name = CarbonDioxide
 				amount = 0
-				maxAmount = 1040
+				maxAmount = 3069
 			}
 			TANK
 			{
 				name = Waste
 				amount = 0
-				maxAmount = 893
+				maxAmount = 6940.9
 			}
 			TANK
 			{
 				name = WasteWater
 				amount = 0
-				maxAmount = 735
+				maxAmount = 8864.6
 			}
+	}
+
+	MODULE
+	{
+		name = TacGenericConverter
+		converterName = CO2 Scrubber
+		conversionRate = 6.0	// # of people - Figures based on per/person
+		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010, LithiumHydroxide, 0.00001189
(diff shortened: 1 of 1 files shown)


stratochief66:
This is probably a good stop-gap. At some point, RN removed my similar code for reasons I don't know.
https://github.com/ctiberious/RealismOverhaul/commit/7ef9f7bb3efff82d317d6c4e78885e1428735a75

In reality, I believe the Skylab actually had an oxygen re-generator (so, no Lithium Hydroxide spent) but implementing that would require a source. If you want to jump to researching that, I'd appreciate it. If not, this is probably a 'good enough' stopgap that is much appreciated as well.

stratochief66:
Ahh. Also, the Skylab would also require a LOX -> O2 generator in this scheme, like the Apollo has, and Skylab did previously:

https://github.com/ctiberious/RealismOverhaul/commit/7ef9f7bb3efff82d317d6c4e78885e1428735a75#diff-bac2c3d741bfcba558a7b22307dabc93L162
 
Very much worth watching: https://youtu.be/JsppUximlcA?t=10m56s

ctiberious:
I thought about a LOX-O2 generator.  I know that worked on the Apollo, but Apollo only needed to run for a couple weeks.  Skylab was occupied for over 170 days (though I know not all in one stretch) and probably could have been occupied for even longer stretches if the political climate had been better.  Because of the longer duration, I was concerned about the LOX boil off.  Adding such a generator would be simple enough but did they really use Liquid O2 or did they take up compressed O2?

Also, that was a great video.  I didn't realize we had a nearly perfect CO2-O2 converter back in the Gemini/Apollo era.  I'm guessing it was too bulky for Apollo (which I know used LH canisters) but if it was used on Skylab, it would be very easy to setup the "CO2 Scrubber" to use CO2 & EC and produce O2 and Waste (Carbon).  Just have to determine what efficiency we want out of it.  I'm guessing not 100% but the video made it seem like the system was pretty close to that.  Do you think 99.5% is too efficient?

raidernick:
@stratochief66 I removed it because tacls had a problem with the co2 scrubbers that caused staging problems with anything it was on. @NathanKell tells me this is now fixed so I guess you can add it back. 

I also removed the co2 scrubbers that were there because Skylab uses a molecular sieve not lithium hydroxide canisters.

stratochief66:
Ahh yes, thanks for the reminder @raidernick 

@ctiberious, try something high like that and see how it goes. Make sure that you still conserve mass.

ctiberious:
So first, since the Skylab didn't include LOX, but used compressed O2 instead, I'm not going to include a LOX-O2 converter.  Instead, I'm going to assume that the Oxygen TANK represents both the breathable O2 actually in the air, plus the extra compressed O2 in storage.  I'm going to set the default so that Skylab includes approximately 50 days worth of O2 for a 6 Kerbal crew.  Assuming we wouldn't have apparatus to store "spare" CO2, I'm setting that TANK to hold 1 days worth for 6 kerbal.  Water, Food, Waste & Waste Water will have 300 days.  And Waste will include space for the waste generated by the CO2-O2 Converter.
For the CO2-O2 Converter (if there is a better name for that, let me know) I'm going to use the following numbers (these are per kerbal):
CO2 = 0.006216 (input)
EC = 0.010 (input)
O2 = 0.0071565375 (output)
Waste = 0.0000027155975 (output)
We want the converter to pull more CO2 out of the air than is actually generated so that built up CO2 can be cleaned.  I've therefore set the CO2 input value at 105% of what a kerbal actually generates per second.
I've left the EC value the same as what it was on the Apollo CO2 Scrubber, mostly because I have no clue how much EC this converter should really use.
We want to try a 99.5% efficiency on the converter so I set the O2 output value at 104.475% of what a kerbal actually generates.
As for waste, I came up with the above value by calculating how many kg of CO2 the converter would use in 1 hour based on the density value in CommonResources.cfg (134.2656 units * 0.000001951 = 0.0002619521856kg CO2).  Then removed weight of the O2 the generator produced in 1 hour (154.58121 units * 0.00000141 = 0.0002179595061kg O2).  That means I've got 0.0000439926795kg of waste each hour which I divided by the density of Waste shown in CommonResources.cfg (0.0000439926795 / 0.00075 = 0.058656906 units).  I then converted that to a per second value (0.058656906 / 6 crew / 60 minutes / 60 seconds = 0.0000027155975 per second).
The converter will only create O2 when there is the requisite amount of CO2 available for a cycle so the end result is a tiny outflow of O2.  A couple of ingame tests show that I'm using about 0.00021 O2/s with 6 kerbal and the converter running.  Which means the 50day of O2 will last for about 26.8 years (9785.7 days) which means the converter is managing about 99.489% efficiency.

raidernick:
you should use the published amount of o2, water and food not make up numbers

Launch masses:

Oxygen:  2,240 kg
Water: 1,166 kg
Food: 670 kg

stratochief66:
@raidernick 

If you're going to be curt, have the decency to include a citation as well.

raidernick:
http://pages.erau.edu/~ericksol/projects/futurspcrft/sp/intro.html

stratochief66:
@ctiberious your logic looks sound. Could you alter the amount of O2, H20, and food at launch to reflect the sourced information from raidernick?

I'm less certain about much much maximum water and food the Skylab could store, vs "amount at liftoff", but that could be revisited some other time?

ctiberious:
Do either of you know, is the density value in CommonResources.cfg the density per gram?  Or is it density per kg?

stratochief66:
not offhand, but water is a good reference to check. water is 1 Tonne (1000 kg) per cubic metre, so I use that for quick checking of the scale used.

ctiberious:
That shouldn't be a problem.  I just set everything to 300 since the current RO setup comes with something like 301.3 days worth of O2.  But I can reset to the actual launch values.   That said, do either of you have an issue with leaving the maxAmount at the higher value?
Also, do we have any idea how much waste and wastewater the Skylab could handle?
And what was the maximum capacity of the batteries?

stratochief66:
@ctiberious I'm fine with a larger maxAmount, I'm not aware of any citations that give us precise values to use though.

My link to my previous Skylab config has some of the values I estimated previously, if you want to see them. Some is obviously just boilerplate/filler, but other parts might be useful:
https://github.com/ctiberious/RealismOverhaul/commit/7ef9f7bb3efff82d317d6c4e78885e1428735a75#diff-bac2c3d741bfcba558a7b22307dabc93L162

raidernick:
I have the battery values 7000 amp-hours(20 batteries at 350amphour each) * 28 volts is 705600 kJ

http://pages.erau.edu/~ericksol/projects/issa/skylab.html#Skylab%20Systems

ctiberious:
I should also point out that the 1166kg of water will not last 170 days.  According to the link raidernick provided, NASA assumed about 5lb/person/day but TACLS actually uses about 8.5lb (3.87072kg) per person per day.  So 1166kg will only last a crew of 3 about 100 days.  And there is a line on that link that states: At the time Skylab was launched, the water stored onboard weighed 2,700 kg (6,000 lbs.).  Not sure if that meant some of the equipment (perhaps the CO2-O2 converter) used the other 1534kg of water.

And the 670kg of food was meant to least 3 people for the entire 170 day mission.  But TACLS uses about 1.64kg/person/day meaning there will only be about 68 days worth of food for a 3 man crew.

raidernick:
@ctiberious check the new link I sent, it has values for food water and oxygen too that may differ from the first link.

ctiberious:
Hmm.  705600kJ.  If kW = kJ / s, does that mean we should set ElectricCharge to 705600?

raidernick:
yes set it to 705600

ctiberious:
So it looks like there was actually closer to 2700kg of water, though some was used for purposes other than drinking.  That said, I'm going to set the tank to hold 2700kg.  The numbers I end up with are:
EC 705600
O2 1588652.5
H2O 2700
Food 2384.1
CO2 3068.9
Waste 391.1
Waste Water 2511.7

Waste and Waste Water can hold 170 days worth from a 3 man crew.  Since Skylab total mission length was intended to be 170 days with a 3 man crew, I figured this made the most sense.
CO2 is 1 days worth of a 6 man crew.  I suppose technically it should be 10% of the total available volume inside the Skylab but I'm not sure how to calculate that nor am I sure that TACLS could handle that level of accuracy.
670kg of food will actually allow 3 kerbals to live for about 136 days so it's not quite enough to cover the full 170 day mission.
2700kg of water support just over 232 days of mission time.

My real concern is Oxygen.  2240kg will support 3 kerbal for over 894 days.  A person consumes about 550L/day of pure oxygen which works out to something like 0.00637L/s so slightly less than TACLS says a Kerbal uses it.  I suppose 550L is really an average or even just a minimum so I don't have a complaint about the TACLS value of 0.00685L/s.  But it does make me wonder.  If Skylab had this great CO2-O2 converter, why did they need to carry nearly 3 years worth of extra Oxygen with them?  Guess they either didn't trust their life support system or didn't want to risk ever running out.

PR #847, opened 2016-01-22
RO-EngineConfigs SXT

SirKeplan:
Liquid Engines apart from Rd-107/108


stratochief66:
Do you expect this to be a case that occurs often?
https://github.com/SirKeplan/RealismOverhaul/commit/923ed8cd72298dd386955b55b4d75df8b1272482#diff-ec7fbe408ec240413534e039cdad2a9aR64

Perhaps it would work instead to use an NK-15/33 template, and use engineMult 30?

SirKeplan:
Regarding nk15/33, yeah possibly, I did that one before i had the multipler setup.  However there's the issue of the minthrust value, the N1 would cut off some engines instead of throttling down. Probably a minor issue.

stratochief66:
Yeah, that is tricky, and an issue worth discussing since there are more mods with engine clusters every day (ie. SSTU) that don't have a special module for disabling engines the way we would like.

The verniers get sorted by refinement, like that has been done for the Atlas (thank you, previous RO'ers) where super research was done to research the thrust and ISP of verniers, and all the separate part values were found. Additionally, some parts model the verniers and others don't (for other parts)

IIRC, the first and second N1 stages would throttle engines on one side or the other to provide control, rather than gimballing (?)

stratochief66:
#870

PR #851, opened 2016-01-22
RO-EngineConfigs FASA Titan Agena

SirKeplan:


stratochief66:
Looks good!

The only thing that seems a bit strange is this:
https://github.com/KSP-RO/RealismOverhaul/pull/851/files#diff-4272427f87e810f628186d14c801ef8eR255

I looked, that is certainly the name of the config for the Agena flight pack in FASA. Does anybody else know why? @NathanKell?

That can probably be fixed fairly easily later through a re-naming to something more fitting, no biggy.

NathanKell:
It should be AgenaSPS. The 'flight pack' is a combination of Agena's cold-gas attitude thrusters and the Agena Secondary Propulsion System (SPS). Bladder-fed ullage / in-orbit orbit adjustment rockets.

SirKeplan:
huh, that is weird naming, never saw that, should be fixed.

stratochief66:
@SirKeplan, could you modify the name here to match your new template name on #829? Then, I think it is good to merge.

SirKeplan:
@stratochief66 it's already been changed to `AgenaSPS`

stratochief66:
Yep, indeed. Thanks!

PR #852, opened 2016-01-22
RO-EngineConfigs FASA Apollo LEM Redstone Other

SirKeplan:


SirKeplan:
@NathanKell @ctiberious.
still need to do the MB-3 in FASA/RO_FASA_Other.cfg, as it is the MB-3 has some LR79 configs, and 2 H-1 configs(RS-27s) should i delete the RS-27s and have the MB-3 as an LR79? or should something else be done, like merging the LR79, H-1 and RS-27 into one EngineConfigs file? (this might make it hard to choose a default config for each one)

SirKeplan:
ah @ctiberious i should have paid more attention to your PR, so removing the RS-27 should be good then?
Also, we seem to be each modifing the same code here again.. :P need to sort this out.

ctiberious:
@SirKeplan Yup.  I think we've pretty well sorted what the values for the LR79 should be between #804 and #827 though both of those are updating more then just the configs.  Most of what I have been doing involves moving parts and configs around on the tech tree.  The stuff I discussed in issue 417 over on the RP-0 Github.  The LR79 config changes mainly came because we decided to make the FASADeltaMB3LFE a viable engine for Thor/Delta recreations.

ctiberious:
Oh, and yes.  I think we finally concluded that FASADeltaMB3LFE should only include the LR79 configs while the FASAApolloLFEH1 part would handle H1/RS-27.

stratochief66:
Sorry, what is going on here?
https://github.com/KSP-RO/RealismOverhaul/pull/852/files#diff-80fff832393ca834aa3db42ad59e8750R126

3 maxThrust values, no @ or other MM modifier?

SirKeplan:
@stratochief66 uuuhhhmmm, yeah that shouldn't be there.

stratochief66:
@NathanKell good to go, beyond the Merge conficts

PR #861, opened 2016-01-24
RO-EngineConfigs RLA Stockalike

SirKeplan:


stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/861/files#diff-1cff77fd4cc218b818014422ce984199R413

should it be R4D11 ? I don't see the additional 1 anywhere.

If changes, that would need to be changed here too:
https://github.com/KSP-RO/RealismOverhaul/blob/d0a3810dfc48a2cdeb876067101290c2f311f5dc/GameData/RealismOverhaul/Engine_Configs/RO_R14D11_Config.cfg

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/861/files#diff-1cff77fd4cc218b818014422ce984199R811

For something like this, I think the unit of engine should be 1 vernier, not 1 set of 4 thrusters. What do you think?

SirKeplan:
the RD0214: yeah, tricky one this one, the part in RLA is actually a fictitious engine based on the RD-0214, the real thing is 1 single engine with 4 nozzles for steering, and is used in tandem with the RD-0213, it's the russians trying to confuse us again with their funny engines :D http://www.russianspaceweb.com/rd0212.html

so IMO the unit should be 1 engine, as it is now.

SirKeplan:
what's the problem with the R4D11? it looks fine to me. did you misread something by any chance? or is it just me not noticing something obvious?

stratochief66:
Possibly, the template is called [R14D11]
https://github.com/KSP-RO/RealismOverhaul/blob/d0a3810dfc48a2cdeb876067101290c2f311f5dc/GameData/RealismOverhaul/Engine_Configs/RO_R14D11_Config.cfg#L4

SirKeplan:
@stratochief66 ah *facepalms*

yeah, that is wrong, i couldn't see that extra 1......

PR #868, opened 2016-01-28
RO RSB Ares I

PhineasFreak:
* Add RO compatibility for the Real Scale Boosters Ares I.

NOTE 1: Requires the new global engine configurations to operate.


stratochief66:
https://github.com/PhineasFreak/RealismOverhaul/commit/742deec2cec0145cdbfcf442eaba896ae89cf39b#diff-ed3a5b63593312a53a2ac441d6c782e0R42

You missed labelling this part RO compatible, possibly other ones. please take a look through for other missing ones.

%RSSROConfig = True

PhineasFreak:
It should be fixed now.

stratochief66:
The flare appears to be a bit far up, coming out of the top of the engine bell with this value, could you change the offset to 0,0,0 instead of ... -0.85? that looks about right to me.

https://github.com/PhineasFreak/RealismOverhaul/commit/742deec2cec0145cdbfcf442eaba896ae89cf39b#diff-11be7cd4a1e6fbdfafa09c5ca4770e3fR13



Also, if you scale the J-2X plume particles up to between 2 and 3 it looks much better with no performance hit, since it is still the same number of particles. That range looks to make the plume about the same size as the stock reconfig'd engine and the FASA one.

https://github.com/PhineasFreak/RealismOverhaul/commit/742deec2cec0145cdbfcf442eaba896ae89cf39b#diff-11be7cd4a1e6fbdfafa09c5ca4770e3fR13

stratochief66:
I also have a rotation of about 1 rev every 12 seconds, the Ares I upper stage RCS doesn't seem to be able to put a dint in that, no real roll control. Nothing I can suggest how to fix, just an observation, no big deal to me.

PhineasFreak:
Wrong notation, it should be +0.85...damn these typos...

Yes, the upper stage is a bit tricky to control. It is because of the low thrust of the RCS.

stratochief66:
I created the bad roll myself, bad decoupling. The lower stage was actually able to kill some roll I picked up, bumping the launch tower.

PR #876, opened 2016-01-29
RN proton fixes

raidernick:
-Update proton engines configs
-Update proton tank configs
-Update proton weights for extra fuel not taken into account originally
-Update block d to have multiple versions
-Update craft files for changes


raidernick:
@NathanKell this is finished and can be merged

PR #877, opened 2016-01-29
RO-EngineConfigs Merlin from Laztek,RealFuels,Starshine

SirKeplan:
consolidate Merlin1 configs into just 1 file.


SirKeplan:
hold of on merge a bit.

SirKeplan:
@NathanKell @stratochief66 @PhineasFreak how does this look?using this as a demo kind of, for the first config that is fully globalized.

Are there bits that should be changed?

NathanKell:
This looks good to me! I would suggest two slight changes, however:
1. I'd put the ignitions line with the IGNITOR_RESOURCE blocks, as things are they're separated by other stuff in between.
2. I'd make the 1-ignition variants use 4 units of TEATEB per ignition, so there isn't excess TEATEB floating around. Now, ideally there would be a way to change RESOURCE nodes from within a CONFIG, but there isn't...so that seems better than having 1-ignite Merlins hanging around with 3 units of TEATEB left, making the player think they can ignite more.

SirKeplan:
done, maybe the TEATEB stuff can be revisited in the future.

PhineasFreak:
@SirKeplan Awesome work!

PR #878, opened 2016-01-29
RO Constellation Essentials

PhineasFreak:
Add RO support for the BahamutoD Constellation Essentials pack.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_BNTR_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_BNTR_Config.cfg
@@ -1,14 +1,15 @@
 //  ==================================================
-//  Bimodal NTR engine.
+//  Bimodal NTR global engine configuration.
 
 //  Inert Mass: 2170 Kg (base configuration)
 //  Throttle Range: N/A
 //  Burn Time: 5400 s
-//  O/F Ratio: 0 (SNRE), 5.37 (TRITON)
+//  O/F Ratio: 0 (SNRE), 3 (TRITON)
 
-//  Source 1: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf
-//  Source 2: http://alternatewars.com/BBOW/Space_Engines/AIAA-2004-3863_TRITON.pdf
-//  Source 3: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20140009579.pdf
+//  Sources:
+//      http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf
+//      http://alternatewars.com/BBOW/Space_Engines/AIAA-2004-3863_TRITON.pdf
+//      http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20140009579.pdf
 
 //  Used by:
 //      BahamutoD Constellation Essentials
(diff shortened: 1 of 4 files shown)


PhineasFreak:
This was in the backlog for some time but it is finally here! And because someone asked for it in it's KSP forum thread...

winged7:
Awesome! I've found that you're working on them and I already made (a couple of months ago)  RP-0 config for all parts.

PhineasFreak:
I was very pleased to boot KSP after making the configs and finding out that they were already supported by RP-0 (how did i missed that?).

Something that is bothering me is that there are two versions of the pack, the original from Bahamuto (that these configs target) and another from KSP forum user MeCripp, who re-distributes it with changes to be compatible with KSP 1.0.5. If the second release is used then hell will break loose...

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/878/files#diff-5abe9455de2bb217d53a3f3b206841b7R7

Does this quoted fuel/oxidizer ratio (5.37 (TRITON)) translate into this KSP mixing ratio? It looks suspect to me because it is so round and simple 3:1

PhineasFreak:
I calculated the actual O/F ratio with the wrong way. The correct ratio should be 0.1571/0.8429.

stratochief66:
No worry, that is what PR review and testing is for :)

PhineasFreak:
True, a clear mind can do wonders on a wall of text that seems perfect to the one that wrote it. And i thank you for that!

stratochief66:
I see two versions here, will it matter which one the users downloads?

http://forum.kerbalspaceprogram.com/index.php?/topic/69581-rssro-constellation-essentials-v131-trusses-engines-adapters-apr21/

stratochief66:
New rule: If a mod isn't available on CKAN, please place a link to the place to download it in the PR :P

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/878/files#diff-1157b11afaabc91fb9c18113d0de507dR186

Where does the nuclear fuel come from? Perhaps you mean to change the name to EnrichedUranium?

So this engine will generate 5 EC/s at all times?

stratochief66:
From the picture on the forum page for this mod, it looks like the engine mount should have 3 nodes for holding 3 engines?
https://github.com/KSP-RO/RealismOverhaul/pull/878/files#diff-1157b11afaabc91fb9c18113d0de507dR19

PhineasFreak:
Yep, error on my end. Fixed.

Since Bahamuto has a stockalike and an RSS version, only the RSS version is supported (the stockalike is missing the Six Way Engine Mount).

Originally, there were 3 nodes, a top node for tank attachment, a middle one used for some unknown reason to me and the bottom one (for generic attachment). The engines were and are surface attachable.

stratochief66:
yeah, I think I figured out from the comments that he means "RSS compatible", not RO.

does he use different part names for one version or the other? what are consequences if someone tries to use this patch with the wrong version?

also, I just finished testing this, the only suggestion I have (not a realism one, just usabiliity) would be to add a tiny amount of 'gimballing' to the NTR engines. I would hope they would plan to do that in the real thing, or else you would have to burn a goodly amount of other fuels to correct for even slightly off centre masses during the ejection burn.

PhineasFreak:
No changes in the part names, only the missing RS-68 engine adapter for an Ares V - like launch vehicle. As i mentioned, MeCripp has repackaged the mod and even then the changes are centered around the removal of the old ModuleEngineConfigs that were included with the base parts and some changes to the resource names. That would break things so i should probably just delete the contents of the engine parts and recreate them to be on the safe side.

I though about engine gimbaling but i did not find any source that would indicate that these engines can gimbal, only that the original NERVA design had two vernier engines using waste Hydrogen from the pump.

stratochief66:
what are consequences if someone tries to use this patch with the wrong version?

PhineasFreak:
They will be stuck on the loading screen.

PhineasFreak:
And it should not be an issue anymore with the latest patch!

stratochief66:
Thanks for the fix. Compatibility is good :)

PR #881, opened 2016-01-29
Configs: Squad Engines PT3

SirKeplan:


SirKeplan:
close as will be covered in coming commits.

PR #883, opened 2016-01-30
SSTU-updates 0.3.27

shadowmage45:
* Fix Orion Service Module RealPlumes config
* Fix fairing attach setup for selectable nodes for Orion Service Module
  * WARNING - This change re-orders some modules and may break existing craft using this part
* add parachute area stats for Apollo CM parachute module for RO scaled mass (FAR and stock compatible)
* add parachute area stats for Orion CM parachute module for RO scaled mass (FAR and stock compatible)
* fix Orion Launch Abort System lower node position being too far up
* add WIP config for Series-C (Soyuz) parts.  I'm a bit lost on most of it, but they are scaled properly, attach nodes are in the right positions, mass is... close...; DM parachute is set for 6m/s touchdown at 2.9t total mass (need to account for that in the ablator mass/etc)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Apollo.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Apollo.cfg
@@ -263,6 +263,10 @@
 		@mainRetractedScale = 0.0078, 0.0078, 0.0078
 		@mainSemiDeployedScale = 0.468, 2.34, 0.468
 		@mainFullDeployedScale = 2.34, 2.34, 2.34
+		@drogueSemiDeployArea = 10
+		@drogueFullDeployArea = 103
+		@mainSemiDeployArea = 258
+		@mainFullDeployArea = 2586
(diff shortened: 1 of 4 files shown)


NathanKell:
@shadowmage45  ooh, I didn't notice you were doing these too, awesome! Always super helpful when the author her/himself steps in / steps up! :)

shadowmage45:
Aye, generally just cleaning up some of the mess that I make with my updates; mostly the bits that Jose and Stratochief haven't been able to figure out.  It certainly helps to have detailed knowledge of how all the plugins work and interact :)

stratochief66:
@shadowmage45 I plan to make and maintain a standing Github issue here for issues/bugs I've noticed. If you ever want to chime in or even fix those things, you are more than welcome.

Some of those things will of course be things you aren't familiar with (like moving an engine over to RealFuels, removing stock bits like ReactionWheels) so also feel free to ignore them :P

#888

stratochief66:
The FX work! No idea how I screwed that up so many times.

PR #892, opened 2016-01-31
Add SXT Station and Base parts configs

Zarbizaure:
I added configs for SXT's base and stations parts. For example, the IKS30 is now 6.6m width, as a Skylab OWS counterpart.
(I hope everything's fine; that's my third attempt to post it in the correct branch and I'm new on GitHub).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
@@ -661,7 +661,7 @@
     @node_stack_bottom,2 = 0.0, -2.25, 0.0, 0.0, -1.0, 0.0, 2
 
     @title = CANIOT-7 Crew Cabin
-    @description = Probodobodyne Inc's first and only foray into crew habitation modules. Bulky, but not uncomfortable. It even features a handy set of exterior handrails. 
+    @description = Probodobodyne Inc first and only foray into crew habitation modules. Bulky, but not uncomfortable. It even features a handy set of exterior handrails. 
 
     @mass = 9.5
     @maxTemp = 1073.15
@@ -691,7 +691,7 @@
 }
 
 //  ==================================================
-//  Ares LK-S3E Heavy Habitat.
+//  Caniot Heavy Orbital Habitat.
 
 //  TAC Life Support configuration.
 //  ==================================================
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
@@ -145,7 +145,7 @@
     @node_stack_bottom,2 = 0.0, -1.09, 0.0, 0.0, -1.0, 0.0, 2
 
     @title = PPD-4 Crew Cabin
-    @description = A slightly cramped ruggadised cabin. Designed for both stations and surface operations. Features usable handle bars around the side.
+    @description = Hold 2 people with one month of supplies. A slightly cramped ruggadised cabin. Designed for both stations and surface operations. Features usable handle bars around the side.
 
     @mass = 3.5
     @maxTemp = 1073.15
@@ -160,7 +160,7 @@
     {
         name = ModuleFuelTanks
         type = ServiceModule
-        volume = 20
+        volume = 50
         basemass = -1
 
         TANK
@@ -189,21 +189,21 @@
 		TANK
 		{
 			name = Food
-			amount = 11.7
+			amount = 175
 			maxAmount = 175
 		}
 
 		TANK
 		{
 			name = Water
-			amount = 7.75
+			amount = 116
 			maxAmount = 116
 		}
 
 		TANK
 		{
 			name = Oxygen
-			amount = 1183.7
+			amount = 17755
 			maxAmount = 17755
 		}
 
@@ -231,7 +231,7 @@
 		TANK
 		{
 			name = LithiumHydroxide
-			amount = 1.5
+			amount = 22.5
 			maxAmount = 22.5
 		}
 	}
@@ -274,7 +274,7 @@
     @node_stack_bottom,2 = 0.0, -1.63, 0.0, 0.0, -1.0, 0.0, 2
 
     @title = PPD-6 Crew Cabin
-    @description = A cruder, but more sturdy forebearer to the the PPD-10. The design specified a tougher skeleton and skin surface, due after a series of 'dicey' touchdowns.
+    @description = Hold 4 people with one month of supplies. A cruder, but more sturdy forebearer to the the PPD-10. The design specified a tougher skeleton and skin surface, due after a series of 'dicey' touchdowns.
 
     @mass = 6.8
     @maxTemp = 1073.15
@@ -289,7 +289,7 @@
     {
         name = ModuleFuelTanks
         type = ServiceModule
-        volume = 40
+        volume = 100
         basemass = -1
 
         TANK
@@ -318,22 +318,22 @@
 		TANK
 		{
 			name = Food
-			amount = 11.7
+			amount = 525
 			maxAmount = 525
 		}
 
 		TANK
 		{
 			name = Water
-			amount = 7.75
+			amount = 348
 			maxAmount = 348
 		}
 
 		TANK
 		{
 			name = Oxygen
-			amount = 1183.7
-			maxAmount = 53265
+			amount = 74000
+			maxAmount = 74000
 		}
 
 		TANK
@@ -360,7 +360,7 @@
 		TANK
 		{
 			name = LithiumHydroxide
-			amount = 1.5
+			amount = 67.5
 			maxAmount = 67.5
 		}
 	}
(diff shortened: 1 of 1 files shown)


NathanKell:
@stratochief66  looks good to me! Have you had a chance to peek? Or anyone else familiar with this stuff, @SirKeplan or @A1Ch1 or @jbengtson maybe?

Zarbizaure:
I would also like to add RP0 configs (because they are in a wrong node) but I prefer to wait for the station line in the tech tree...

NathanKell:
Awesome!

stratochief66:
If something is big enough to hold people, I would expect it to be able to store more than 40L of stuff, no?
https://github.com/Zarbizaure/RealismOverhaul/commit/e1c4846d343de4694309236ee8c7fcb9ba40ad57#diff-91b89e94a038c121d938ca99b5164aecR292

Is there any reason this part holds (from the name, what I assume to be) 6 people, but contain no life support supplies like the previous part does? Similar thing with this part,

https://github.com/Zarbizaure/RealismOverhaul/commit/e1c4846d343de4694309236ee8c7fcb9ba40ad57#diff-91b89e94a038c121d938ca99b5164aecR292

It might be good to state in the description how many people it is expected to hold, and possibly roughly how many days of life support the storage should cover?

Any reason these aren't full to start?
https://github.com/Zarbizaure/RealismOverhaul/commit/e1c4846d343de4694309236ee8c7fcb9ba40ad57#diff-91b89e94a038c121d938ca99b5164aecR363

Beyond these things, this definitely looks decent. Apologies @Zarbizaure that I hadn't seen these.

Zarbizaure:
You're wright, i'll work on this today. I was more concerned about the size and scale of these various parts but since it's okay I will tweak their capacities and descriptions. I just wondered why there is two different container settings: One for ModularFuelTanks and one for TACLS. Would it be better to set them both at the same volume? Or should I tweak the Modular one so that it allow a reasonnable amount of fuel to be stored, depending on the station's size? (and keep another setting for TACLS in order to have some days/months of supplies)

stratochief66:
I'm not sure what you mean about two seperate tanks. Here is an example of the one Skylab I'm aware is well configured for RO:

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg#L73-L142

It has one combined tank. The FASA service modules are similar too, with just one tank.

Zarbizaure:
Ah ok, I just took the same configs that were included with the Lacuga Pod. I will merge the two tanks as soon as possible (and also update the Lacuga).

PhineasFreak:
@Zarbizaure Since i wrote the configs for the SXT "Lacuga", i have to comment on the tank system: If you take a close look, TACLS does not create more tanks, it edits the currently existing one so that they have all the required life support resources. This is done on the separate patch after RO - if TACLS is not installed then none of these resources will be added.

https://github.com/KSP-RO/RealismOverhaul/pull/892/files#diff-91b89e94a038c121d938ca99b5164aecR185

stratochief66:
@PhineasFreak thanks for clarifying.

I don't believe that is the way we do things for reproduction craft, because that would mean the mass is dependent on whether the user has TACLS installed or not. For example, if someone builds the FASA Apollo, all the life support supplies will still be there (giving the same craft mass) but they won't be consumed or transformed.

So, if part of your goal is to make the craft challenging to put in orbit and move around to being realistically massed with life support systems and supplies, then that is the way to go, even if you aren't making a historical reproduction craft.

This has the benefit of having more a more realistic challenge in the masses, just without the hassle of actually managing the life support systems and supplies like someone with TACLS installed will have to do.

Zarbizaure:
Well...so I don't really know how to change it now. At least it works, but as you said, it depends if the users has installed TACLS or not. If I change it to match the Skylab file you gave me, how will it react on a non-TACLS install? (since food, water etc are present in the Modular Fuel Tank withouth any check).

PhineasFreak:
@stratochief66 the TACLS supplies do not weight much (compared to the rest of a crewed craft). For example, the Orion MPCV supplies are meant to last for 20 days (4 crew) and the mass does not exceed 40 kg (including waste products). They will be lost "in the noise" of all the other craft parts.

stratochief66:
@PhineasFreak I will have to look at the numbers, since that doesn't seem right. Food alone should be more than that, plus water and Lith Hydroxide, etc. Also, keep in mind some supplies are stored in the service module (ie. water) and transferred to the capsule via umbilical.

Of course, I'm not aware of a very well configured Orion (SSTU is a WIP of course) in RO

Zarbizaure:
So, I will keep it as @PhineasFreak did, but if anyone would like to change it to an unified config, feel free to do that.
Also, it would be cool to add electricity comsuption rate to these station parts.

PhineasFreak:
@stratochief66 hmm, you are right, the value that i reported is for a lesser number of days. Anyway, after some rethough, it feel it would be better to include the TACLS resources with the main RO config and make an AFTER patch to include the required TACLS converters (so that the mass is maintained and KSP does not complain about missing modules).

@Zarbizaure you can do this with a bit hacky way:

```
MODULE
{
    name = ModuleGenerator
    isAlwaysActive = true

    OUTPUT_RESOURCE
    {
        name = ElectricCharge
        rate = -0.2
    }
}

All you have to do then is to change the rate (here it draws 200 Watts).

stratochief66:
The FASA Apollo is my personal standard. It looks like just including the TACLS modules doesn't cause any trouble when TACLS isn't present. @NathanKell can probably shed some light if I am incorrect in that. https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg#L312-L319

@NathanKell, is what FASA-Apollo does correct, or should the TACLS converters be handled as @PhineasFreak states?

(sorry @Zarbizaure we are having the discussion here on your PR, but you bring up a good and important issue we want to address now :)   )

NathanKell:
A MODULE that points to a class that is not loaded is silently ignored, so it's fine to do that. Also, since we depend on RF and RF depends on CRP, we have the resources TACLS uses. And we don't want masses to change based on if you have TACLS or not. So all resources should be added in all cases, we don't need to NEEDS anything on TACLS.

Zarbizaure:
Okay, so the config given by @stratochief66 is the one to choose. I will push a new commit once I will have updated to that template (may I also change the Lacuga @PhineasFreak ?).

Zarbizaure:
what are the correct rate for supplies per person? I'm currently re-configuring each of the life support module, but I'm not sure from which config to start.

stratochief66:
@Zarbizaure
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/ThunderAerospace/TacLifeSupport/PluginData/TacLifeSupport/LifeSupport.cfg

About the 200W per person; I'm honestly not sure. That varies substantially per actual RL craft. For example, Gemini was much more per person than Apollo. Depending on what you are modelling your part(s) after, it might be good to state it (ie. this is like a Skylab) and try to find historical sources for that hardware to pull/extrapolate from. Or if you are unable to find such sources, base it on the existing skylab config:
https://github.com/KSP-RO/RealismOverhaul/blob/68333af35b8de3c1c922ab0dcd8cfa4993206efe/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg#L73-L142

note, this skylab uses a heavy, advanced co2 -> O2 converter, where-as Apollo & Gemini use Lithium Hydroxide canisters meant for a shorter mission.

NathanKell:
I'm going to merge this in for now, and fixes can then be made to the existing file. One thing I would like--and I'll do a quick pass right now--is for the PPD-4 and 8 modules to be scaled at 3m in diameter. They can help replicate early station designs (Mercury station plans, or early MOL plans, or even the original space-station-first approach station designs pre-lunar-goal).


Regarding the CANIOT, I presume 5m is broadly correct for Salyut 7?

NathanKell:
Ok, looking at the file to edit and looking at the original cfgs, and... @Zarbizaure did you test all these? Because it looks like the nodes are very wrong, and only some of the MODEL nodes in the part cfgs are touched... :(

Zarbizaure:
All my nodes were correct after edit. Which model seems wrong for you?
As you said, I may downscale the PD4 to 3m and then adjust the nodes, but with its currents size nodes are ok.

NathanKell:
I checked. The ladder bits on the PPDs were not rescaled, and the nodes on many of the other parts are inside the edges of the parts. Is there any particular reason the nodes you specify don't match the scaling factor (i.e. your stated scale divided by the original MODEL scale) times the original attach node position?
Also, you were using @node_stack_bottom,1 and @node_stack_bottom, 2  when there's only one node_stack_bottom in the original cfgs (and that would be index ,0 anyway, not ,1) and the node_attach was not even being patched.

Apologies for not catching all this earlier, btw, I was looking at the stated sizes and masses and capacities rather than the nuts and bolts of the scaling.

NathanKell:
Oh, were you maybe using a very out of date version of SXT? It's possible things changed...

Zarbizaure:
Mhhh...I'm up to date with SXT, I don't have time now but I will work on that tonight. The nodes may not be correct (I copied-pasted the Lacuga and edited node 1 and 2) but it worked fine for me; and I did not change anything in the origna config.

PR #895, opened 2016-01-31
RO RSB Saturn

PhineasFreak:
* Add RO support for the Real Scale Boosters Saturn IB and V parts.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Saturn.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Saturn.cfg
@@ -1,14 +1,14 @@
 //  ==================================================
 //  Sources:
 
-//  1: http://history.nasa.gov/SP-4029/Apollo_18-12_Launch_Vehicle-Spacecraft_Key_Facts.htm
-//  2: http://history.nasa.gov/SP-4029/Apollo_18-19_Ground_Ignition_Weights.htm
-//  3: http://history.nasa.gov/ap12fj/pdf/a12_sa507-flightmanual.pdf
-//  4: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19740021163.pdf
-//  5: http://www.spaceaholic.com/index.php/Detail/Object/Show/object_id/21
-//  6: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20100027314.pdf
-//  7: http://alternatewars.com/BBOW/Space/Reference_Spacecraft_Engines.htm
-//  8: http://history.msfc.nasa.gov/saturn_apollo/documents/Instrument_Unit.pdf
+//  http://history.nasa.gov/SP-4029/Apollo_18-12_Launch_Vehicle-Spacecraft_Key_Facts.htm
+//  http://history.nasa.gov/SP-4029/Apollo_18-19_Ground_Ignition_Weights.htm
+//  http://history.nasa.gov/ap12fj/pdf/a12_sa507-flightmanual.pdf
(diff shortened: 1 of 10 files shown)


stratochief66:
Too drunk to unit but I believe this is wrong?
https://github.com/KSP-RO/RealismOverhaul/pull/895/files#diff-d4693868e0e9fac9e5275529afe231d7R234

IIRC, 1EC/s = 1kW

stratochief66:
This thrust might be a bit too high. I find it interesting that it appears to be double the thrust of the FASA APS. IIRC, we have a good sourcing for the FASA APS. @NathanKell?
Two of the FASA APS are sufficient to ullage a lunar-ready Saturn V payload, so this thrust is perhaps overkill anyway.

https://github.com/KSP-RO/RealismOverhaul/pull/895/files#diff-d4693868e0e9fac9e5275529afe231d7R234

https://github.com/KSP-RO/RealismOverhaul/pull/895/files#diff-d4693868e0e9fac9e5275529afe231d7R234

I'm not as confident about the FASA APS ISP values, but it might be worth sourcing those values.

stratochief66:
Any this is way cleaner than my usual way to remove stock resources from parts. I will try to reuse this in the future :) 

https://github.com/KSP-RO/RealismOverhaul/pull/895/files#diff-d4693868e0e9fac9e5275529afe231d7R234

PhineasFreak:
The problem with the ElectricCharge is that while it is true that 1EC = 1 kW, then you have an avionics unit that uses 0.2 EC/s. Think how fast electricity will disappear if i only specify 0.58 units...

The same is true with the batteries on the FASA Agena avionics unit: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Agena.cfg#L92. Agena has something like 70.2 kWh of available battery power but this would mean to have 70.2 units of EC, along with a probe core that uses 0.25 EC/s --> ~280 seconds of life time. Watt-hours makes much more sense.

stratochief66:
Watt-hours makes more sense in what way?

If we have a source for how much battery power a piece of hardware has, that is what we should be putting on the parts in RO.

The ideal solution would be more advanced resources flow methods, which might not even exist, I really suck at those (paging @NathanKell ) My operational solution has been to lock the resource that I don't want to flow, until I want to use it. For example, capsule pyrobatteries (for re-entry chute firing) would easily be drained if the craft was drawing off of the, but I don't make them huge to counter that, I just lock the charge in them until the appropriate point in the mission.

1 EC/s = 1 kW         1 EC = 1 kW*s

548 Wh = 548 W * (3600s) = 1972800 W*s = 1972.8 kW*s = 1972.8 EC  (?)

If that calculation is correct, then the part with 548 Wh should contain ~1973 EC

PhineasFreak:
All sources used for the parts are available at the top of each configuration file.

SE-7 RCS thruster data used on the Saturn V APS units: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20100027314.pdf (pages 137 & 138).

APS unit data: http://history.nasa.gov/ap12fj/pdf/a12_sa507-flightmanual.pdf (pages 134 & 135).

I have also left a comment for the reason of this high thrust values: https://github.com/KSP-RO/RealismOverhaul/pull/895/files#diff-d4693868e0e9fac9e5275529afe231d7R264. Since it is impossible to separate the thrust values for each axis anymore, i decided to use the higher thrust (doubled) by nessecity. Ullage is just even more easy to manage now.

PhineasFreak:
And i am again caugh for not understanding basic math...you are 100% correct.

Time to fix a load of these errors!

PhineasFreak:
Consumption values should also be more in-line with FASA.

stratochief66:
Let us know when you're done adding to this PR, it tests out good so far.

PhineasFreak:
I can safely say that it is done (for now, something always pops up later to update!)

PR #897, opened 2016-01-31
RO RSB Atlas V

PhineasFreak:
* Add RO support for the Real Scale Boosters Atlas V (only for the 401 version for now).


PhineasFreak:
Please hold this a bit before merging so that i can include all the Atlas V parts.

PhineasFreak:
And with that last patch all Atlas V parts are RO - compatible.

Note: there are some separation issues regarding the 5.4m fairings, even with the high ejection force.

stratochief66:
Let me know when you are done adding to this so I can test it

stratochief66:
Also, which other unmerged PRs this relies on

PhineasFreak:
For the Atlas specifically, the only thing that is suggested (not a hard dependency) is the AJ-60A inert mass fix by https://github.com/KSP-RO/RealismOverhaul/pull/915, otherwise there are none.

PR #898, opened 2016-02-01
RN Updates for exp mods

raidernick:
- Add configs for tanks and engines for r7, preliminary unfinished

Do not merge this until i give the ok!


raidernick:
@NathanKell this is good to merge

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/898/files#diff-98797fa417645e6128654a28b0636e31L84

You delete the code for this part, but I don't see it get added anywhere else, where did it go?

Also, why did you rename all the parts base names to start with rn_, that is very craft breaking to anybody who had these craft in flight.

raidernick:
i removed it because I combined all the parts into 1 part and use firepsitter texture and mesh switcher now and i renamed all the parts because i was combining them anyway, so it's going to break everything anyway. It's something i've been meaning to do for a while; for instance, I used to have 4 different core stages, they were identical in stock except the textures, I combined them all into 1 model now that you switch the texture onso anyone who had those will have broken saves anyway, irrelevant to RO.

I also renamed stuff I didn't change to keep with the new convention of naming. SOme of my parts had simple names like "luna" or "vernier", this is unacceptable and can cause conflict with other mods.

PR #900, opened 2016-02-01
Updated RD-191 impulse, thrust values

Ravenchant:


SirKeplan:
huh, strange. that weight is not dubiously low. it looks like the mass as it is is too high, so you could change it to the value of 2.2

Also, what i have been doing elsewhere is putting any references at the top of the file, so we can see them in one place.

Ravenchant:
Oh, I meant low in comparison to the current value, but it seems to be correct. Dry masses given by all the sources I've looked at are: NPO Energomash- 2200 kg, Spaceflight101- 2290 kg, Russianspaceweb- 2200 kg, Astronautix- 3230 kg. Clearly Astronautix is the odd one out.

And I'll write the sources at the start from now on, thanks.

SirKeplan:
ok great, yeah, astronautix can be very unreliable at times, for russian/soviet stuff atleast.

PR #903, opened 2016-02-03
Add M55 Minuteman config

A1Ch1:
Clone of part that was switched to engineType so there was no longer ModuleEngineConfigs to modify. Fixed by deleting engineType, adding ModuleEngineConfigs.


stratochief66:
Fix how? please provide some PR description. What was wrong/broken, etc.

A1Ch1:
Clone of part that was switched to engineType so there was no longer ModuleEngineConfigs to modify. Fixed by deleting engineType, adding ModuleEngineConfigs.

A1Ch1:
Better fix - add M55 global config. Tested.

PR #904, opened 2016-02-03
Global configs patch1 (SSTU stuff)

SirKeplan:
Engines done:
F-1
F-1B
J-2
RL10
SSME
AJ10-137
M-1
Vinci
(effectively the engines for SSTU, but touches a lot of other mods)

**Possible Issues:**
SSTU RS-25 uses ModuleEngineIgnitor, which seems quite unusual, should this be removed?
A few gimbal values where changed because of inconsistencies like with the F-1, but i chose a value with references.
I am putting ModuleEngines* in the Global Engine files for ignitions etc, this seems simpler especially when otherwise it would involve a lot of copy paste bloat.

It's a large PR, atleast github shows it all.


SirKeplan:
There was a bug which made the AJ10-137 gimbal not work, this PR fixes it.

PhineasFreak:
The ModuleEngineIgnitor is a (deprecated i believe) ullage & ignition simulation module from the EngineIgnitor mod. This mod has not been updated for a long time.

I suggest removal of the module and replacing it with the RF provided parameters.

PhineasFreak:
Also, some engines might have different ullage or ignition requirements. An example of this is the LR-87 that RO provides: early versions could not be restarted but newer versions could.

NathanKell:
Wait, what? When was the LR87 reignitable? That said, since ignitions is per-CONFIG that really isn't a problem, right?

I concur re: ModuleEngineIgnitor.

A1Ch1:
Back when first introduced it had multiple (infinite, I think) ignitions. That was later fixed on at least the AIES version by the time the gas generator exhaust was added, don't know about the versions found elsewhere.

PhineasFreak:
@NathanKell as per the original parameters, it was: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Parts/Engines/LR87LH2.cfg#L179.

Also, the ullage and ignitions will only be set once for all engines so i do not find a reason why each engine config should not have it's respective parameters.

SirKeplan:
@PhineasFreak ah, just noticed a bug related to what you said, i was trying to save effort and file bloat, but forgot a precise working of ModuleEngineConfigs, i will have to put ignitions per engine config instead of at the top in ModuleEngines* atleast for some engines.

also, will remove the ModuleEngineIgnitor code in a bit.

NathanKell:
@PhineasFreak Ah right, that's my speculative upgrade to a speculative vacuum-optimized LR87-LH2 config. I thought you meant one of the actual real, let alone flown, LR87 marks. :)

@SirKeplan: I had thought that per our discussion on IRC you were always putting ignites, IGNITOR_RESOURCE, ullage, and pfed in the CONFIGs for uniformity.

PhineasFreak:
I have not done any research on the LR-87 so i assumed that this was a real variant, my fault.

NathanKell:
NP. In fact I thought I left a comment in the file about that, but apparently I didn't. My mistake. :]

SirKeplan:
@NathanKell you are right, it makes sense to do it that way, i was..... experimenting.

PR #905, opened 2016-02-04
SRMs: KW (3)

A1Ch1:
Second half of KW changes to engineType configs.


A1Ch1:
Third of three sets now.

stratochief66:
I don't think any of your KW PRs caused this, but there appear to be 2 Gem 46 and 2 Gem 60 engines, for each set there is one with a nosecone and one without, which could presumably be used as a first stage since it has an attach node on the top.

Anyway, for both of these sets, one uses a thrust curve with a label including "Fixed". Could you please ensure both parts in each set are using the template?

stratochief66:
Beyond that, the rest of the PR checks out well.

A1Ch1:
I believe this is also being caused by the configs for one version of the GEM 46 being in this PR and its cousin being in another. I'll double check.

A1Ch1:
No problems when the Castor 4A config is inclued

stratochief66:
Gotcha. So, when both are included together, it works out. Thanks :)

PR #906, opened 2016-02-04
SRMs: KW (2)

A1Ch1:
Second of three sets of changes


stratochief66:
I am getting odd behaviour from the %engineType = GEM-63

It appears to have the Castor-4A thrust specs. I confirmed I have the GEM-63 template locally.

I downloaded the most recent Github version of RO, then added the one file from this PR that is changed.

A1Ch1:
The GEM 63 behavior should be because the 63 is a clone of the Castor 4 but the Castor 4 config had to be split into a different PR. I'll double check:)

A1Ch1:
Yup - works when the changes are merged together

PR #911, opened 2016-02-04
RO RSB generic parts

PhineasFreak:
* Add RO support for the Real Scale Boosters AJ-60A and GEM-60 radial decoupler.
* Add RO support for the RSB strut connector.


PhineasFreak:
The radial decoupler is common between the Delta IV and the Atlas V so i decided to push it along with the final Atlas V configs.

PR #915, opened 2016-02-05
RSB Engines

PhineasFreak:
These were already stashed before merging but i forgot to push it.

* Decrease the gross mass of the AJ-60A SRM to include the mass of the
SRM decoupling kit.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Engines.cfg
@@ -65,7 +65,7 @@
     {
         name = ModuleFuelTanks
         type = HTPB
-        volume = 52569
+        volume = 16780
         basemass = -1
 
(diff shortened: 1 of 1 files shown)


PhineasFreak:
Renamed due to the addition of the F-1B and the RL10B-2 engines.

PhineasFreak:
And again because of the RS-68 and the GEM-60.

stratochief66:
https://github.com/KSP-RO/RealismOverhaul/pull/915/files#diff-6bb525f6473232cc4d068d711eccdc83R68

This gives them ~90 tons of fuel a piece, which is roughly 3x what they should have. Where did those numbers come from?

PhineasFreak:
I multiplied the propellant mass instead of dividing it (by 1.77). Fixed.

stratochief66:
i recall an engine throttling far more than I expected that it should. I don't recall offhand which one it was though. Was it possibly because of this?: 
https://github.com/KSP-RO/RealismOverhaul/pull/915/files#diff-6bb525f6473232cc4d068d711eccdc83R899

PhineasFreak:
These are just templates, they do not do something (dummy parameters).

PR #917, opened 2016-02-06
RN Luna 9 fix (remove RW), CSS scaling

raidernick:
-Remove reaction wheel as rcs on the model file has been fixed.


raidernick:
to clarify the spacecraft IRL didn't have reaction wheels, that was a cheat/hack for RO to make up for the lack of proper rcs, which has now been fixed.

PR #918, opened 2016-02-06
RSB Ares I bugfixes

PhineasFreak:
* Fix the commented battery amount of the Ares I upper stage.
* Make the avionics electricity consumption be the same with or without Remote
Tech installed.
* Increase a bit the electricity consumption of the upper stage antenna.
* Fix a wrong required resource by the Ares I 5 meter interstage adapter
RCS thrusters.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ares_I.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ares_I.cfg
@@ -1,13 +1,13 @@
 //  ==================================================
 //  Sources:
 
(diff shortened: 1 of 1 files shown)


stratochief66:
Your previous notation was for 5145 W*h, which would be 18522 units of EC (kW * s) Here you have a new notation that matches the amount of EC, but nothing pointing to how you arrived at that amount of EC in the first place.

Could you point me to the source for the EC?

PhineasFreak:
No data were available for the batteries of the Ares I (as the only thing that was ever constructed for it was a dummy upper stage). The value here is calculated for the upper stage to have enough time to push the Orion stack into orbit and deorbit.

stratochief66:
Gotcha. Could you put that in the file, instead of the amount of kWh, since
listing it that way leads me to believe it is a cited value. Thanks :)

PR #921, opened 2016-02-07
RO RSB Delta IV

PhineasFreak:
* Add RO support for the Real Scale Boosters Delta IV launch vehicle family.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
@@ -1,0 +1,134 @@
+//  ==================================================
+//  Sources:
+
+//  http://www.ulalaunch.com/products_deltaiv.aspx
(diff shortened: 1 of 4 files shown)


stratochief66:
These PRs are really becoming a mess @PhineasFreak. Can you post #xxx links in each PR that relies on the content of other PRs?

PhineasFreak:
Only the engine configs from https://github.com/KSP-RO/RealismOverhaul/pull/915.

stratochief66:
Look alright. You seem to have used a 'kerolox upper' plume on the RS-68, which is a 'hydrolox lower' engine. Could you change that over?

PhineasFreak:
The RS-68 may be a hydrolox engine but the FX is correct because it is an cooled by ablation: https://www.youtube.com/watch?v=ermr2spJYKo

stratochief66:
yeah, good call on the FXs. I watched a few videos of the RS-68, and it definitely glows yellow like a kerolox, but I'm fairly sure the core engine wasn't producing copious amounts of smoke.

PR #925, opened 2016-02-09
RSRM global engine configuration

PhineasFreak:
* Add a global engine configuration for the Space Shuttle 4 segment RSRM.


A1Ch1:
Technically the 4-segment is just the RSRM, not the RSRMIV. The 5-segment is referred to as the RSRMV but that's to distinguish it from the Five-Segment Booster (FSB) that was tested in the early 2000s.

A1Ch1:
Additionally it was only manufactured up to ATK: while Orbital ARK offers a four-segment booster, it's different from the RSRM that was flown on the shuttle.

PhineasFreak:
@A1Ch1 i have seen so many names for these that i cannot differentiate one from another. Shuttle supposedly had two versions (pre and post Challenger), later a filament wound (only tested once or twice) and after that, the Constellation 5 / 5.5 segment ones. And today, Orbital has a whole line named "RSRM", ranging from 1 to 5 segments.

Should i rename it to just RSRM? I remember that a pack added all of the Orbital's new RSRMs and that is why i wanted to differentiate one from another (RSRMI, RSRMII etc).

A1Ch1:
@PhineasFreak The SRM refers to the motor itself, while the SRB refers to the entire booster with all its associated hardware. There was the original SRM, followed by the High-Pressure Motor (HPM) up to Challenger, then the Redesigned Solid Rocket Motor (RSRM). There was also the FWB, ASRM, and FSB, none of which flew. Here we're modeling the RSRM, so I'd use that name. For the family in the Orbital ATK catalog, naming them the same as in the catalog seems like the natural way to distinguish them, I believe it's something like "RSRM (X segment)".

PhineasFreak:
OK, now i can see the whole picture (and make the necessary changes).

PR #926, opened 2016-02-10
RN multifixes

raidernick:
-Fix vosktok again
-Fix skylab trs
-Add proper configs for trs and docking port

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CSS/CSS_Mike_NZ.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CSS/CSS_Mike_NZ.cfg
@@ -676,6 +676,7 @@
 		MidChordSweep = 0
 		b_2 = 9
 		TaperRatio = 1
+		transformName = Elevon
 	}
 }
 @PART[ETDoors]:FOR[RealismOverhaul]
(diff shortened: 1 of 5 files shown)


ctiberious:
I'm curious about the stats you're using for the Skylab TRS.  When I was looking into the part, I wasn't able to find a whole lot of information about it, but what I did find was some weight and thrust data.
From the information I was able to fine, the total weight of the TRS was around 9684lbs (4401.82kg), of which 6177lbs (2807.73kg) was the Hydrazine fuel.  That puts the dry weight of the TRS at 1.594t (though it looks like that should include the docking probe) but the mass you're using is 2.884.
Also, based on the weight of Hydrazine as listed in CommonResources.cfg (0.001004t/L), my math comes up with 2796.54L of Hydrazine but the value you're using is 2790.68 (2801.84kg).  That's not a major difference but still curious about where you came up with it.
It's perfectly possibly that you managed to find better source material than I did but since there is no mention of it in the file changes, I was just curious where your data was coming from.
The information I found came from the following sources:
http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19780011828.pdf
https://en.wikipedia.org/wiki/File:Teleoperator_Retrieval_System.jpg

raidernick:
9684-2209-1298=6177lbs, 6177 lbs in megagrams = 2.80184, 2.80184/0.001004= 2790.67729084L

A1Ch1:
The input values have only 4 significant digits so the output only has 4 sig figs - 2791 L

PR #928, opened 2016-02-10
Skylab TRS

ctiberious:
I noticed that the Skylab TRS says it's setup for RO/RP-0 (doesn't have a non-RO or non-RPO notice) but its still setup to use "LiquidFuel" and "Oxidizer" and the mass is totally off.  These changes should set it to realistic values.  Unfortunately I couldn't find a whole lot of documentation about the TRS but I did find a graphic that included some weight values (https://en.wikipedia.org/wiki/File:Teleoperator_Retrieval_System.jpg) and a NASA write up with some further information (http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19780011828.pdf).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -237,7 +237,113 @@
 }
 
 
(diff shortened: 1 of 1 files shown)


stratochief66:
raidernick appears to currently be working on that. I recommend you take a look at his PR and see what he is working on: 
https://github.com/KSP-RO/RealismOverhaul/pull/926

raidernick:
i already did this

PR #929, opened 2016-02-11
RO RSB STS/DIRECT

PhineasFreak:
* Add RO support for the Real Scale Boosters STS/DIRECT parts.


PhineasFreak:
This requires the RSB engine configs from https://github.com/KSP-RO/RealismOverhaul/pull/915.

PR #930, opened 2016-02-11
RO CST-100 pack updates

PhineasFreak:
* Balance the CM electricity consumption to be the same with or without RemoteTech installed.
* Update the LAE engine block to the new engine system (requires the update engine config from https://github.com/KSP-RO/RealismOverhaul/pull/909).
* Fix the SM inert and propellant masses due to the engine mass changes.
* Fix an issue regarding the docking node patching of IDA.
* Add a Real Plume config for the LAE engine block.


SirKeplan:
You have the title/manufacturer/descriptions fields in both the engine part and the engine config in #909 surely the fields in the engine config take precedence and make these useless.

PhineasFreak:
I forgot to add it, thanks for reminding me!

PR #933, opened 2016-02-13
RN multifixes #2

raidernick:
-Add configs for mariner1/2
-Add ec to juno ii ss
-css fix axis limits for control surfaces
-add agena b fasa configs


raidernick:
I removed the DECQ configs because of several reasons:

1) He releases new updates on a regular basis, these updates completely break the configs as names, folder structure, resource names keep changing
2) Some of his parts have bad collider issues that change randomly on each update
3) Some parts are not centered properly at the origin and this changes with each release, affecting all nodes
4) There are crossfeed issues with the boosters that can't be fixed using that stack search hack because the tanks are in multiple pieces
5) The control surfaces null ref using far modules
6) the tail null refs on it's own and the model file needs to be fixed
7) He releases his own unfinished/broken ro/rss configs that conflict with these here

stratochief66:
1. please don't throw so many unrelated changes together in one PR, or we have no choice but to reject them as a unit

2. the fasa agena mass change makes sense. please cite your source(s) at the beginning of the part file, or in the PR notes.

3. Lines like those have moved these parts to use the engine templates, so don't reverse them without an explanation that includes a good reason.
https://github.com/KSP-RO/RealismOverhaul/pull/933/files#diff-4272427f87e810f628186d14c801ef8eL255
https://github.com/KSP-RO/RealismOverhaul/pull/933/files#diff-4272427f87e810f628186d14c801ef8eL313

stratochief66:
And the DECQ notes all seem to be good justifications. I don't know if whole-hog deleting the code/files is the best route though. @NathanKell ?

NathanKell:
It will be saved in git history, so deleting is fine. I concur on that, though I concur with @stratochief66 on the other bits--we're not touching your own patches, but please don't revert the engine work elsewhere :)

raidernick:
@stratochief66 I didn't change the engines configs, the fasa configs I have don't have the new changes in them I didn't know it was changed, it must have happened recently and my local copy is not up to date.

stratochief66:
@raidernick I read through the scrollback about this on the IRC. I have also accidentally created PRs that cause a mess with previous changes, not knowing it, it happens. in reviewing a PR, all I can do is state what it looks like is happening, since it isn't reasonable for the PR creator to explain every single line of code to avoid confusion.

what I always do, first or second thing in looking at a PR is to look at the "Files Changed" section, since that shows all the actual code changes it contains. I would definitely recommend doing that in the future yourself as well, since if your PR shows a change you didn't intent there, you will be the first to know and can close the PR and figure out what went wrong. in a case like that, you can try to figure out what happened and ask us for advice, we are always interested in helping a PR make it through review :)

sorry if point 3 came across as an accusation that you did something you didn't intend. I can only describe what something appears like in the code so that we can sort out what was intended (either with an explanation, or a reversal of the code change)

PR #934, opened 2016-02-13
RO RD-107/108 global engine config update

PhineasFreak:
* Add global title, manufaturer and description fields (reuses text by the STX RD-107/108).
* Add gimbal capability (the baseline RO patch must support this).
* Update ullage simulation and ignition requirements (apples to each engine configuration separately).


SirKeplan:
Looks good.
Few minor things i noticed.

At the top of each you say ~5 mix ratio, did you mean ~2.5 maybe?
oing `!MODULE[ModuleEngineConfigs] {}` is probably bad practice, better to make sure any `engineType` setting is qualified with a `@` or `%` i think.
I'm also not sure about the setting of HTP as an ignitor resource, though i suppose it makes sense for it's actual use.

PhineasFreak:
Yup, another typo!

The deletion of the node was previously there and i wasn't sure if it was a good or a bad idea. Now that i have some feedback on that, i can see that it might cause some problems.

HTP is the equivalent of TEATEB on these engines (it is used to spin up the turbopumps and maintain them for the duration of the flight).

On another note, should we separate the two engine configs into RO_RD107 and RO_RD108? I understand that they have a lot of commonality but they have completely different applications.

SirKeplan:
@PhineasFreak yeah, i think separating them makes sense.

NathanKell:
@PhineasFreak HTP should be used as a PROPELLANT only, like kero and LOX, not as an ignitor resource. For now we are not using any ignitor resources except EC or TEATEB (or the like), i.e. we are not using any resources for ignitor resource that the engine uses normally. The wastage of the resource (what their being added as ignitor resources simulates) should and will be handled by RF.

As to splitting them up: Yes please do. The SXT-clone RD-107 and 108 are separate parts, and that should be the case for the global configs. Yes, they do share a bunch of bits here and there, but then so did LR79 / LR89 / LR105.

PhineasFreak:
That feature of RF was unknown to me, thank you for clarifying it.

SirKeplan:
well, this looks ready to merge IMO.

PR #937, opened 2016-02-13
Global NK-9 and NK-9V config

SirKeplan:
give NathanKells NK9s a global config files

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_NK9.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_NK9.cfg
@@ -1,0 +1,70 @@
+//  ==================================================
+//  NK-9
+
+//  Throttle Range: N/A
+//  Burn Time: N/A
+//  O/F Ratio: 2.5
+
+//  Sources:
+//	http://www.astronautix.com/engines/nk9.htm
+//	http://www.b14643.de/Spacerockets_1/East_Europe_2/N-1/NK/index.htm
+
+//  Used by:
+//      Bobcat
+//  ==================================================
+//
+@PART[*]:HAS[#engineType[NK9]]:FOR[RealismOverhaulEngines]
+{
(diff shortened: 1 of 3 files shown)


NathanKell:
<3

NathanKell:
Ah, this shows I had a missed line in the original config. NK-31 can relight, AFAIK. That's presumably why it's heavier. Don't know about ignition count though, something like 2-5?

SirKeplan:
Ah, ignitions. so the NK-31 is basically an upgraded NK-21 for the N-1F. now the 1st and second stage engines where upgraded to support multiple firings for testing purposes for the N-1F, but i'm not so sure about the 3rd and 4th stages.

and i don't know whether the the NK-21 ever supported multiple firings, i assume it would, but other contemporary engines didn't, and 1 ignition is all you technically need if the 3rd stage puts you in a good LEO.

NathanKell:
Ya NK-39 was the third stage version.  NK-31 was the fourth (TLI) stage
engine.  I did read it was explicitly built for multiple ignitions.

Sent by my thumbs, slowly.

SirKeplan:
Added an NK-39.

Another thing is the NK-19/39 would have had throttling capability, as the 3rd stage used differential throttling instead of gimbaling, that might mean the NK-21/31 could throttle too.

stratochief66:
PR looks good to me. Since @NathanKell already commented here, I assume this is good to merge then? (As good as it can when I'm too busy/lazy to test in game)

SirKeplan:
Yeah, i don't think there can be any issues as it is, we can make mods to these engines later if we have more info.

NathanKell:
Heh. Was just about to respond. :)
AFAIK the 19 and 21, vs the 31 and 39, are reversed: the NK-21 was the single-engine variant, but the NK-39 was the single engine variant (and restartable, where the 31 was not).

PR #938, opened 2016-02-13
Squad NSTAR update

PhineasFreak:
* Update to the new engine system.


SirKeplan:
Ah cool, funnily enough i was just about to go through some squad engines myself, if you're doing this(even just this engine), i think it makes sense to do it fully and move the title/description fields etc. across too

PhineasFreak:
Yes, i have some engines prepared. I was aiming to update the Squad engines at first and then continue from the most used mod packs to the less ones.

OK, time to break the editor for good!

PhineasFreak:
I will close this PR as it will be part of a bigger one for all the Squad engines.

PR #940, opened 2016-02-13
Global Engine Configs - CryoEngines

SirKeplan:
~~RS-68 still to do.~~ done for cryo


stratochief66:
@SirKeplan could you add some comments stating what is actually done in this PR? From the name, i assumed it was applying global configs to some of the cryoengine parts, looking into the code it appears you add some features to some of the existing configs (yay!, but not at all evident from the name or notes)

Could you adjust my SSTU J-2X to fit with the ullage now included in the J-2X template (pretty please)?
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_J2X.cfg#L30-L48

SirKeplan:
@stratochief66 ok will do, i was going to do pretty much that anyway for engines affected in other mods.

and the changes here, it's pretty much just moving the ignitions, gimbal, and descriptions across like they should be, but there's a little more i might add to this PR.

SirKeplan:
@stratochief66 J-2X ignitions count has changed, but there is a source.

stratochief66:
@SirKeplan if there are too many changes to a file for it to show on the "Files Changes" page of a PR, I can't review it, so I can't merge it, simple as that. @NathanKell might have a different criterion, but that is mine.

NathanKell:
I too am a sucker for the web interface, so I'd prefer that as well. :)

SirKeplan:
@stratochief66 @NathanKell am i missing something here? this PR only has fairly short changes. did you mean to be commenting on PhineasFreak's PR?

NathanKell:
Ya, this one shows fine for me. :)

SirKeplan:
This should be ready to go now.

PR #944, opened 2016-02-15
FASA agena b

raidernick:
-Add agena b
-Rename agena d guidance unit to just agena guidance unit so as not to confuse people as it can be used on agena b also.


raidernick:
@stratochief66 I was working with an old copy of RO and I did not have those engine changes. I did not change anything related to the parts, only added new configs. I just made sure I had an updated file and made my changes again here and it didn't re-add the old configs this time.

stratochief66:
looks good to me; @NathanKell ?

Issue #939, opened 2016-02-13
Global Engine Configs progress


SirKeplan:
just an issue for tracking which mods have full global engines added, and who is working on them. per mod is not always best as some mods are large, but we can note that.
## Mods
- [x] Squad - @SirKeplan 
- [x] CryoEngines - @SirKeplan #940 
- [ ] KW Rocketry (this is getting by me when i do engines in other packs anyway)
- [x] Bobcat Soviet engines
- [x] SSTU (mod is constantly updating however)
- [x] AIES 
- [ ] BahamutoD
- [ ] FASA
- [ ] RLA
- [ ] CMES
- [ ] Tantares
- [ ] TantaresLV
- [x] SXT
- [ ] Laztek -  Partially done
- [x] Starshine apart from Athena
- [ ] VSR #1071 
- [ ] OLDD
- [ ]  AtomicAge
- [ ] KK Launchers
- [ ] KSLO
- [ ] Taerobee
- [ ] Raidernick - @raidernick

PhineasFreak:
Since this is a community thread, i want to ask for some help: the Squad LR-79 engine includes all possible engine configurations (up to RS-27A) but the actual global engine configs are separated (a logical thing).

What engine configuration should be used? Only the LR-79, the RS-27 or hack it a bit and include both?

SirKeplan:
For the LR79, i think we decided there should be a clone. So i think that makes it one engine for the LR79 and one that is H1/RS27.

PhineasFreak:
OK, on it. Thanks for the info!

SirKeplan:
No problem, another thing @PhineasFreak, you might want to do the Squad Engines in 2 or 3 separate PRs, as github doesn't show the diffs when there are a lot of changes to 1 single file.

PhineasFreak:
Yes, that's what i already do, don't worry.

Also, we have to port all the required fields and modules to the global engine configs. Should i do it as i go and release them along with the engine overhaul or create new PRs for each global engine config? I am afraid that it will clutter severely the GitHub PR list and confuse everyone with the cross referencing.

stratochief66:
@SirKeplan @PhineasFreak do you still want this open for organization/tracking reasons?

Were there any big mods remaining that you wanted to move over to the Global Configs?

Perhaps it would make sense to write a short tutorial w/documentation on how to make use of them so they can easily be applied to new mods/parts?

SirKeplan:
[...]

PR #945, opened 2016-02-15
RO squad engines update

PhineasFreak:
Note: still a WIP, do not merge yet.

* Update the Squad engines to the new centralized engine system.
* Update the global engine configs wherever required.
* Separated the LR-79 and H-1/RS-27 as per @SirKeplan (https://github.com/KSP-RO/RealismOverhaul/issues/939)
* General cleanup.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_H1_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_H1_Config.cfg
@@ -1,12 +1,48 @@
-// this engine is the H-1 and RS27
-// Squad, FASA, KW
+//  ==================================================
+//  H-1/RS-27
+
+//  Throttle Range: N/A
+//  Burn Time: 165 s (H-1), 275 s (RS-27/A)
+//  O/F Ratio: 2.23
+
+//  Sources:
+//      http://www.rocket.com/rs-27a-engine
+//      http://www.alternatewars.com/BBOW/Space_Engines/Rocketdyne_Engines.htm
+//      http://www.apollosaturn.com/sibnews/sec4.htm
+//      https://web.archive.org/web/20041022144943/http://history.nasa.gov/SP-4206/ch4.htm
+//      http://www.astronautix.com/engines/rs27a.htm
+
+//  Used by:
+//      FASA
+//      KW Rocketry
+//      Squad
+//  ==================================================
+
 @PART[*]:HAS[#engineType[H1]]:FOR[RealismOverhaulEngines]
 {
+    %title = H-1/RS-27 Series
+    %manufacturer = Rocketdyne
+    %description = The H-1 is an upgrade to the LR-79 engine that propelled the Saturn-I and IB vehicles, as well as late-model Delta rockets (as the RS-27). The H-1/RS-27 are optimized for the first-stage main engine role. The RS-27A has a higher expansion ratio for increased performance at altitude since liftoff thrust on the Delta II is augmented by solid boosters and the core burns rather longer. [1.0m]
+
+    @MODULE[ModuleEngines*]
+    {
+        %EngineType = LiquidFuel
+    }
+
+    @MODULE[ModuleGimbal]
+    {
+        @gimbalRange = 8
+        %useGimbalResponseSpeed = True
+        %gimbalResponseSpeed = 16
+    }
+
+    !MODULE[ModuleAlternator]{}
+
 	MODULE
 	{
 		name = ModuleEngineConfigs
 		type = ModuleEngines
-		configuration = H-1b
+		configuration = H-1
 		origMass = 0.988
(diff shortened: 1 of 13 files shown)


SirKeplan:
Looking good, some things i noticed looking through the changes.

- ignitor resources: when i asked @NathanKell about the inconsistent use of fuel as an igniter resource on some engines, he suggested leaving them out entirely for the meantime(as the fuel consumption would depend greatly on the type and size of the engine)
- LMAE ignitions count: hmm another thing we need a source on
- What is the source on the new LMDE mass value? just averaging the two?
- Noticed you are adding helium to some hypergolic engines, not a criticism, just wondering what the rationale is.
- Crash tolerance and breaking tolerance etc(RD-253): looks like you deleted these, thid is consistent with other configs, but should we have these values specified for all engines in some way?
- do we want to include `!MODULE[ModuleJettison]{}` in the configs which you put it in, i assumed it would be better to leave this in the .cfgs for the mods?
- maybe include the ModuleAlternator bit in the NSTAR config?

Also, the Squad_Engines file is too big to review on github independently :P, though i realise we can look through the commits.

stratochief66:
The RO_Squad_Engines.cfg changes need to be broken up, as currently there are too many changes in the one file to make it reviewable.

Personally, I am fine with the ignition data containing fuels right now, since the goal of this project at this stage is to move over what already exists in various RO configs into the new form. At a future time, we will be cleaning up ignitions and fuel amounts for restarts.

So @PhineasFreak, @SirKeplan I would say not to worry too much about getting those fuel resources for re-ignition precise, unless @NathanKell disagrees strongly with that.

NathanKell:
It is easy enough to not copy them over, so I would say for consistency we should not use them for now, and instead let RF handle it (by starting up at low throttle/Isp). The startup resource requirement was a crutch, IIRC, to simulate low-Isp during engine start transient, but it's better to do that in RF itself.

stratochief66:
@SirKeplan if there are too many changes to a file for it to show on the "Files Changes" page of a PR, I can't review it, so I can't merge it, simple as that.

SirKeplan:
@NathanKell with not including fuel ignition resources, does that apply to the HTP in the RD-107 too? #934

NathanKell:
@SirKeplan yes, no extra HTP should be spent for igniting the engine at the moment.

PhineasFreak:
@SirKeplan good questions, let me adress them:

* I did not know that RF already covered the ignition resource situation. I will correct them as i go.
* A NASA document (see end of post) mentions that the ignition count for both LMDE and LMAE was 2. I assume that the 10 ignitions were set for game-play reasons.
* From the same document i got the inert mass. I am not sure though because different sources mention wildly different values.
* The pressure - fed engines by definition require a way for propellant pressurization. Helium is the mostly used noble gas for that application (used in way like the HTP in the soviet engines). I asked Nathan for that in the KSP forum and, judging from his reply, he agreed that we should implement that.
* Crash tolerances and breaking forces are already set by the "RealismOverhaul_Global_Config" to their respective values (or if it a copied part, from the parent one). I guess that we could specify custom ones but it should be then done for every part that RO supports and have standards for each category (engines, structural parts, ulitity, electrical etc).
* If a patch for ModuleJettison exists somewhere then i missed it and it should be removed ASAP. It glitches the engines badly...
* The NSTAR config is a bit hacky: it uses a ModuleAlternator with a negative output to simulate the electricity consumption for every throttle value. KSP cannot use ElectricCharge in a sane manner as a propellant resource (if you try then it ends like requiring absurd amounts of it).

I though that the reviewing was done via a git client and/or via a text editor. You are correct to indicate that you cannot view the changes via the web version of git but i do not know how to break it in smaller parts (that's why i was pushing them as separate PRs).

SirKeplan:
@PhineasFreak ahah, that all makes sense then, didn't realise RealismOverhaul_Global_Config set the crash tolerances etc.

Yeah, there's no quick and simple way to split a PR, that's why i suggested doing it split into 2 or 3 before you made this one. oh well.

NathanKell:
Apologies if I gave the impression that we should be, right now, including pressurization resource requirements for engines. We should not. Indeed, the tank masses for highly-pressurized tanks are set precisely so as to abstract the mass of pressurant. To start adding pressurant requirements to engines would mean I would need to change up RF's tank mass system.

Regarding ignition resources: I meant that RF can (it is possible) and RF should (I will do it soon) handle the low-Isp period of starting an engine. It does not do so now. But I still believe those IGNITOR_RESOURCE requirements should be removed for anything that is also a PROPELLANT.

stratochief66:
@PhineasFreak bump, regarding turning this into multiple PRs so we can get started reviewing this?

SirKeplan:
#967 for my edits to this.

PR #950, opened 2016-02-19
SRM Configs: ATK Pack

A1Ch1:
Add configs for motors in ATK Pack. Integration of pack will be in separate PR.


SirKeplan:
@A1Ch1 what we have started on recently is moving `title`, `description`, and `manufacturer` as well as gimbal and fuel ullage etc(ullage obviously not relevant here), into the the new EngineConfigs files too.

A1Ch1:
@SirKeplan yes, that will be step two. I can already only fit two or three SRM configs into a single PR before the changes are no longer visible.

SirKeplan:
Ah great, makes sense.

PR #959, opened 2016-02-19
Added AJ10-Mid reliability

mellon85:
Configuration copied from TF realism overhaul (was assigned at wrong engine anyway)

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_Mid_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_Mid_Config.cfg
@@ -98,8 +98,11 @@
 		reliabilityStart = 0.99
 		reliabilityEnd = 0.998
 		reliabilityDataRateMultiplier = 2
-        	techTransfer = AJ10-37,AJ10-42,AJ10-142:50
+        techTransfer = AJ10-37,AJ10-42,AJ10-142:50
 	}
+}
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[AJ10-118E]]]:BEFORE[zTestFlight]
+{
 	TESTFLIGHT
 	{
 		name = AJ10-118E
--- a/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_Mid_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_Mid_Config.cfg
@@ -89,3 +89,24 @@
 		}
 	}
 }
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[AJ10-104D]]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
+	{
+		name = AJ10-104D
+		ratedBurnTime = 377
+		reliabilityStart = 0.999581 // ignition chance will be this, reliability will be (1-this)*0.0025
+		reliabilityEnd = 0.999979 // and the number for the 10000 data end of the curve
+		reliabilityDataRateMultiplier = 2
+        techTransfer = AJ10-37,AJ10-42,AJ10-142:50
+	}
+	TESTFLIGHT
+	{
+		name = AJ10-118E
+		ratedBurnTime = 377
+		reliabilityStart = 0.99
+		reliabilityEnd = 0.998
+		techTransfer =  AJ10-37,AJ10-104D,AJ10-42,AJ10-142:50
+		reliabilityDataRateMultiplier = 2
+	}
+}
--- a/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_Mid_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_Mid_Config.cfg
@@ -95,10 +95,10 @@
 	{
 		name = AJ10-104D
 		ratedBurnTime = 377
-		reliabilityStart = 0.999581 // ignition chance will be this, reliability will be (1-this)*0.0025
-		reliabilityEnd = 0.999979 // and the number for the 10000 data end of the curve
+		reliabilityStart = 0.99
+		reliabilityEnd = 0.998
 		reliabilityDataRateMultiplier = 2
-        techTransfer = AJ10-37,AJ10-42,AJ10-142:50
+        	techTransfer = AJ10-37,AJ10-42,AJ10-142:50
 	}
 	TESTFLIGHT
 	{


mellon85:
I tried to apply them to different setup, but it the configuration were not applied, if I am remembering right just one of them gets applied in the end (or overwrites the other for all I know).

Actually I should fix the reliability as stated in the comments from NatahanKell. The first value should be the ignition chance while I used the reliability values from the AJ10 configuration from TF (that were applied to the AJ10-Early instead of the mid).

The interpretation is harder actually, chances are higher at the beginning and at the end of the rated burn time, and grows worse later on, it's not linear.

stratochief66:
@NathanKell, shouldn't there be a way to apply the configs separately, or
is the way we want to proceed with multi-config engines?

I know there is a second function overlayed/multiplied that makes things
more risky right at ignition, and cranks on some multiplier as you approach
and exceed the rated burn time, but aside from that second function is my
breakdown relatively correct?

mellon85:
I would say the actual success probability is higher as you are considering the worst chances at each single point in time, while from burn time to rated burn time the chances are lower. But I can't actually say much about it, I don't know how it's implement behind the scene

NathanKell:
The LMDE example in the repo applies all three to the LMDE part, so yes, they can apply.

I went with 10000 data as standard to simplify the configs. If you want to change the rate at which data is accrued during flight, see https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/TF_LMDE.cfg#L21-L22

As for how a given chance in TF works, IIRC it is not a strictly per-second chance. @anxcon or @Agathorn could say better. For now I would look at the existing engine configs in TestFlight to figure out appropriate values for a given technology level. They translate here as: reliability Start and End directly equate to the ignition success chance at 0 and 10000 data, and the fail chance during burn start and end are (1 - reliability start) * 0.0025, which broadly comports with the (existing) configs I surveyed in TF.

mellon85:
You are both right, I must have had a broken configuration in something when I was trying splitting them up in 2 different configurations.

mellon85:
Then I have to ask where to put the configurations. Splitting it in this way allows to have the configuration in any file as the configuration is applied to the configuration and not the engine.
Any idea or should they stay where they are?

NathanKell:
Looks like a good place to me!

stratochief66:
Yep, I would say the location is good for now. Once/if the library grows,
we will probably end up giving them their own folder, but no need to worry
about that right now.

NathanKell:
The code looks fine, but a question on the numbers: Were they really that reliable? 104D was Able-Star, and out of 10 flights of the first mark, it failed to ignite once, and produced reduced thrust once. Of the 8 flights of the second mark, it failed to ignite once. As for AJ10-118E...not that many flights, and no failures. But I'm not sure that I would say (per the 99.8% reliability figure at the end of testing) that it really only had a 1/500th chance of failing to ignite, and a proportionally small chance of failing during its burn.
You may have based that on my using those numbers for the LMDE. But the LMDE is really a special case here, it was an engine that *had* to work. (Not quite as desperately as the LMAE, since an LMDE failure only *might* strand astronauts at the moon, rather than *certainly* stranding them as an LMAE failure would, but still). That's very different from your average LV engine.

NathanKell:
Actually, the numbers are a bit worse than that because we need to discard any Ablestar flights where the Thor stage failed as not stating one way or the other whether the AJ10 would worl.

mellon85:
Actually i copied the values from the AJ10-104D in TestFlight, where it is applied to the wrong engine, I didn't do much research on it. And I just noticed that I clearly put wrong values.. as per 

https://github.com/KSP-RO/TestFlight/blob/master/configs/RealismOverhaul/RO_Engines.yaml

it should be 0.98 and 0.998

NathanKell:
Ah, ok. That makes sense, although I think @jwvanderbeck may be overestimating ignition reliability (although that may well lead to correct numbers for total reliability).

jwvanderbeck:
I may have.  I recall in my research that once I got past early era engines it became harder and harder to find concrete numbers, son instead I started using numbers that gave good progression instead.

NathanKell:
Ah, that could well be :)

mellon85:
I'll put it to 0.98 and 0.998 then, or if you prefer to have a rougher start 0.97?

NathanKell:
Going by the historical data, though, there was 1 ignition failure, 1 burn failure, and 1 unstated failure (could be either), out of 17 flights. That's somewhere between 88% and 94% ignition chance, and about the same odds for successfully completing the burn. My guess is that would push the reliabilityStart even lower (since this engine did build on prior ones, and was static-fired), and the ending reliability only slightly better than those numbers.

NathanKell:
[The historical data](http://space.skyrocket.de/doc_lau/thor_ablestar.htm)

mellon85:
There's also that the AJ10-142 in AJ10-Early has 0.95 and 0.995 ignition chance in test flight (and those configurations should maybe be put in RO than in TF).
True that over 19 launches 3 had this engine failures, and 4 other we can't say.
If we don't count the Thor failures it has a 20% failure rate. And in those 4 launches it means it might have failed at let's say once more. Hence 4 failures over 19 for a a total failure rate of 21%.

A1Ch1:
Failures of a lower stage have to be classified as "no test", statistically the test has no value. So 3 failures in 15 trials, 80% reliability.

A1Ch1:
@mellon85 @NathanKell To estimate reliability we'd generally use a Bayesian estimate to give an estimate of the chance of success in a subsequent trial:

x = (k+1)/(n+2)

where x is the probability, k is the number of successes, and n is the total number of trials. So here we have (12+1)/(15+2) = .76

Using the Bayesian estimate also allows us to calculate reliability even in cases where all trials are either failures or successes without having to use 0% or 100%.

NathanKell:
@A1Ch1 good point.

@mellon85 yes the other AJ10s probably need revision as well.

One other point: due to how TF works, we *do* need to separate ignition fail chance from any during-the-burn failures. So if you want a total failure probability it would be (1 - reliability) [for the igntiion fail chance] + (burn time * reliability * 0.0025) [for the fail chance during the burn]

I am now wondering if I should normalize the latter by the burn time, so the reliability given is reliability *for the stated burn* not reliability *at any given second of the burn*.

NathanKell:
Hopeful bump :)

jwvanderbeck:
That is a mighty impressive MM script you have there!  Obviously the
abstraction costs you flexibility though and this is an example of that.  I
really think ignition settings should be separate.  Looking at what you
have now actually I am not sure how you handle engines designed to light
under different conditions.

- John Vanderbeck
- http://www.johnvanderbeck.com

NathanKell:
@jwvanderbeck we can handle ignition reliability differing from duty-cycle reliability via the spread parameter. Regarding ignition conditions, I looked at your existing RO configs and didn't notice any difference in those dynamic pressure curves between boosters and uppers, so I figured it wasn't worth adding special behavior here. That said, anything generated by this script can be changed in AFTER[zTestFlight] so things can be customized. :)

NathanKell:
And thanks! @sarbian added some bits to MM specifically for this, so it'll get cleaned up a lot (no need for multiple copies of the node, ability to apply to arbitrary CONFIGs even without nodes created!)

jwvanderbeck:
Well to be clear in terminology here, if I understand you correctly I
believe what you are calling "ignition reliability" is just the reliability
curve of the engine, and the fact that it is more likely to fail during
those first few seconds.  But that failure is generic, not necessarily an
ignition failure but just a failure and it rolls on the table.

What I am referring to is the ignition pressure curve which allows an
engine to specifically have a chance at an ignition failure based on
(dynamic?) pressure.  So that engines that aren't designed to be air lit
can be modeled as such.

- John Vanderbeck
- http://www.johnvanderbeck.com

NathanKell:
@jwvanderbeck  Nope, I'm talking about baseIgnitionChance in that module ( TestFlightFailure_IgnitionFail ), which is not the cycle-based failure multiplier nor the general "while running" failure chance. That is a pure chance curve modified by flight data.

Regarding the pressure curve in that module, it is (AIUI) applying a multiplier to the base ignition chance based on the dynamic pressure. That is the curve that I did not see vary across your configs, and therefore kept. Air lighting is hard, pretty much no matter what kind of engine you have. :)

jwvanderbeck:
Ah yeah forgot about that value oops :p



- John Vanderbeck
- http://www.johnvanderbeck.com

NathanKell:
@mellon85 sorry, I changed the syntax of these slightly. You're tagged in the LR101 commit changing that, but can you also add the Q multiplier [as demonstrated in the LMDE config](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/RO_LMDE_Config.cfg#L174-L203)?

NathanKell:
@mellon85 now the reliability in both cases can be specified directly. I suggest using @A1Ch1's method.
Ignition reliability: (15+1) / (17+2) = .842 (and let's say that's around the midpoint, so start would be .8 and end would be .86)
Cycle reliability: (14+1) / (16+2) = 0.833 (and as above, so let's say 0.8 and 0.85)

mellon85:
@NathanKell I was trying to derive your numbers from http://space.skyrocket.de/doc_lau/thor_ablestar.htm but I can't get the same ones. For Ignition we have that over 19 launches there were 5 failures about it of which 4 where on the lower stages or separators and 1 a real ignition fail.
So it's (14+1)/(19+2) = 0.714 
And for cycle of all the launches there is an additional 2 failures of the completion of the burn (flight 14 and 9), hence
(12+1)/(19+2) and gives me 0.619.
I am probably missing something :)

A1Ch1:
@mellon85 Failures of lower stages don't count; since the AJ10 never fired those flights give us no information about how the engine we're looking at performed. So now we have 17 tests instead of 19. Of those, there were two ignition failures (counting insufficient thrust as an ignition failure), flight 9 and flight 10. So (15+1)/(17+2).

For burn reliability, I'm thinking @NathanKell has (14+1) / (16+2) from excluding the outright ignition failure (flight 10), leaving us with 16 tests and 14 successes. I'd additionally exclude flight 9, since we've already counted the insufficient thrust as an ignition failure (maybe there's a way to simulate this kind of partial failure?), resulting in 14 successes in 15 trials: (14+1) / (15+2)=.882

NathanKell:
@A1Ch1 yes, reduced thrust is a failure mode in TestFlight, it can happen either during ignition or during the flight. :)
And yes you're correct as to why I did those numbers. However, I violated the strict algorithm you presented since it seemed odd that despite having fewer runtime failures than ignition failures, and having a better unmodified ratio, the chance of a runtime failure ended up *higher* with the Bayesian method.
Should that still be gone with?

NathanKell:
Ohhh I forgot to subtract the ignition failures. @A1Ch1 you're correct. :)

A1Ch1:
@NathanKell Hmm, it seems like we're including flight 9 as both an ignition and a burn failure. If we count it just as an ignition failure and don't include it in the stats for burn failures themselves, we're left with one burn failure in 15 trials, which gives us a lower chance of burn failure (11.8%) than ignition failure (15.8%)

A1Ch1:
@NathanKell Ah, alright :)

NathanKell:
Oops, I made this conflict. I'll merge manually and apply A1Ch1's numbers. @mellon85 thanks! :)

PR #961, opened 2016-02-20
NERVA Reliability

mellon85:
Added NERVA reliability

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_NERVA_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_NERVA_Config.cfg
@@ -34,3 +34,14 @@
 		}
 	}
 }
+# http://www.alternatewars.com/BBOW/Space_Engines/NERVA_PV.png
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[NERVA-NRX]]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
+	{
+		name = NERVA-NRX
+		ratedBurnTime = 36000 # 10 hours
+		reliabilityStart = 0.999994
+		reliabilityEnd = 0.999997
+	}
+}


NathanKell:
That seems...remarkably reliable.

mellon85:
that's the only numbers I could find from their specs.. the fact that it was never built is skewing them a lot. I wouldn't know what else to put as starting reliability

stratochief66:
Since it is rated for such a long time, that might help bring the true
failure rate up (assuming these failure changes are per second)

If this was the only factor (I know it is not) in a 10 hour burn, it would
have a 20% chance of failure at start, 10% at max data

NathanKell:
The underlying failure rate will be (1 - (the specified reliability) ) * 0.0025, so 0.99 equates to a base failure rate of 0.000025. If that is applied once per second (which may or may not be the case--I am not very familiar with how it is employed, @anxcon @jwvanderbeck ) that is a failure chance of 0.0025% per second, or 0.15% per minute, or 9% per hour.

NathanKell:
Uh, that makes the starting reliability equate to a failure chance of 0.13% per _day_ of burn time (!)

anxcon:
while not exactly 1 second (its 1.0 + whatever until next pass) you're close enough
and that only counts time during (for engines) running the engine

reliabilitycurve is mapped for part data 0 to 10,000 (usually)
each second has a roll, when roll < value at that spot in curve, failure

there is momentary failure rates, but that requires a 2nd custom module extending TestFlightReliability and it be setup as momentary, i believe engines have one for startup related extra fail chance (havent looked at it recently)

jwvanderbeck:
Engines have a non-linear momentary adjustment so that they have a higher chance of failure at first ignition, and then a progressively higher and increasingly severe chance of failure as they exceed their rated burn time until eventually they just have gotten so far past that they are practically guaranteed to fail.  This is the `cycle` property on `TestFlightReliability_EngineCycle` .  For example:
https://github.com/KSP-RO/TestFlight/blob/master/configs/RealismOverhaul/RO_Engines.yaml#L326

That engine will have 10 times its normal failure rate from 0-5 seconds after ignition, after which the multiplier drops to 1 from 5 seconds to 155 seconds, then it starts rising again until it hits a multiplier of 100 times failure rate at 267.50 seconds.

jwvanderbeck:
That said looking at this commit it appears you guys are doing something new with the configs? So I'm not sure what is up :)

anxcon:
@jwvanderbeck NathanKell basicly made an MM script to setup testflight using that above TESTFLIGHT node, which gathers some values from ModuleEngineConfigs and such - basicly skipping the yaml that nobody seems to know how to use heh plus benefit of being more inheriting of changes to other RO MM scripts

NathanKell:
@jwvanderbeck yeah, basically I took your existing RO configs and tried to boil them down such that they can be generated with *only* burn time, reliability start, and reliability end (with optional tuning factors like how sharp the kink in the reliability curve is). It links ignition failure chance and constant failure chance by a ratio of 0.0025 and it maintains the existing cycle setup (bad for the first five seconds, then ok til burn time + 5, then climbing to 10 and burn time * 1.5). 

[You can see the generation code here.](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg) And [here you can see a commented example](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/TF_LMDE.cfg)

NathanKell:
@mellon85 what numbers did you find in the specs?

mellon85:
Only reliability, service life and number of restarts

you can find restarts and reliability info here
http://www.alternatewars.com/BBOW/Space_Engines/1992_NERVA_Derived_Engines.pdf
http://www.alternatewars.com/BBOW/Space_Engines/NERVA_PV.png
http://www.alternatewars.com/BBOW/Space_Engines/NERVA_II_Specs.gif
http://www.alternatewars.com/BBOW/Space_Engines/NERVA_Nozzle.png

but having been used for real launches makes it complicated to calculate anything

NathanKell:
Ok, if that's the reliability then I would take just as chance of successfully completing the entire burn time. So you may indeed already have the right numbers, sorry.

mellon85:
The problem is that it's really really high, and actually I don't know how to express that the 10h lifetime is in 10 60-minutes burns.
And it's still possible to start with a lower reliability to actually have some testing effort, but I would have no idea what to consider fair/realistic..

NathanKell:
@mellon85 sorry, I changed the syntax of these slightly. You're tagged in the LR101 commit changing that, but can you also add the Q multiplier [as demonstrated in the LMDE config](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/RO_LMDE_Config.cfg#L174-L203)?

Regarding NERVA reliability: Just specify the stated reliability now, the burn time will be included automatically.

mellon85:
for the LMDE I see you put 0.1, as it's not meant to ignite in atmospheric condition, but I can't find any data regarding NERVA or the LR101

NathanKell:
Hmm, this is now failing to merge as well. Let me pull it in manually then. And yeah, like, like the AJ10, it should have that line. The LR101 was a first-stage vernier, so it lit on the pad. :)

PR #963, opened 2016-02-21
Change to SXT Station and Base parts configs

Zarbizaure:
A patch to #892 as @NathanKell requested.

Changed the node configs, it should be fine now.
Rescaled the PDD4 and PDD6 to 3m witdh (The Salyut Command Desk were 2.9m width; it's not exactly a counterpart but it will give more flexibility for building station).
PDD6 now holds just 3 crew (as Salyut/Almaz stations)
The Skylab now features an orbital lab.
[Station Part Sheet](https://onedrive.live.com/redir?page=view&resid=9AE39A0EBF42E208!360496&authkey=!AOv701J1tt1oiA8) : if you want to add some details or ideas about the config you can do it there.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
@@ -4,7 +4,7 @@
 //  Realism Overhaul configuration.
 
 //  Dimensions: 6.0 m x 4.0 m
-//  Gross Mass: 7815.0 Kg
+//  Gross Mass: 5800.0 Kg
 //  ==================================================
 
 @PART[SXTSPKTRCabin]:FOR[RealismOverhaul]
@@ -22,13 +22,12 @@
     @rescaleFactor = 1.0
 
     @node_stack_top = 0.0, 1.912, 0.0, 0.0, 1.0, 0.0, 3
-    @node_stack_bottom,1 = 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, 3
-    @node_stack_bottom,2 = 0.0, -1.843, 0.0, 0.0, -1.0, 0.0, 2
+    @node_stack_bottom = 0.0, -1.843, 0.0, 0.0, -1.0, 0.0, 2
 
     @title = SPKTR-10 Lacuga Storage Container
     @description = Hold 2 people with 30 days of supplies. An extension adapter for the PPD-10 Hitchhiker storage container.
 
-    @mass = 7.8
+    @mass = 4.8
     @maxTemp = 1073.15
     !vesselType = NULL
 	!RESOURCE[ElectricCharge]{}
@@ -111,7 +110,7 @@
 			amount = 45
 			maxAmount = 45
 		}
-	}
+    }
 	
 	MODULE
 	{
@@ -128,8 +127,8 @@
 
 //  Realism Overhaul configuration.
 
-//  Dimensions: 3.0 m x 1.6 m
-//  Gross Mass: 3600.0 Kg
+//  Dimensions: 3.0 m x 1.8 m
+//  Gross Mass: 3200.0 Kg
 //  ==================================================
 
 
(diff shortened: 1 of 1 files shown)


NathanKell:
Regarding the science lab--good point! :)

Regarding the PPD-4 and -6, I think the *stats* you had for them before were fine; they represent early (MSS, proposed early Apollo stations, early Soviet station ideas) station parts just fine that way. The only problem was the size, and the temps (which change should be applied to all station parts).

Regarding the fixes I committed--you are now back to no longer rescaling the extra MODEL nodes on those parts I touched, and missing the node_attach attach node. Please don't undo those fixes, instead apply them to the other parts. Also, the scaling I had in there for PPD-4 and -6 rescale the original 2.5m to 3m; were the original parts not exactly 2,5m? They appeared so to me ingame.

Is the CANIOT-7 statted to be part of Salyut 7 as its name implies? In that case it might be worth changing its name to the romanization.

Zarbizaure:
Oh well, I hadn't seen your commit. Seems great. What exactly is doing the node_attach node?
I will take your configs for the PDD4 and 6. About the CANIOT 7 (meaning Salyut) I think we should find it a better name, because it's an heavy orbital module (or, we could rescale it down to 2.9 or 4.15m, but we already have Salyut components). Is there any ~5m witdh station in real life?

NathanKell:
node_attach is the node that allows you to surface attach a part to another part. Basically, you need to patch everything in the original cfg: if it has 3 MODEL nodes, patch all three; if it has 4 different node_whatever lines, patch all 4, etc. :)

I'm not aware of anything that's 5m offhand--all the Freedom/Fred/ISS stuff is wider AFAIK, maybe one of Korolev's early station plans for original-N1 launch? It can just be a generic 5m hab section though....

Zarbizaure:
Ok, I'll do my best to change these nodes configs.
About the mass: why are yours so heavy? If we take the PDD10 as reference (6.5T empty), PDD4 and 6 should be lighter (2.5 and 5T?). Or maybe PDD10 config is wrong.
So we need to find another name for the CANIOT 7 (since there isn't any 5m width space station...), what about Turtle?

NathanKell:
RO doesn't support TweakScale, so I would highly recommend not having it installed when making RO pull requests. :)

I kept the mass of them as they originally were, which was slightly higher per person than the Hitchhiker as RO configures it. That makes sense for their intended role, which is early (early-mid 1960s) hab modules rather than later (1970s+ ) hab modules.

If it's just going to be fairly generic, I'm not sure we do need a new name...

Zarbizaure:
I uninstalled it and the nodes/scale were ok, you could double check if you want.
It still needs to be a bit lower than the PDD10...so 7T is too high for the PDD6; it's possible to use lighter materials, but it does not explain twice the same volume for a lower mass (PDD6 ~ 107m^3, PDD10 ~ 220m^3).
We could keep the CANIOT 7 name but it may create confusion rather than organizing station parts correctly.

NathanKell:
Apologies, got sidetracked.
The idea with the PPD-4 and -6 is that most of the weight comes from the habitation equipment rather than the pressure vessel itself. The -6 holds 4 people, so it seems reasonable to mass about the same as the -10 (larger, but rather better technology). Is that implausible?

Ah, yes, fair point on the CANIOT. Although it does have a rather different color scheme so there is that, right?

Zarbizaure:
Ok, so it should be just under the weight of the PDD10, but not over. (I don't remember the dry mass of the PDD10, so the PDD6 could weight about one hundred kilogramm less). That is indeed plausible (and mass depends more on wall surface + equipments rather than pressurized volume).
Well, so we just need a better and appropiate name. Actually its color scheme isn't a salyut one but it still looks (a bit like a [salyut core](https://upload.wikimedia.org/wikipedia/commons/d/dd/Model_of_Salyut-7_with_two_Soyuz_spacecrafts.JPEG); however the black and white alternance is more reminding an american module.

NathanKell:
6t and 3.5t make sense. I don't think the crew capacity should be lowered however.

PPD-A maybe?

Zarbizaure:
Fine, we just have to change it back to 4 (I will recalculate the supplies duration btw).
Or PDD-H, the H standing for "heavy".

NathanKell:
PPD-H sounds good! :)

NathanKell:
The 4 person patch doesn't seem to have made it in. I'll do that after the merge, given release is soon. :)

NathanKell:
Also, once again, *please* don't revert my description changes. RO does not use the campy Kerbal descriptions of parts.

Zarbizaure:
Oh sorry for the description overwrite; I did not see that you changed it.
I though it will work by deleting the @CrewCapacity parameter (because the PDD-6 is initially planned for 4 crew) however if it does not work, you just have to set it to 4.

PR #967, opened 2016-02-23
Phineas freak ro squad engines update

SirKeplan:
Edits noted in commit message, only other thing is some engines have had their default config set to the earliest available, this seems to generally make sense though.
There's still more to do in the Squad_Engines file engine wise, this seems to be just part 1, if @PhineasFreak doesn't reappear soon :P, i can do that.

I'll merge this PR later/tomorrow, unless anyone objects.


NathanKell:
Not if I merge it first! :) Thanks for this!

PR #970, opened 2016-02-24
Update RL10B-2 isp and thrust

SirKeplan:
@ctiberious

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_RL10_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_RL10_Config.cfg
@@ -1,6 +1,16 @@
-//RL10 engine
-//Squad, Fasa, KW, CMES, CryoEngines
+//  ==================================================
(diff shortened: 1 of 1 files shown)


SirKeplan:
Hmm, noticed the Delta IV payload guide(2007) has an isp of 462, not the current 465.5. So sometime between 2007 and 2013 the stated performance of the engine changed.

http://snebulos.mit.edu/projects/reference/launch_vehicles/Delta/DeltaIVPayloadPlannersGuide2007.pdf

A1Ch1:
From what I've seen, most earlier sources (Delta III and early Delta IV) give an Isp of 462 s, while the AR website and the later Delta IV guides give 465 s. FWIW most secondary documents (eg evaluations of the engines for SLS and other launchers) use 462 s. The original goal of the RL10B-2 was 465 s but I remember early engines didn't quite meet that. It's possible the engine was later uprated to 465 s.

PR #971, opened 2016-02-25
Fix root rescale bug in ROMini + helper script.

jmrd98:
Hello!
Thanks to an unexpected snow day, I finally managed to do as I have been threatening to, which is to clean up and submit this code for a PR!

Several things to note: 

- The "fix" here is to uncomment some existing code which was said to have failed in 1.0.4. It did indeed fail, such that while the models were rescaled correctly, the attach nodes were not. This is somewhat less then satisfactory from a gameplay standpoint.

- By using this method, we avoid entirely bug where a root part scaled with rescaleFactor will revert to its original size on load/quickload.

- This brings us to part two: The python script. This is a reimplemented shell script I posted on the forums a while ago. This script walks through a directory tree, opens each .cfg file found and does a regex match for any line starting with "node". We collect that into an internal list, remove duplicates, and sort it, yielding a list of every unique type of node_* in your Gamedata folder. We then feed that into a template for a MM script to rescale the nodes based on a chosen rescaleFactor. 
It sounds complicated, but trust me, it works *very* well. (NOTE! In testing tonight, I had it blow up on a "node_collider = node_collider" line; I hadn't ever seen that before, except in one old part mod I was testing, so I hope it isn't TOO common... I will have to handle that in the code if it is.)

The usage is simple, simply run it from Gamedata/ (or KSP root, it doesn't care), and specify an output file for your generated MM config. Example: ./generateNodes.py Gamedata/zROMini/scaledNodes.cfg
It was written in python 2.7, but I honestly can't think of anything that would goof in python 3. (No promises, but let me know if it does, and i'll look into it.)
It should be perfectly content to run under windows thanks to the magic of python, as well!

- Of further note: I have spent a fair amount of time going back and forth with procedural fairings vs rescaling. At the top of ROMini you will find a section of code intended to bring PF into line, but it seems to fail at this time. No errors result, just no... results result. I have left it in so that someone else may either find my silly mistake, or else please feel free to cut it from the patch if it is out of scope for this project. 
However, the main rescaling code explicitly excludes PF in order to make them work at all. The result is a game which has all it's dimensions seemlessly scaled by 1.6 in this implementation, but for the PF numbers. (They are unscaled, but working correctly in all other aspects.)

- ... There is no last bullet point. 
The code was a GREAT help to build off of, i'm sure I would be still on the starting blocks rescaling if I started from scratch. It seems quite stable in my testing. Balance wise, the only thing I would try to look at is the SRBs, but that is out of scope for this PR.

I will leave it at that, as this note has gone on far too long! Licensing! The python script is for all practical purposes public domain, but CC-BY-4.0 as indicated in the guidelines works as well too.

Thanks all!

### DIFF
--- a/ROMini.cfg
+++ b/ROMini.cfg
@@ -142,6 +142,7 @@
 }
 // This _should_ work, but scale seems to fail in 1.0.4...
 // It DOES work in 1.0.5 in all aspects but for not catching attach nodes... doh!
+// Based on the above comments, I expect it *should* work in 1.1, but i'm editing blind at work at the moment...
 @PART[*]:HAS[@MODEL]:FOR[zROMini]:NEEDS[!RealismOverhaul]
 {
 	@MODEL,*:HAS[~scale[]]
@@ -160,18 +161,11 @@
 	}
 }
 // We need to DISABLE this one so we don't scale twice.
+// I don't see why we wouldn't have to keep this disabled/removed under the new 1.1 conditions.
+// Is this not redundant with the @scale in the model node above?
 //@PART[*]:HAS[@MODEL]:FOR[zROMini]:NEEDS[!RealismOverhaul]
 //{
 //	@scale *= #$rescaleFactor$
-//}
-
-// If this worked, it'd be a miracle, but MM doesn't support this sort of wildcarding, as far as I can tell. LEAVE DISABLED, hope for new features in the future.
-//@PART[*]:HAS[@MODEL]:FOR[zROMini]:NEEDS[!RealismOverhaul]
-//{
-//	testnode = #$rescaleFactor$
-//	@node_*,*[0] *= #$rescaleFactor$
-//	@node_*,*[1] *= #$rescaleFactor$
-//	@node_*,*[2] *= #$rescaleFactor$
 //}
 
 @PART[*]:HAS[@MODEL]:FOR[zROMini]:NEEDS[!RealismOverhaul]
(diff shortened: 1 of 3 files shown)


NathanKell:
Neat stuff! However I'm happy to mention (though it's not in the dev notes yet so it's only quasi-public) that I fixed the rescaleFactor bug for 1.1, so this won't prove necessary fairly soon. :)

As I would like to keep this cfg purely a cfg with no added dependencies, I'd suggest instead you release your work on your own thread (and it's actually a great general purpose tool too of course!)

NathanKell:
Regarding PF...what changes do they need? Ah, so the step sizes matches the new sizes (1m, 2m, 4m, 6m) ?

NathanKell:
1.1 is out! rescaleFactor works!
@jmrd98 can you strip this to just the better scaling?

jmrd98:
Well, my phone made the point for me by eating my post. ("I'm not about to try to edit it on my phone")

Tl;dr: tonight + tomorrow day = shot, but evening is good to go. I'll try to get it in shape. Never figured out the PF thing. Haven't had a chance to load 1.1 yet. Looking forward to trying it on OS X. Many thanks for innumerable fixes that made it in, in case accolades get forgotten in the 64-bit stampede of glee. o7

jmrd98:
Non-update-update: I had hoped to get something out by now: It may be into friday, unfortunately. My work is tied tightly to the academic semester, and we're in the last two weeks before finals here. The 'hair on fire' quotient is rather high...

I just made some blind edits while at work (don't tell). The helper script and sample cfg are now entirely un-needed. I have no idea how to un-PR them. Close this and hit a new PR entirely?

Anyway, i'll try to test the changes, but it'll be a few days in all likelihood. I don't know if my linux will play ball with 1.1 re: the troubles I see on the bugtracker... Hoping OS X will, as mentioned :)

stratochief66:
Yeah, setting up a new PR makes sense. Please add a reference to this PR (#971) as there was a lot of related conversation here :)

NathanKell:
Closing in the hope of a new PR. :)

PR #972, opened 2016-02-25
Global Configs Squad Continued

SirKeplan:
Do not merge yet.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_190_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_AJ10_190_Config.cfg
@@ -73,13 +73,6 @@
                 DrawGauge = False
             }
 
-            PROPELLANT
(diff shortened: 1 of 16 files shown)


SirKeplan:
I re-instated the gimbal on the NERVA, however the engine model doesn't visibly gimbal, am i using the wrong transform??

apart from that, nothing else i need to add to this PR.

NathanKell:
My guess is Ven's model again doesn't support gimbaling since it was removed from the stock NTR in 1.0.

NathanKell:
(Goes on the list of mu's to check I guess, when you have a sec--you have taniwha's blender mu importer, right? )

SirKeplan:
@NathanKell hehe, nope, but i will get it.

SirKeplan:
@NathanKell neither of those Ven's models support gimbal. thrustTransform is an empty node

PR #973, opened 2016-02-26
RN Thor add

raidernick:
-Add thor able configs


raidernick:
Propane was omitted because the resource does no currently exist. This will be changed in future releases when the resource is added.
Sources: 
Engines: Stock engine configs
Able/Altair: http://www.accur8.com/19740072500_1974072500%20Vanguard%20Engineering%20Summary.pdf
Thor Stage: Astronautix various pages
Wikipedia has proven unreliable as their pages often contradict one another, every website seems to have some confusion between the various designations.

PR #974, opened 2016-02-26
RN Thor/Thor-Able/Thor-Delta

raidernick:
-Tweak thor using new values, make WIP parts as WIP


raidernick:
@NathanKell ping me in irc before releasing RO. I found a new source that has every single thor and delta rocket including even the fairing weights. I am going to be trying to get all the configs completely redone and maybe even add some new parts before the release.

PR #975, opened 2016-02-26
Global configs squad nasa

SirKeplan:
The RSRMV thrust curve has changed here.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RO_RL10_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RO_RL10_Config.cfg
@@ -1,6 +1,16 @@
-//RL10 engine
-//Squad, Fasa, KW, CMES, CryoEngines
+//  ==================================================
+//  RL10 family of engines.
 //
+//	Squad, Fasa, KW, CMES, CryoEngines
+//
+//
+// RL10B-2 variant:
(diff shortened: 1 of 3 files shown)


SirKeplan:
should be good to merge if the changed RSRMV curve is ok.

SirKeplan:
tagging @A1Ch1 as it's your curve. plus, i think you forgot to add gimbal to the RSB one? (which is effectively added in this PR now)

A1Ch1:
@SirKeplan Don't think I did RSB motors, so I'm not sure which curve its currently using: the old curve was the curve for the proposed Shuttle Five Segment Booster, and the curve I uploaded to the master config is for the RSRMV (the curves differ in the thrust bucket depth). I've never downloaded the RSB packs, but if the motor is intended for Ares I or SLS it should use the curve in the master config.

A1Ch1:
@SirKeplan Checked curves, yes, the one in the master config should be the correct one here. In fact I'm not sure why the Squad RSRMV doesn't already use the master config since I remember including itin the Squad configs... must have gotten left behind at some point when I split the PRs into multiple parts.

Only thing I'd change at this point is the gimbal range: should be closer to 7 degrees than 3.5 degrees.

SirKeplan:
thanks @A1Ch1, haven't seen a source on that yet, but sounds a reasonable figure.

PR #981, opened 2016-02-27
VSR habitat

winged7:


NathanKell:
@stratochief66 can you have a peek? The base stats look fine but you're the LS expert.

PR #982, opened 2016-02-27
An-225 loading ramp

winged7:


NathanKell:
That seems *remarkably* light for the loading ramp on the largest aircraft ever flown. :]

winged7:
So maybe 5 tons? I have seriously no idea.

NathanKell:
Ok. Let me know when you do have an idea and I'll reopen. :)

PR #983, opened 2016-02-27
SXT science nose cone

winged7:
I've also tried to make Remote Tech configuration for antenna but it didn't work.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/Science.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/Science.cfg
@@ -49,11 +49,26 @@
 {
     %RSSROConfig = True
 
-	@title = GC-MS
+	@title = GC-MS + Antenna
 	@manufacturer = Generic
 	@description = The Gas Chromatograph - Mass Spectrometer (GC-MS) is an analytical instrument used for identification of multiple different substances in a selected gas sample, capable of detecting even trace amounts of them. Ideal of analyzing the atmospheric composition of unknown environments.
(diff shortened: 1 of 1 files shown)


NathanKell:
Didn't work how so? I may be able to see what the problem is if you put the cfg back.

winged7:
@PART[SXTsciencenosecone]:FOR[RealismOverhaul]
{
    %RSSROConfig = True

	@title = GC-MS + Antenna
	@manufacturer = Generic
	@description = The Gas Chromatograph - Mass Spectrometer (GC-MS) is an analytical instrument used for identification of multiple different substances in a selected gas sample, capable of detecting even trace amounts of them. Ideal of analyzing the atmospheric composition of unknown environments.
	@mass = 0.001
	@maxTemp = 1073.15
}
@PART[SXTsciencenosecone]:FOR[RemoteTech]
{
	!MODULE[ModuleDataTransmitter] {}

	@MODULE[ModuleRTAntenna]
	{
		@Mode0OmniRange = 0
		@Mode1OmniRange = 4000000
		@EnergyCost = 0.0015
		@TRANSMITTER
		{
			@PacketInterval = 0.4
			@PacketSize = 0.29
			@PacketResourceCost = 0.0075
		}
	}
}

NathanKell:
Ah. That will only work if the part already has that MODULE (you're modifying it). If you need to add it, you need to actually add the Module rather than using @. Check out how in, say, in the cfgs of RT itself, such modules are added.

winged7:
ok, now it works correctly.

PR #986, opened 2016-02-27
Adjust VSR BE-4 plume

A1Ch1:
Adjust position and energy/speed so plume extends from bell correctly


SirKeplan:
Minor thing. This is the BE-4 Not BE-3.

A1Ch1:
@SirKeplan You are of course correct :) Fixed

PR #998, opened 2016-03-01
FASA Nova fixes

ctiberious:
I wanted to see if I could build the Nova rocket but noticed that RO didn't have a craft file for it.  So I took the craft file that comes with FASA and used that as a template to build my own Nova rocket.  During the assembly, though, I discovered that a few of the parts didn't appear as though they had ever been properly setup for RO.

The 2nd stage fuel tank doesn't have a mounting point for the decoupler that is supposed to be below it.  It also has too many engine mounting points.
The 1st stage fuel tank doesn't have engine mounting points setup correctly.  Engines mount above the mount point instead of below.
The fairing that separates the 1st and 2nd stage isn't long enough to allow the 2nd stage to include an M1 engine which appears to be what the FASA craft file uses.


ctiberious:
Can't figure out how to modify multiple files with a single PR without installing the github desktop which I don't currently have.

NathanKell:
Modify the first file. Create the PR. Then click on the commit so you're viewing the commit. That will take you to your *own* repo, where you can edit files to your heart's content. Make sure the branch you're on is the branch the PR uses. Each commit you do on that branch (which you can do directly via 'commit' since it's your own repo) will get added to the PR.

stratochief66:
I haven't tried to build the FASA Nova since KSP started requiring properly oriented attachment nodes. it could be that the nodes on the stage or the decoupler are not oriented properly (top 'pointed' up, bottom 'pointed' down) 

You say too many nodes for engines, but you don't say how many there are, or how many you think is right. Nova as a pretty general design name that covers rockets substantially larger than Saturn. some plans exist with 8x J-2 engines (which I think matched the number of engine attachment points) and others with 2 M-1 engines) If the fairing for the FASA Nova won't fit the M-1, perhaps it wasn't meant to? if you can find a way to make a fairing/interstage by stretching the part or however to allow either version of the Nova second stage to be made, that would be handy. However, I can't think of a straightforward way to have 2 different mounting plates for the second stage, 1 with 8 mounts for J-2 and another for 2 M-1s. Again, if you can propose a working solution, that would be great :) Personally, I just build Nova-esque rockets using procparts and the customizable engine mount.

ctiberious:
Thank you @NathanKell I knew there had to be a way to do it.  I've consolidated all the fixes for the Nova on this one PR.

ctiberious:
@stratochief66 I used the craft file for the Nova rocket in FASA (FASA Apollo Nova.craft) as the template.  I know there were a number of different design possibilities for "Nova" and I've got no clue which design this craft file is based on but it seemed like a good place to start.  In this craft file:
1) Stage 1 of the Nova included 9 M-1 engines.  The Stage 1 Nova tank that we have in RO has one properly setup attachment point for an engine, plus 8 improperly setup points.  Basically the nodes were "1.0" when they should have been "-1.0".  I also had to reposition 4 of the nodes as they were too close to the center node for M-1 engines.
2) Stage 2 of the Nova included 1 M-1 engine surrounded by 8 J-2 engines (again, in the craft file I based my build on).  All the available nodes are correct for the engines, but there is no remaining node to attach a decouple to.  I just relabeled the attachment points and added a bottom node specifically for a decoupler.
3) The decoupler that sits between stage 1 and 2 in the craft file is "FASAApolloStrJ2StrJ23m5mDecNova".  That's the same part we have in RO_FASA_Nova.cfg.  It just wasn't sized up correctly to cover the M-1 engine that appears in the FASA craft file.

Again, this was all based on the FASA craft file.  I know that wasn't the only possible configuration and it would be possible to add more engine nodes so we could make more configurations available.  But these changes so here were specifically so it would be possible to recreate the only craft template I have to work with right now.

ctiberious:
Actually, that configuration (8x J2 + 8x F1) is also valid with the changes I'm proposing here.

ctiberious:
The craft file from FASA is for a monster of a rocket.  It stands about 170m tall.
FASANova First stage consists of 9 M-1 engines (SLT 1.6, est burn 76s, est dV 1520SL).
FASANova Second stage consists of 1 M-1 plus 8 J-2 engines.  (est burn 168s, est dV 2637vac).
Saturn S-II third stage consists of 5 J-2 engines (est burn 341s, est dV 4892 vac).
Saturn S-IVB fourth stage consists of 1 J-2 engine (est burn 428s, est dV 4353 vac).
Standard Apollo LM
Standard Apollo CSM
This beast should be able to put an additional 85t into LEO meaning it could be a viable launch engine for a manned mars mission.  Again, though, the point of the changes is just to make it possible to create the craft from FASA.  Since we have no RO-Nova craft at this point, it was the only thing I had to work with.

stratochief66:
These node direct changes definitely need to be made. The first stage engine nodes. https://github.com/KSP-RO/RealismOverhaul/pull/998/files#diff-f482f68db86597c18c4090a36b72856eR73

@ctiberious the second nova stage is labelled as being the second stage of the C-8 Nova, Here is a little info about that vehicle design: http://www.astronautix.com/lvs/saturnc8.htm

After you've had a change to see this, I intent to fix those node directions in a separate PR and close this one unless something new comes up.

ctiberious:
@stratochief66 I hadn't realized the Nova parts in RO were intended for a version of the "Nova" different than what comes in FASA.  Since they are, that works for me.  I would suggest, however, that we either include a craft file in RO for the Nova, or add to the description for each part to give a better idea how the rocket is intended to be setup.

stratochief66:
If you want to open a new PR that corrects the node direction, and add a note to each of the Nova parts (like the one currently one second stage) that it is meant to be part of the C-8 Nova, you are welcome to it. As for the craft file, maintaining the existing historical FASA ones is enough of a hassle for me, I don't plan to imagine up a payload stage for the C-8 Nova as well, since none existed. :P

If you search rocket repositories, or even Wiki and Astronautix I don't believe you will find a craft with more than 2 M-1 engines on the second stage, and none that use M-1s in the first stage. Certainly nothing at all like how you describe the FASA Nova was ever planned. That is what led me to be very confused by your proposal.

ctiberious:
Can you see any issues with changing the title of the stages from "FASANova" so something like "Saturn C-8"?

stratochief66:
I believe title changes are fine, not name changes though. Nova is a pretty
vague name, considering there were so many Nova proposals from the late
50s/early 60s through post-Apollo plans. Saturn C-8 seems an appropriate
title to use.

ctiberious:
Now the only thing I'm changing is to correct the engine nodes on the Stage One tank so they work properly, and updated title and description for the FASANova parts.  Now the Stage One tank, Stage Two tank, and Interestage are all listed as "Saturn C-8" and I've added to the descriptions so players can more easily understand what it is they are building.

stratochief66:
Thanks :)

PR #999, opened 2016-03-01
FASA Nova fixes, part 2

ctiberious:
I wanted to see if I could build the Nova rocket but noticed that RO didn't have a craft file for it. So I took the craft file that comes with FASA and used that as a template to build my own Nova rocket. During the assembly, though, I discovered that a few of the parts didn't appear as though they had ever been properly setup for RO.

The 2nd stage fuel tank doesn't have a mounting point for the decoupler that is supposed to be below it. It also has too many engine mounting points.
 The 1st stage fuel tank doesn't have engine mounting points setup correctly. Engines mount above the mount point instead of below.
 The fairing that separates the 1st and 2nd stage isn't long enough to allow the 2nd stage to include an M1 engine which appears to be what the FASA craft file uses.


ctiberious:
Consolidated this fix in with the other Nova fix on #998

PR #1000, opened 2016-03-01
Three ignitions for J-2

Schnobs:
Opening that can of worms again. Under the current system, three ignitions is less wrong than two.

See http://ksp.schnobs.de/stuff/notes.txt for the time being. I'm aware that there's supposed to be a proper solution someday, these notes were compiled for the issue/discussion about the would-be proper solution, but that fails to happen and I'm growing impatient.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/J2_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/J2_Config.cfg
@@ -56,7 +56,7 @@
 			
 			%ullage = True
 			%pressureFed = False
-			%ignitions = 2 //Definately a FIXME right here
+			%ignitions = 3 //Definately a FIXME right here
 			IGNITOR_RESOURCE
 			{
 				name = ElectricCharge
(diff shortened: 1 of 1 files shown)


A1Ch1:
Thanks for putting together those notes on the J-2 ignition sequence :)

One thing I noticed in the S-IVB section: the cold helium spheres didn't contain liquid helium, they contained gaseous He at the temperature of LH2. To elaborate a little:

Pre-pressurization = ground pressurization of tank for first burn
Pressurization = maintenance of tank pressure during burns
Re-pressurization = pressurizing tanks prior to restart

All flights had both cold He bottles and ambient temp He bottles. Initially (prior to AS-503), ambient temp He bottles were used for re-pressurization of the H2 and O2 tanks. Gas from the cold He bottles was heated by the J-2 during burns and used solely for pressurization of the O2 tank (the H2 tank was pressurized with gaseous H2, also heated by the J-2). After the introduction of the H2/O2 burner, cold He was also used for re-pressurization and the ambient He was relegated to backup. Thus all flights after the introduction of the H2/O2 burner carried enough He for three ignitions, as demonstrated by Apollo 9.

PR #1001, opened 2016-03-01
XLR81 (Agena) global config

A1Ch1:
* change name to military designation like other contemporary engines
* add sources
* adjust thrust (engines variants rated at 15 klbf, 16 klbf, and 17 klbf)
* rename configs to use their XLR designations where appropriate (unknown designation for Model 8096-39)
* update FASA Agena and VSR Agena to remove fields handled by master config
* remove ullage requirement from Agena D engines (which used propellant sumps):
agena ullage 2
* change XLR81-BA-13 (8247) to 15 total ignitions per sources
* add title, manufacturer, and description to global config
* add modulegimbal to global config

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Agena_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Agena_Config.cfg
@@ -11,17 +11,17 @@
 {
 	@title = XLR81 (Agena) Vacuum Engine
 	@manufacturer = Bell Aerosystems Company
-	@description = Gas-generator nitric acid/UDMH vacuum engine for the Agena upper stage / spacecraft. Uses storable propellants and starting with the 8081 model it can be restarted. The 8247 model was used for the Gemini Agena Target Vehicle flights (GATV was a modified Agena D). It is a modified 8096 (Agena D) engine with support for more than two restarts (rated for 14), but otherwise largely identical.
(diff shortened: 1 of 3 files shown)


NathanKell:
Awesome! While you're there can you verify that Agena D dry masses are correct now? I can't remember whether that was only talked about or fixed.

A1Ch1:
0.132 t (engine)
0.3 t (propellant tanks)
0.28 t (avionics)
0.005 t (antenna)
=0.717 t, so slightly heavy

The Flight Pack adds another 0.2 t, but it includes the SPS along with the RCS. Since the Agena D didn't come standard with the SPS the 600-odd kg weight isn't the one we want. I've got the mass of the SPS and associated hardware in one of these books, I'll look it up tomorrow. To model the basic Agena D it may be best to clone the Flight Pack but disable the SPS engines, leaving just the RCS, and then adjust part masses accordingly.

NathanKell:
Makes sense to me. Some stuff is already cloned for B anyway. Do you think you could have that done by mid tomorrow? I need to release a hotfix for the broken FASA_Titan file, but if possible I'd like to wait until those changes are made. :)

A1Ch1:
Yup that should't be any problem

NathanKell:
Awesome! I'll release once that's in, pending any other changes.

PR #1002, opened 2016-03-01
Remote tech range fixes

winged7:
Pioneer 10/11 - 20mln km seems too low for me, for now 1,5 bln km - same as in the description
Voyager - increased to 10bln km.


SirKeplan:
The descriptions says effective range, which is correct, the Pioneer 10/11 does indeed function out at Jupiter already, remmeber RO uses the Root RangeModelType, so to calculate the effective range you need to take into account the range of ground stations too. formula is min(r1,r2)+sqrt(r1*r2)

https://remotetechnologiesgroup.github.io/RemoteTech/guide/settings/

winged7:
ok, thanks. Closed.

PR #1005, opened 2016-03-01
RO Texture Replacer helmets

PhineasFreak:
* Add Earth as a viable body for suit/helmet removal.


NathanKell:
Should it be patching or should it be adding a new TextureReplacer node? I don't know how TR works.

PhineasFreak:
TextureReplacer has a unique loading system for mod or user configurations (better explained here: https://raw.githubusercontent.com/ducakar/TextureReplacer/master/GameData/TextureReplacer/@Default.cfg).

In short, parameters are not getting patched via MM. TR searches for available configs with the correct structure and applies each change separately.

PR #1016, opened 2016-03-03
SSTU RD-107/8 Fixes

SirKeplan:
Another problem with these engines is the thrust vectoring seems very weak, a rocket with a single RD-108 can't hold a straight course.  not sure what's the best resolution this now though.


NathanKell:
They're relying on features that won't be in KSP until 1.1, I think.

SirKeplan:
What features would that be?

Also, i've realised why the gimbal is so weak, the RD108_118_Config.cfg is setting the gimbalrange to 4.5, however that range that is being set is for the verniers, so it's making the thrust vectoring as a whole pretty pitiful.

NathanKell:
Just that: different thrust transforms having different outputs, and axis-limited gimbals.

PR #1017, opened 2016-03-04
RN scout craft updates

raidernick:
-Update scout crafts


raidernick:
@NathanKell this is good to merge

PR #1021, opened 2016-03-05
AIES weight/range/power brought in line with stock/RT antennas.

Schnobs:
These things look more modern that the standard set, so I've given them somewhat better data. If/when they get integrated with RP-0, they should be 70s tech and beyond.


NathanKell:
These should also be done on FOR[RealismOverhaul] not final.

NathanKell:
@Schnobs poke?

SirKeplan:
@Schnobs double poke/what's going on?  It'd be good to have these properly in RO.

SirKeplan:
now in #1108

PR #1023, opened 2016-03-05
Update Squad NERVA engine to reflect original 1970s NERVA I design.

SirKeplan:


stratochief66:
Good, I think.

The nuclear reactor relies on the hydrogen as a coolant, so I'm not so sure about the fully throttleable part. I would expect it would either be unthrottleable or have a very slow throttle time (which would be a cool game effect to see, since I believe the tools we have are capable of that) But of course we want to research that, not go with my gut :P

SirKeplan:
@stratochief66 ahh, i did read something about a slow throttle, i'll dig a little to see if i can find numbers.

SirKeplan:
Greatly slowed the throttle, however engine shutdown is still instant, is there a way to fix this? (do we even want to fix it?)

NathanKell:
Shutdown being instant is an RF...somewhere between design choice and bug. Not something that can be fixed cfg-side atm.

SirKeplan:
ok. well, it makes gameplay a lot easier i guess anyway :D

winged7:
I think we should keep two NERVAs: the original one from 1970s and the current one (925s isp, lower mass) which would be available much later in the RP-0 tech tree.

NathanKell:
I thought we had such a thing, from Atomic Age.

winged7:
Atomic age is not required by RP-0.

NathanKell:
Yes? Neither are many other partmods...

winged7:
I think we should give as many possibilities as possible in RP-0 with limited amount of mods.

SirKeplan:
hmmm, so you are suggesting we have a clone of the squad NERVA,i don't know, maybe. haven't seen a good source on the 925s isp one yet anyway.

NathanKell:
There are NTRs in SXT or VSR that could be configured as late-period NTRs.

Zarbizaure:
I will try to create config for later NTRs from SXT (at least Phoebus and Fat Mun, because they look cool) and will do a pull request soon. Do you have any reference for later nuclear engines?

SirKeplan:
no good references for late 80s/ 90s designs one yet, but alternatewars has good info on 2 other 1970s engines.
http://www.alternatewars.com/BBOW/Space_Engines/ROVER_NERVA_Engines.htm The NERVA II and the SNE(NERVA Alpha) might be good to have.

Zarbizaure:
I ook a look and the Phoebus looks like a Nerva. However, the "Fat Mun" looks like the actual Phoebus. Should we configure the FatMun to be a Phoebus counterpart?

SirKeplan:
I don't know, as far as i know know the Phoebus was never designed to be used in space and was just a development reactor, so it wouldn't make much sense to me.

Zarbizaure:
Of course. The NERVA II was supposed to be a Phoebus derivative. I don't exactly know which source to trust, so I'm going for the following:
Isp (vac): 850s
Thrust (vac): 906.7 kN (195 000 lbf)
Mass: 12.86 T (28.530 klb)

stratochief66:
The only modern NTR design I recall with a high ISP is the one for the constellation program, and found in RO_BahamutoD_Constellation here:
https://github.com/KSP-RO/RealismOverhaul/blob/fce4a2e2229b7d013f9e25399640e8c2be27ea89/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg

@winged7 as few mods as possible seem to be 'required' by RP-0 but more and more are becoming compatible & optional. With the 64 bit memory headroom we will gain with KSP 1.1 I expect more mods will be 'required' (not a game breaking, hard 'required' though @NathanKell is there/should the be a way in CKAN to distinguish between those two?) to at least flesh out the tech tree nodes through modern times.

@winged7 there seem to be a few NTRs already well configured in RO, and a passing view of tree.yml suggests at least 3 are priced and placed for RP-0. https://github.com/KSP-RO/RP-0/blob/74039f6f734726aaf6b10f433ce40ecd4b51c403/tree.yml#L2979

stratochief66:
@Zarbizaure the NERVA II appears to already be configured here by @PhineasFreak (although with no source cited I can find.
https://github.com/KSP-RO/RealismOverhaul/blame/4224f720fd8dafd1a652857a9a5d4515359b999f/GameData/RealismOverhaul/RO_SuggestedMods/CMES-WIP/RO_CMES_Propulsion.cfg#L2877

Zarbizaure:
They are a lot of various different sources for the NERVA II thrust level: somes around 66 kN, others with 907kN...I will configure one ("FatMun", because it looks like Phoebus 2A which was a NERVA II predecessor) with 907kN and one another, and smaller, maybe as an NTR Mars transfer engine (thrust change from 15 to 25 klbf, so between 69 and 116kN). http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf The config @PhineasFreak wrote seems close to that so I may "steal" his work.

SirKeplan:
@stratochief66 that's actually not a NERVA II, that would be a later development Bimodal NTR engine after the NERVA program was well over.

Zarbizaure:
Sorry if I'm making that much noise on this thread, but does anyone knows how to manage ModuleSurfaceFx inside a module engine? There is no plume and no gimbal on the NTR engines I configured. Should I start a pull request even if plumes and gimbal aren't currently working?

winged7:
@stratochief66  I know, I've placed and priced them.

PhineasFreak:
The CMES BNTR is the same design as the one from the Constellation Essentials and is a TRITON design (can be operated with or without LOX augmentation).

1. http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf (page 15)
2. http://alternatewars.com/BBOW/Space_Engines/AIAA-2004-3863_TRITON.pdf

NathanKell:
@Zarbizaure  noise is fine. :) For surface FX you can leave as is, but for plume you need to add a plume config in RO's plumes folder. Probably best is to make a copy of the Squad NTR's plume config and save-as it into the new correct folder (SXT) with the correct name for the new part you want, and update the partname in the patch. Then tweak the values to fit the new engine.

Zarbizaure:
@NathanKell Done :) As well as RP0 configs! I set the bimodal plume TRITON config (the one which use Hydrolox) as an hydrolox upper plume; I don't know how it really looks in reality but at least it looks differents from the classic Hydrogen NTR mode plume.

stratochief66:
@Zarbizaure if you look at the list of RealPlume Generics here, you see @Felger has made a few specific for NTRs. But any plume is better than no plume, so changes over to an NTR one can be done at a later time, even by a different persion. Just getting Any plume on it is good for now :)

https://github.com/KSP-RO/RealPlume/tree/master/GameData/RealPlume/000_Generic_Plumes

PR #1026, opened 2016-03-06
SXT NTR configs: NERVA 2 and Bimodal NTR proposal

Zarbizaure:
DO NOT MERGE NOW
Gimbal and plumes of both engines aren't working, could anyone help?
SXT engines seems to deal with SurfaceModuleFX, maybe there is something wrong with this.


SirKeplan:
I just opened the model files in blender, to get a gimbal you'll need `gimbalTransformName = thrustTransform`, however the model doesn't support any gimbal animation unfortunately.

Zarbizaure:
Thanks, I will test and see how it works.
Ah, i did not see that I forgot to rename this one.

Zarbizaure:
Apparently RealPlume doesn't like spacing typo, but I'm close to making work these plumes. I will update the pull request soon (and then place both engine in RP0).

Zarbizaure:
I did some messes with commits but it should be fine now, ready to be merged.

PR #1033, opened 2016-03-08
Added TestFlight configs for WAC-Corporal, XASR-1, and AJ10-27

jwvanderbeck:


jwvanderbeck:
Did I do right?

NathanKell:
The cycle reliability is the reliability for the whole cycle. Since the WAC apparently had a reliability of 88%, I would suggest numbers more like:
igniteStart: 0.9
igniteEnd: 0.97
cycleStart: 0.86
cycleEnd: 0.93
That should lead to 88% reliability nearer the end than the start, with final reliability slightly better (90.2%) than was achieved in toto.

Per our discussion in IRC, and per the notes in the example config, the cycle reliabilities are for completing the entire cycle, so 1 - (1 - ignition) - (1 - cycle) should = final reliability for a launch.

PR #1038, opened 2016-03-08
RSB Engines additions

PhineasFreak:
* Add RO support for the Castor 120 SRM and the XLR81.
* Normalize crash tolerances between SRMs and liquid engines.


NathanKell:
I think you mean HTPB, but it's just in a comment so it doesn't really matter. :)

NathanKell:
Also, why the crash tolerance increases? 12m/s is already a pretty high tolerance

PhineasFreak:
I am still trying to find some values that are sane for each part type (engine, electrical, solar, structural etc). I will probably adjust them all at once since RO RSB is still missing some parts.

And time to fix that comment... :P

PR #1048, opened 2016-03-09
SXT Airbags

Zarbizaure:
Not a lot of change here, just mark these two airbags as "RO-supported". I did not change the scale or anything because I think it's better to do it with tweakscale (on these parts).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Utility.cfg
@@ -832,11 +832,13 @@
 @PART[SXTAirbag]:FOR[RealismOverhaul]
 {
     %RSSROConfig = True
+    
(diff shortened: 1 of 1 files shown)


NathanKell:
Per our discussion on the other PR, please change the description to be RO-appopriate rather than kerbalish. Otherwise looks good! :)

Zarbizaure:
Well, I kept them because they are fun! But that's fine, I will change them to something more realistic :)

NathanKell:
RO does not believe in fun. It has surgically removed it from KSP.
...
...
:P

Zarbizaure:
Trust me or not, but i find RO really fun to play (and also really rewarding...). Watching rocket explode with even more thrust and boosters is always fun.

NathanKell:
Hence the :P -- was poking a bit of fun at the stereotype that RO sucks the fun out of KSP. No, it's that RO's fun is a different kind of fun. :)

PR #1050, opened 2016-03-10
Big Gemini

ctiberious:
Update and setup various parts for the Big Gemini spacecraft.  Changes include:
1) Standard Big Gemini crew module crew capacity increased to 7.  This allows a 9-man Big Gemini setup (including two man cabin).
2) Rescue (White) Big Gemini crew module crew capacity increased to 10.  This allows 12-man Big Gemini setup (including two man cabin).
3) Both Big Gemini crew modules have the same tank capacity.  The white module uses all it's capacity for life support.  Since the black module allows for fewer crew, it has some extra tank space available which represents this module having some additional storage space.
4) Big Gemini Retro Module has all life support and equipment removed.  It is not setup just as the retro pack, just like the standard retro module for the normal Gemini.  I did include the ConnectedLivingSpace module, though, so that crew can pass through the retro module to make use of the rear docking port.
5) Big Gemini Service Module graphics have been realigned and oriented.  You no longer have to flip the part to set it up for use.
6) BGSM no longer includes science or crew components.
7) BGSM Storage capacity (i.e., Volume) has been increased significantly to account for the 1310 cu ft (37095.07L) of cargo space the module appears to have been capable of(see illustration at http://www.astronautix.com/graphics/b/bigcutaw.jpg)
8) BGSM includes ConnectedLivingSpace module so crew can make use or rear docking port
9) It looks like the BGSM as originally added to RO may have been intended to include built in engines but they didn't work.  I've removed those modules and added in the RCS module to allow for forward orbital translation to occur.
10) Big Gemini Docking Ring has been resized and adjusted similar to the FASAApollo_DockingDevice.  It also includes a ConnectedLivingSpace module.
11) Big Gemini Decoupler has been correctly sized for the 6.6m base.
12) I've cleaned up and balanced out the various Big Gemini life support and fuel cell modules so they properly maintain mass.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -300,7 +300,7 @@
 		{
 			name = Waste
 			amount = 0
-			maxAmount = 162
+			maxAmount = 166
 		}
 		TANK
 		{
@@ -315,7 +315,7 @@
 		converterName = CO2 Scrubber
 		conversionRate = 3.0	// # of people - Figures based on per/person
 		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010, LithiumHydroxide, 0.00001189
-		outputResources = Waste, 0.00003847, false
+		outputResources = Waste, 0.00003932, false
 	}
 	@MODULE[ModuleRCS]
 	{
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg
@@ -121,8 +121,9 @@
 	@node_stack_top = 0.0, 1.058411, 0.0, 0.0, 1.0, 0.0, 2
 	@node_stack_bottom = 0.0, -1.176027, 0.0, 0.0, -1.0, 0.0, 4
 	@title = Big Gemini Passenger Compartment
-	@description = This is the lower half of the Big Gemini re-entry module, the passenger compartment.
+	@description = This is the lower half of the Big Gemini re-entry module, the passenger compartment.  This module gives the Big G a max 9-man crew capacity with additional cargo space.
 	@mass = 2.02456
+	@CrewCapacity = 7
 	CoMOffset = 0.0, -1.5, 0.0
 	!RESOURCE[MonoPropellant]
 	{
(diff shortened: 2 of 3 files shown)


ctiberious:
I've adjusted the built in life support supplies on the Gemini to account for a max 14 day flight.  I also rebalanced the CO2 Scrubber so it processes CO2 at the same rate as Apollo while correclty maintaining mass.
I removed the tank and LS changes that the Rescue (White) pod had since the pod itself was essentially the same as the normal pod.
I altered the fuel cell and LOX-O2 generator on the utility pack so that they maintain proper mass.  Also increased the available volume to account for these alterations.

ctiberious:
While verifying that mass was being properly maintained on the Gemini parts, I noticed an issue with the Apollo CO2 Scrubber.  I've corrected that and slightly increased the available waste storage on the Apollo CM to account for the discrepancy.

ctiberious:
Think I only used "x-man" in two places (the two Big G crew modules) so those are both fixed now.

Yes, the FASAGeminiBigGDock is not setup to only be passable.  You can't put crew into it.  The only reason it has any Oxygen supply at all is for the LOX-O2 converter to function correctly.

Only thing that may still need to be adjusted is the strength of the RCS thrusters on the FASABigGeminiRetroModule.  Those come directly from the FASAGeminiUtilitySasRcs which is designed to handle maneuvering for the 3.5t Gemini.  They may be under powered for the 15-50t Big G but I wanted to run a few more tests before tweeking those values.

stratochief66:
Sounds fair. So we shouldn't merge this yet? Please leave a message here
when you're fine with it being merged

ctiberious:
I'm changing the texture used for the Big G Retro Module.  The original texture didn't handled the 3.9m to 4.6m taper as well as the texture I'm using now.  I also read (http://www.astronautix.com/craft/bigemini.htm) that the retro module only included the retro engines, plus water and oxygen.  It didn't include any attitude control thrusters like the comparable module on the standard Gemini setup.  I've therefore removed the RCS module from this part and I've added a 2nd "fuel tank" setup specifically for life support to store the bulk of the flights water supply and backup O2.  This module includes textures for four "Gemini Mini SRB" to act as the nozels for the retro engines.  About the only real issue with this texture (that I can see) is that it is no longer "hollow".

I'm changing the texture used on the Big G Service Module so that we now only use the FASAGeminiBigGDock graphic.  I also added four Apollo RCS thruster textures since the same page I cited above says the service module includes the attitude control system.  The illustrations I've found seem to agree.  I've also removed the bulk of the water from this module (moved to the retro module) and reduced the storage volume accordingly.  The ModuleRCS thrust on this part has been doubled as compared to the RCS on a standard Gemini.  The tests I've run so far seem to work well with this thrust setting.

Lastly, I've altered the size of the Big Gemini Docking Ring to more closely match the dimensions I'm seeing on various illustrations (see above website).

At this point, I think this PR should be good to go.

PR #1052, opened 2016-03-12
Updates for SSTU 0.3.29.11 - SRBs

shadowmage45:
## Features
* Add first pass of modular SRB parts (3 variants)
  * Full part config files custom tailored for specific lines of SRB models
  * Including thrust differentiation for different lengths/segments, burn time, and thrust curves
  * RSRM (1-6 segments, with diameter scaling)
  * UA120 (1-8 segments, with diameter scaling),
  * GEM-40/46/60/xx (multiple diameters and lengths).
  * Additional body/nozzle/nosecone variants exist that are as-yet unused; more SRB model-lines may be created in the future (feel free to use these as examples)
* Disable original SSTU-MSRB parts (not configured for RO, might as well remove the configs)
* Add replacement stand-alone SRB parts as-per stratochief's request (not needed though with the modular ones working finally); feel free to remove these if they are deemed unnecessary

## Known Issues
* Thrust curves for many segments are copied from the closest available;
  * GEM uses the same curve for all lengths / diameters
  * UA120 uses the 1207 curve for all lengths / diameters
* Diameter scaling is enabled with thrust scaled for constant burn-time (cubic)
  * Might want to flip this around for square scaling for both thrust and resources
  * These are controlled by the thrustScalePower and resourceScalePower options in the MSRB module; change both to 2 to use square scaling
* Effects scaling and positioning is WIP - needs a plugin side fix that should be available soon.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_EnginesBasePatch.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_EnginesBasePatch.cfg
@@ -102,3 +102,8 @@
 //remove the placeholder engines
 -PART[SSTU-SC-ENG-Merlin-1B]{}
 -PART[SSTU-SC-ENG-H1]{}
+//remove original MSRB parts
+-PART[SSTU-SC-ENG-SRB-A]{}
+-PART[SSTU-SC-ENG-SRB-B]{}
+-PART[SSTU-SC-ENG-SRB-C]{}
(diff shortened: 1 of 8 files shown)


stratochief66:
Thanks again @shadowmage45!

I tested through the different RSRMs and they worked very nice. I found a couple of small typos that I will patch after merging.

jbengtson:
These patches define parts without declaring a requirement for SSTU or Realism Overhaul itself.  They will cause errors if SSTU is not installed.  Please correct.

PR #1054, opened 2016-03-13
Fix node position on service bays

SirKeplan:
The nodes were outside the part.

the 4m bay actually looks a fraction too large for 4m, should it be scaled down a tiny amount too?


NathanKell:
Uh....that will cause the rescale bug. They use MODEL nodes, so you can't have non-1.0 rescaleFactor.

NathanKell:
@SirKeplan did you have a chance to test? IIRC you were going to, but i may have missed the result in IRC.

PR #1055, opened 2016-03-13
Rd-103-107-techtransfer

stratochief66:
Just a suggestion, as I don't know well balanced values for this. @NathanKell might weigh in with some wisdom, or @raidernick.

It might also make sense to get a small tech transfer from V-2, maybe 5 or 10%?

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD107_RD117_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD107_RD117_Config.cfg
@@ -340,6 +340,7 @@
 		ignitionReliabilityEnd = 0.96
 		cycleReliabilityStart = 0.85
 		cycleReliabilityEnd = 0.95
+		techTransfer = RD-103,RD-103M:50
 	}
 }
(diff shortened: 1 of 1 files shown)


NathanKell:
50% transfer seems quite high for an engine that did not have that much in common with the RD-107/8 except for the HTP-powered turbopump. I mean, the RD-103 even had a spherical combustion chamber still.
That said, I agree some transfer makes sense, something like 10%? 20% tops maybe?

stratochief66:
@NathanKell alright, I will go with 20% for now maybe. I think the starting reliability needs to be calibrated as well. I plan to spend some quality time with http://space.skyrocket.de/directories/launcher_ussr_russia.htm to educate myself a bit about early RD-107 & RD-108 usage.

@NathanKell could you apply a transfer for the V-2 to the RD-103 then? sounds like they are far more similar than I had thought, and they current have -no- tech transfer.

NathanKell:
Yep, will do. V-2 to X-405 is 10%, so 20% for the RD-103 makes sense.

Also, IRCing you an idea.

PR #1058, opened 2016-03-14
Update RL10 Config

A1Ch1:
numbered sources
edit description
adjust mixture ratios
adjust CECE throttle to average of target
add inline citations
adjust TF burn time for RL10C-1 and CECE
adjust TF reliability for CECE


NathanKell:
Given the very high rated burn times I would suggest using a data rate multiplier for those configs so it doesn't take hours and hours of burn time to get data. Otherwise looks good!

A1Ch1:
I'm not very familiar with TF configs, but the burn time is about 9x longer than the RL10B-2, would a rate multiplier of 5 deliver data at a reasonable rate? The engine is designed to deliver extremely high reliability from the start since CECE was intended to lead to an engine for crewed applications like lunar landers where a failure would likely be non-survivable, so I think that improving on the reliability further should take more than the typical amount of data.

NathanKell:
That seems fine to me, yeah.

SirKeplan:
I'd be tempted to go out and give the CECE-Base 17.6:1  throttling, as that is what it demonstrated here http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20100032918.pdf

A1Ch1:
Well, remember that what was demonstrated was a throttling on a modified RL10A-4-2

A1Ch1:
Whoops my mobile cut out most of my reply. I'll fix that when I'm back at the laptop.

SirKeplan:
poking @A1Ch1

NathanKell:
@A1Ch1 is there anything more coming? Or should I merge this as it stands?

stratochief66:
It looks good to merge. Some things might get tweaked later of course, Test Flight configs will probably need some refinement after we've played with them.

A1Ch1:
Sorry, my phone decided to die a few days ago so I've got no internet on my laptop except public wifi until I'm back in town.

NathanKell:
Youch! Best of luck!

PR #1059, opened 2016-03-14
SRB name error fixes

stratochief66:
2 SRBs were broken due to errors in the names.

An ISP was off by 9 in one instance.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_RO_SRB2.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_RO_SRB2.cfg
@@ -194,7 +194,7 @@ MODULE
 		atmosphereCurve
 		{
 			key = 0 276
-			key = 1 240
+			key = 1 249
 		}
 		curveResource = PBAN
 		thrustCurve
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_RO_SRB4.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_RO_SRB4.cfg
@@ -1,7 +1,7 @@
 PART
 {
 module = Part
-name = SSTU-RO-ENG-SRB3
+name = SSTU-RO-ENG-SRB4
 author = Shadowmage
 
 TechRequired = veryHeavyRocketry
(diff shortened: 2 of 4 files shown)


stratochief66:
There also seems to be a bug where the roll control engine on the RS-68 won't change off of burning Oxidizer.

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_RS-68.cfg#L56-L60

It just doesn't seem to take effect for some odd reason.

NathanKell:
What's the original cfg for it? What does configcache show for it?

stratochief66:
Original:
https://github.com/shadowmage45/SSTULabs/blob/master/GameData/SSTU/Parts/ShipCore/engines/SC-ENG-RS-68-New.cfg

My config cache:
http://pastebin.com/Ff9DvwyA

My current attempt to patch it:
http://pastebin.com/1ycXL3Hx

The current patch for it in Github:
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_RS-68.cfg#L2-L154

One of the other MM patches might be hitting it as well, but I don't think so.

SirKeplan:
@stratochief66 this is a wierd issue, noticed the same ingame. i have  not seen any obvious fix yet.

NathanKell:
@stratochief66 ok so the issue is in the patch somewhere since the configcache shows the wrong result.
My suggestion:
After the engine patches, do:
``@MODULE,*:HAS[@PROPELLANT[*]]
{
    !PROPELLANT,* {}
    PROPELLANT
    {
        name = LqdOxygen
        ratio = foo
    }
    PROPELLANT
    {
        name = LqdHydrogen
        ratio = bar
    }
}

SirKeplan:
@stratochief66 i haven't tested this ingame yet, but i'm pretty sure i know what is causing the problem.

SSTU/Patches/LH2-Engines-New.cfg  in the SSTU mod is probably causing the issue.

SirKeplan:
Ok, so deleting SSTU/Patches/LH2-Engines-New.cfg  solves the issue, it's not an issue with RO.
The problem is likely that LH2-Engines-New.cfg is getting run after RealismOverhaul, and setting the propellants. If it ran before it shouldn't be a problem.

tagging @shadowmage45

shadowmage45:
Hmmm....
Been meaning to roll those stats into the base configs.  As I include CRP with the distribution, there is no need for those to be a separate patch.
Consider it fixed; will be available with the next SSTU update.

Might be similar problems with some of the other patches;  if you run into any, let me know and I'll throw a RO exclusion into the NEEDS block.

PR #1060, opened 2016-03-14
RSB 0.8 Updates

PhineasFreak:
* Remove conflicting craft files.
* Add RO support for the new Delta III and Saturn I parts.
* Various other fixes.


stratochief66:
are the stock craft files compatible with RO? if not, it might make sense to rename your craft to avoid the conflict rather than deleting them.

PhineasFreak:
Some parts have been rescaled so not all crafts are compatible. I will check what crafts need to be updated and rebuilt them.

PR #1067, opened 2016-03-16
SSTU minor fixes

FFCJoseEduardo:
just added SSTU payload base, Soyuz booster tank and fixed Orion top node position

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Misc.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Misc.cfg
@@ -10,7 +10,7 @@
 	%RSSROConfig = True
 	@title = NASA Docking System
 	%manufacturer = Boeing
-	@description = The NASA Docking System (NDS) is a spacecraft docking and berthing mechanism being developed for future US human spaceflight vehicles, such as the Orion Multi-Purpose Crew Vehicle and the Commercial Crew vehicles. The NDS is NASA’s implementation of the International Docking System Standard (IDSS), an attempt by the ISS Multilateral Coordination Board (MCB) to create an international spacecraft docking standard. NDS is also known as the international Low Impact Docking System (iLIDS). Parachutes NOT included.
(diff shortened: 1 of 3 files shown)


NathanKell:
@stratochief66 ? Your department, I don't know much here :]

stratochief66:
@FFCJoseEduardo you break the unicode/character encoding here.:
https://github.com/KSP-RO/RealismOverhaul/pull/1067/files#diff-842b2557e40947501aed5fb16806a89bR13

Increasing the scalability by 2x at all tech nodes looks very arbitrary, and also possibly not meaningful since the SSTU parts aren't placed and priced in RP-0 as far as I'm aware. I'm a bit curious if/how that syntax would even work, but given the other things I'm not sure in what context this would improve anything.

I thought the SSTU Orion node had already been adjust recently, but looking back through BLAME I don't see it. I had thought someone had complained about the escape cover sitting too low and a PR changing that within the last month.

Theysen:
> I thought the SSTU Orion node had already been adjust recently, but looking back through BLAME I don't see it. I had thought someone had complained about the escape cover sitting too low and a PR changing that within the last month.<

That was me , though I only reported it in IRC and iirc, @stratochief66 fixed the issue in a release after I reported it. The LES assembly works fine and has correct positioning in current version.

stratochief66:
@Theysen  Thanks for filling in that gap in my memory :)

FFCJoseEduardo:
which release? because when I downloaded the githib repo version it was sitting low and messing with the docking, now it matches exactly where it should, it seemed like it was attaching the LES correctly, but it was a bit lower than it should

also, this change wasn't for RP-0, but for stock career, diameters over 10m aren't available in career without that, and since I didn't want to balance the nodes I multiplied by 2 the radius, otherwise you would only have up to 2.5m tanks until near the end of the tech tree (which would be fine in stock), the node that unlocks 3.75m tanks in stock is the same for SSTU, so even with this change you can't make a shuttle ET without the very last rocketry node

again, this change was for stock career, not RP-0, but if anyone wants to step in and do a RP-0 config, their location is already shown

as for breaking that, I have no idea how that happened because I didn't even touch the description... however that should be fixed now

stratochief66:
How could the LES mess with the docking port? I mean, you jettison the LES before you use the docking port. I just attached the Orion LES to the Orion and it looks to sit just fine. The curve of the LES and base appear to blend perfectly.

http://i.imgur.com/wBRaeE5.png

There isn't really a 'stock career' with RO. The RO career mode -is- RP-0. I think I see what you mean, like if you want to play in a larger solar system without RP-0, but wouldn't that also be without RO?

Yeah, the broken encoding is odd, thanks for fixing it.

but given that the LES appears good to me and the other changes seem confusing or possibly unnecessary, I'm going to close this PR.

PR #1071, opened 2016-03-16
Global Configs for Vens Stock Revamp.

SirKeplan:


jwvanderbeck:
Would be nice to get the TestFlight configs in at the same time when adding new Engine Configs.  Just so that we stay on top of them and don't end up with a huge backlog.

NathanKell:
Oh, I agree. But we already have a huge backlog as it is, and progress is progress... :]

PR #1074, opened 2016-03-17
RSB 0.8.1 Updates [Part II]

PhineasFreak:
* Add RO support for the PSLV parts.
* Add RO support for the LR101 & Vikas engines.
* Balance the masses of all procedural fairing base to the equivalent PF
ones.
* Improve the inert mass values of all parts.
* Fix some typos.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ares_I.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ares_I.cfg
@@ -13,7 +13,7 @@
 //  Ares I payload fairing base.
 
 //  Dimensions: 5.5 m x 1.2 m
-//  Inert Mass: 150 Kg
+//  Inert Mass: 1200 Kg
 //  ==================================================
 
 @PART[RSBfairingAresI5m]:FOR[RealismOverhaul]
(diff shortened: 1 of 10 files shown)


stratochief66:
Looks good to me, aside from the need for ingame testing. @PhineasFreak could you please provide some craft files to test the new craft you're supporting here?

PhineasFreak:
Sure, no problem! It will be also compatible with the base RSB craft files (so no more conficts).

stratochief66:
I missed this comment, thanks again! :)

stratochief66:
The second stage tank appeared to have the wrong balance of UDMH/NTO for the engine, although I could be wrong there. I am very unfamiliar with this vehicle but on my first test it appeared to run out of one fuel early.

The ground level thrust appeared to be very high!

I'm not sure why the core stage has an RCS system, given how high the aerodynamic forces are during most of the first stage burn. I suspect the engines on the boost and/or core stage should have less gimbal, since despite the crazy high dynamic pressure I didn't have any trouble keeping the craft from tumbling and being ripped apart.

PhineasFreak:
@stratochief66 sorry for the late reply!

I had used a different version of the Vikas (i think it was the 1+) initially. Then, i found out that it currently uses the 2 version and forgot to update the O/F ratio. Fixed.

Yes, the thrust is really high! It gets off the pad like a bat (and to imagine the craft was using the small strap-on SRMs...)

The core stage is designed to fly without boosters (CA - "Core Alone") so it needs these thrusters for roll control. With boosters that is no longer an issue (they provide additional control authority).

stratochief66:
Excellent. So I only noticed one oddity then, the fuel mixture thing.

And no worries about the 'late reply', it looked good so we just threw it
into the most recent release as it looked 99% good and we trust you to fix
the 1% when you can :)

PR #1075, opened 2016-03-17
J2 restarting

ctiberious:
I know there is some complaint about handling restart capability on the J2 engine.  S-IVB-200 (used on Saturn 1B) had no restart capability.  S-IVB-500 (used on Saturn V) could be restarted up to 3 times, though it was never started more than twice.  According to the J-2 technical specifications, the built in auto-starter was effectively unlimited so the J-2 could really be started as many times as you had fuel and proper pressurization.  Maybe there is a better way to handle restarting the J-2 but this is a workable solution using the existing systems.
I've reset the "ignitions" value to 99.  For some reason I can't set "ignitions = -1" to get unlimited ignitions so I've opted for 99 since it's a high enough limit to qualify as "unlimited" for most purposes.
I've added an "IGNITOR_RESOURCE" of "Helium" which will limit the number of restarts you can have on a flight.  I set the amount of Helium required for each restart to 250L.  I came up with that figure because the S-IVB-500 included eight 3.5 cu ft (approx. 99L) "cold helium spheres" inside the LH2 tank.  That meant the S-IVB-500 included approx. 792L of Helium for fuel pressurization which was enough for up to 3 ignitions.  That means each ignition used approx. 264L.  I rounded off at 250 for ease and because I can't be sure that the Helium wasn't used for other purposes.
Finally, I've added 250L Helium to the J-2 engine itself.  This allows a single ignition version of the engine to be used as is.  No need to change existing vehicle configurations unless you need more than a single ignition.  If you need more, just add 250L/restart/engine of Helium to an existing tank.

I will add updated configs for the J-2X, J-2T and M-1 momentarily (they'll be on the same PR).

The one problem I've come across with this setup is that Default tanks can't currently hold Helium.  Only ServiceModule and Fuselage tanks currently have the ability to store Helium.  For the FASA S-IVB tank, I'm going to add a 500L Helium capacity (with current amount set to 0) so that both S-IVB-200 and S-IVB-500 versions can be modeled.  I'm wondering, though, if RealFuels should be updated so that Helium is a valid fuel for other tank types.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/J2T_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/J2T_Config.cfg
@@ -35,11 +35,16 @@
 				key = 1 300
 			}
 			ullage = True
-			ignitions = 3
+			ignitions = 99
 			IGNITOR_RESOURCE
 			{
 				name = ElectricCharge
 				amount = 2.0
+			}
+			IGNITOR_RESOURCE
+			{
+				name = Helium
+				amount = 250
 			}
 			cost = 0
 		}
@@ -65,11 +70,16 @@
 				key = 1 290
 			}
 			ullage = True
-			ignitions = 3
+			ignitions = 99
 			IGNITOR_RESOURCE
 			{
 				name = ElectricCharge
 				amount = 2.0
+			}
+			IGNITOR_RESOURCE
+			{
+				name = Helium
+				amount = 250
 			}
 			
 			cost = 1000
@@ -83,5 +93,11 @@
 	}
 	!MODULE[ModuleAlternator]
 	{
+	}
+	RESOURCE
+	{
+		name = Helium
+		amount = 250
+		maxAmount = 250
 	}
 }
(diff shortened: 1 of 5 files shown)


SirKeplan:
For `ModuleEngineConfigs` i believe you set ignitions to `0` if you want infinite.

I like this kind of idea, but i think something similar was rejected in the past. tagging @NathanKell because i won't get the specifics right.

ctiberious:
@SirKeplan Hmm.  I didn't try "ignitions = 0" because I noticed some other unlimited restart engines (i.e. Atris) were using "ignitions = -1".  But I'll give 0 a try.

NathanKell:
0 is for inside a CONFIG. -1 is for in a ModuleEnginesRF.

ctiberious:
Gotcha.  I can definitely update these to use "ignitions = 0" then.  That allows the J-2 and it's various variants and offshoots a restart limited controlled completely by the amount of Helium you bring with you.

NathanKell:
Helium will not be added to Default tank type; it's a resource that needs to be highly pressurized, so it won't work in a low-pressure tank.

As to the meat of the PR itself. I am going to close it for now, because this needs to be part of a larger rework to how ignitions are handled. We need changes TestFlight side as well. I'll look at doing this on a larger basis during the 1.1 update cycle, when we'll have to rejigger a bunch (and when save-breaking will be more acceptable--recall that your change will make in-flight J-2s unignitable).

PR #1083, opened 2016-03-21
Remove stray tab breaking LR105-NA-7.1 propellants

ec429:
Caused it to run off 100% kerosene(!)


SirKeplan:
thanks for catching this, must have snook past me.

PR #1085, opened 2016-03-22
Update SSTU_SLS_Upper_Stages.cfg

Max-Q:
Both ICPS and EUS fixed to work with SSTU version SSTU-0.3.30-pre2


stratochief66:
Thanks again!

Tested in game, both EUS and ICPS look much better new. 

For the historical record, it is good to modify the name of the Pull Request or of a commit to describe what the issue is, and what the fix is, not just "Update SSTU_SLS_Upper_Stages.cfg"

As the EUS and ICPS model in stock SSTU had been adjusted, and more parts were added like RCS ports, these fixes place all sub-models into reasonable and right looking places.

@Max-Q any idea of what bit of SSTU-RO configs you want to improve next? :)

PR #1086, opened 2016-03-22
SXT PPD-8 Lander Can

PhineasFreak:
* Add RO support for the PPD-8 lander can. Stats similar to the Mk2 Lander Can.


stratochief66:
Looks fine to me, I haven't tested it in game though.

@PhineasFreak as far as I know, you are fine to include TACLS modules without the MM conditional. If someone doesn't have TACLS installed these just fall on deaf ears and don't throw any exceptions, as far as I'm aware.

https://github.com/KSP-RO/RealismOverhaul/pull/1086/files#diff-20764d5ab23f171354174f6ea2325669R573

PhineasFreak:
I know that KSP does not have a problem if a module does not exist but it spams the logs with useless error messages, making it harder to separate and review real errors.

PR #1088, opened 2016-03-22
Tantares Venera Lander fixes

SirKeplan:
fixed scaling.

also removed the reaction wheel as it seemed unlikely that it would have one. (if this is incorrect i'll put it back)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Tantares/RO_Tantares_Venera.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Tantares/RO_Tantares_Venera.cfg
@@ -3,8 +3,12 @@
 @PART[Fomalhaut_Control_A]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	%rescaleFactor = 2.05
-	%scale = 0.4878
+	@MODEL
+	{
+		%scale = 2.05, 2.05, 2.05
+	}
+	@node_stack_bottom = 0.0, -0.74293025, 0.0, 0.0, -1.0, 0.0, 2
+	@node_stack_top    = 0.0,  0.74293025, 0.0, 0.0, 1.0, 0.0, 1
 	%mass = 0.56
 	
 	%title = Venera Lander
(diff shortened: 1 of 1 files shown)


NathanKell:
@Niemand303 @raidernick maybe you guys know about the reaction wheel?

stratochief66:
Unrelated (apologies for butting an RP-0 question in here) but any ideas what a fair price would be for this part in RP-0?

https://github.com/KSP-RO/RP-0/blob/42239e845642e2f944797eef77664e37774bd2e8/tree.yml#L2144-L2145

stratochief66:
@SirKeplan Unless I missed something, you just deleted the modifications to the Reaction Wheel, not the wheel itself.

https://github.com/KSP-RO/RealismOverhaul/pull/1088/files#diff-341de1458de3f39573edd3db567f6345L22

you want to use something like:

!MODULE[ModuleReactionWheel] {}

to remove it entirely

SirKeplan:
@stratochief66 haha, woops!

stratochief66:
@NathanKell  should this line have an @ or %?
https://github.com/KSP-RO/RealismOverhaul/pull/1088/commits/49e061398e6029b087d4c6b090c00762e0c7be7a#diff-341de1458de3f39573edd3db567f6345R35

Beyond that, I think the PR is a great step forward for the part :)

NathanKell:
% because the node probably doesn't already exist. Might use a % for gTolerance as well just to be safe, but I doubt anything else is setting it.

stratochief66:
@SirKeplan 

Once you add the above, good to merge I think. There are some other things I want to add to this part as well I believe, I like the scientific instruments it had in-built. (That I can do in later PR)

SirKeplan:
@stratochief66 AFAIK this should be good to go.

stratochief66:
Thanks @SirKeplan. In your research for this, do you know roughly what series of Venera had a lander that looked like this?

SirKeplan:
Venera 9 and onwards it looks like.

PR #1089, opened 2016-03-22
SLA/IU Collider workarounds

ctiberious:
Based on numerous tests I've run, I'm 99.9% sure that the collider problem is on the IU texture itself.  I tried creating a single part that was made up of the two "FairingWall3m" of the standards SLA and the "Apollo_IU".  But when the LEM separates from the SLA, it is still picking up phantom momentum as it's "pushed" out of a texture. We definitely someone that can look at the collider on the texture itself and fix it.  But in the mean time I've got an update to the standard SLA so you no longer need a separate decoupler for the LEM.  Those changes include:
1) I found some documentation that puts the SLA at "7-foot long" which works out to 2.1336m.  The existing part is 2.07m.  This doesn't seem like much, but every mm reduces the amount of phantom momentum when the LEM separates.  So I've corrected the height of the basic SLA to be 2.1336m long.
2) I've slightly adjusted the fairing, top and bottom attachment points to account for the tiny increase in size and also to remove any clipping the fairing may have with the CSM.
3) I've raised the LEM attachment point to account for no longer needing an extra decoupler.

I've also added two new interim parts.  The first is a 2.6m long SLA.  It's not drastically larger than the standard part (about +22%) but does cut the phantom momentum by about half.  And there is no phantom momentum if you build your own LM so nothing is below the descent engine.
The second is a 3m long SLA.  It's large enough that the normal LM doesn't hang below it resulting in no phantom momentum upon LEM separation.

I also reduced the size of the attachment points on the SLA and IU so it's easier to find and place everything.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
@@ -18,16 +18,16 @@
 	@mass = 0.2845
 	@MODULE[ModuleDecouple]
 	{
-		@ejectionForce = 0.001
+		@ejectionForce = 0.2
 	}
 	@MODULE[ModuleEngines*]
 	{
-		@maxThrust = 0.01
+		@maxThrust = 0.3
 	}
 	@RESOURCE[SolidFuel]
 	{
-		@amount = .003
-		@maxAmount = .003
+		@amount = .10
+		@maxAmount = .10
 	}
 }
 +PART[FASAStrFairing3m4x]:AFTER[RealismOverhaul] // probably a better way to do it than to overwrite a random part
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
@@ -38,31 +38,32 @@
 	}
 	!MODEL,0{}
 	!MODEL,1{}
+	//Both https://en.wikipedia.org/wiki/Apollo_(spacecraft) and http://www.apollosaturn.com/asnr/SLA.htm list the length of the SLA as 7 feet which would be 2.1336m
 	MODEL
 	{
 		model = FASA/Gemini2/FASA_Gemini_LR91_Pack/FairingWall3m
-		scale = 1.761, 0.69, 1.761
-		position = 0.0, 1.492, 0.0
+		scale = 1.761, 0.7112, 1.761
+		position = 0.0, 1.0668, 0.0
 		rotation = 0, 0, 0
 	}
 	MODEL
 	{
 		model = FASA/Gemini2/FASA_Gemini_LR91_Pack/FairingWall3m
-		scale = 1.761, 0.69, 1.761
-		position = 0.0, 1.492, 0.0
+		scale = 1.761, 0.7112, 1.761
+		position = 0.0, 1.0668, 0.0
 		rotation = 0, 180, 0
 	}
-	%node_stack_top = 0.0, 6.081, 0.0, 0.0, 1.0, 0.0, 4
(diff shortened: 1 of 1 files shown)


stratochief66:
@ctiberious could you post a link to the documentation? We like to keep those around for posterity :)

Also, thanks. I will need to test this in game of course.

ctiberious:
Woops.  I put the links in the file changes themselves but forgot to include them here.
https://en.wikipedia.org/wiki/Apollo_(spacecraft) states, "The SLA was composed of four fixed 7-foot-tall (2.1 m) panels".
http://www.apollosaturn.com/asnr/SLA.htm states, "the lower or aft panels, about 7 feet long".
Since 7ft = 2.1336m, that's what I set the standard SLA to.

stratochief66:
Tested good; I tried out the standard base (slightly longer now from your research source) and the long one that avoids the IU collider.

I noticed the SLA panels no longer move away on staging (not due to anything you did). Did you notice that as well? They still contain a little solid fuel, and it is consumed, but they don't really move away. I would consider roughly 10x more push off on staging appropriate and plan to look into that after this PR is merged.

ctiberious:
Yes, I did notice.  I noticed in a number of my tests that when the panels get jettisoned (and the LEM is "inside" the IU) one panel would sometime fly away while the other panels might just move a little or might even seem to come towards the stack.  I'm guessing that the ejection force (0.001) and the thrust (0.01) may have been turned down to try and "correct" this.
I think the whole stack is imperceptibly moving because of the clipping that's taking place so when the panels are detached, they are acting weird.  And once you remove the clipping by using the long SLA, the panels only get the momentum imparted by the ejection and thrust they are supposed to have.  I can increase the values so that the panels will open properly but if I'm right, the changes will really only be visible when clipping isn't going on.

stratochief66:
I'm not so sure about the clipping causing the strangeness, I'm currently trying it with your Long SLA, so I wouldn't expect any clipping

Could you try these values, and if your satified add them to the PR?

	@MODULE[ModuleDecouple]
	{
		@ejectionForce = 0.2
	}
	@MODULE[ModuleEngines*]
	{
		@maxThrust = 0.3
	}
	@RESOURCE[SolidFuel]
	{
		@amount = 0.10
		@maxAmount = 0.10
	}

ctiberious:
You shouldn't get any strangeness from clipping on the long SLA but that also means that the panels were probably underpowered.  It's only when you have clipping (like with the normal SLA) that the panels acted weird.  At least that was my experience during my testing.  One set of tests I ran included building the "Saturn V" but not including anything below the SLA  (i.e., Apollo CSM, then SLA with the LEM inside).  I used KCT to put that directly into a 300km Earth orbit.  Jettison the panels and decouple the CSM, the panels would slowly, and more importantly, evenly, float away.  But if I reran that test and added the IU, that same process would result in the panels acting strange.  Normally one panel (at random from what I could tell) would fly away rather quickly, the two panels directly next to that one would just sort of float away as expected, and the panel opposite the first would actually kind of float back towards the SLA.

I assumed that in the past someone was trying to correct the "fly away" panel problem and that's why the thrust and ejection force were dropped to almost nothing.  But if you take the clipping out of the equation, then yes, we need to tweek those values on the panels.  I'll add those values to this PR once I run a flight with them.

ctiberious:
I always thought the SLA panels hinged open until they got to a certain point, then were pushed away by springs or something.  Didn't think they actually continued to spin as they floated away.  Of course, even if that is correct, not sure if it's something we can recreate in KSP.  The values you came up with at least get the panels away from the craft so you don't have to worry about bumping into them later.  :)  I've added those values to this PR.

stratochief66:
I think getting more realistic panel deployment would require some research and possibly new models of the panels, so definitely not on the table at present.

As a pedant, I noticed you have @maxAmount = .10 instead of @maxAmount = 0.10. I don't know if it could possibly make a difference, but I've tested with the leading zero :P

ctiberious:
I didn't see an issue with the values in the test I ran but I added the leading zero in just for you.  :)

stratochief66:
Many thanks, the voices are silent again (for now) :P

PR #1091, opened 2016-03-23
Skylab storage capacity

ctiberious:
In a previous PR we corrected the amount of life support that is included with Skylab to reflect what the actual Skylab included, but we didn't adjust the available storage volume.  The 85.5kL of space meant you could easily setup the Skylab with nearly 2 years worth of life support.  I'm not sure if that volume was made up or if there was some documentation to go along with it.  Barring any documentation, I've lowered the volume so that there's only about 45% additional storage capacity.

I've also altered the initial RCS fuel capacity.  According to http://pages.erau.edu/~ericksol/projects/issa/skylab.html#Skylab Systems the Skylab's "Thruster Attitude Control Subsystem" was powered by "Cold gas propellant (nitrogen)" but the Skylab we have uses MMH/NTO for it's RCS thrusters.  I'm not sure if that was done intentionally or not.  And I didn't want to mess with the existing RCS without knowing exactly why MMT/NTO was used in place of the actual Nitrogen system.  But I did want to make sure that the same quantity of fuel was available so I've adjusted the amount of MMH and NTO available to approximately match the weight of Nitrogen that would have been included.  Here's my math:

http://pages.erau.edu/~ericksol/projects/issa/skylab.html#Skylab Systems says Skylab included "22 propellant storage spheres" and that each sphere "stored [nitrogen] at 3,100psi at 100 degrees F in 4.5 cubic foot titanium spheres ".
According to http://siei.org/usefulequations.html, "A K-cylinder holds about 1 cubic foot [at STP, and]....A K-cylinder of nitrogen (at 3000 psi) holds roughly 250 scf"
Based on the above, the Skylab has 22 spheres, each holding approximately 1125 standard cubic feet (scf) of nitrogen.  That works out to 24750 scf, or 700841.95L of nitrogen.  That is roughly 876.75kg of nitrogen based on the density provided in CommonResources.cfg.
At a 45.2%/54.8% MMH/NTO ratio, 332.35MMH + 402.95NTO works out to 876.75kg.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -74,7 +74,7 @@
 	{
 			name = ModuleFuelTanks
 			type = ServiceModule
-			volume = 85500.0
+			volume = 25000.0
 			basemass = -1
 			TANK
 			{
@@ -85,14 +85,14 @@
 			TANK
 			{
 				name = MMH
-				amount = 92.54
-				maxAmount = 92.54
+				amount = 332.35
+				maxAmount = 332.35
 			}
 			TANK
 			{
 				name = NTO
-				amount = 112.2
-				maxAmount = 112.2
+				amount = 402.95
+				maxAmount = 402.95
 			}
 			TANK
 			{
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -84,15 +84,9 @@
 			}
 			TANK
 			{
-				name = MMH
-				amount = 332.35	//see RO github #1091 for explanation of value
-				maxAmount = 332.35
-			}
-			TANK
-			{
-				name = NTO
-				amount = 402.95	//see RO github #1091 for explanation of value
-				maxAmount = 402.95
+				name = Nitrogen
+				amount = 700841.95	//see RO github #1091 for explanation of value
+				maxAmount = 700841.95
 			}
 			TANK
 			{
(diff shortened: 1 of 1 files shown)


ctiberious:
Added notation to the NTO/MMH values to explain where the amounts came from.

Adjusted the ModuleDataTransmitter values for the Skylab antenna so they match the SM antenna.  I'm not positive how powerful the actual transmitter would have been but I'm guessing it would have been similar to what was available on the Apollo.

stratochief66:
Some of that 85.5kL of space needs to be reserved for the crew, methinks :P 

The Skylab antenna only needed to transmit data from LEO, where the SM had an antenna for communicating at lunar distances, so I don't believe they would be comparable.

The largest crew the Skylab ever carried was 3, and at that capacity I believe it had as much space per astronaut as the ISS, roughly (I saw a chart to that effect recently). 

So, at some point a mod would have to step in to trade-off crew capacity/space and supply space I think.

Edit: Ahh, so you're lowering the available space. I think that is fair as the supporting mods stand now. In theory, a mod like RealFuels or something new could step in to allow the trade-off of space reserved for supplies and space reserved for crew. Your reduced volume for supplies seems more reasonable, I think.

ctiberious:
The Nitrogen mentioned on the site I listed is definitely for what was called "Thruster Attitude Control Subsystem (TACS)" which is part of the "Attitude and Pointing Control System".  I'm assuming that's the NASA term for what we call the "RCS".

As for the antenna information, we currently don't change the ModuleDataTransmitter values for RO.  So we're just using the stock values which (for one thing) means Skylab antenna uses a lot of power.  As a comparison, the Apollo Dish in FASA / RO uses:
packetInterval = 0.18 / 0.4
packetSize = 2 / 1
packetResourceCost = 20.0 / 8
Compared to the Skylab in RN_Skylab / RO:
packetInterval = 0.4 / 0.4
packetSize = 2 / 2
packetResourceCost = 20.0 / 20.0
I used the RO Apollo values mostly because I don't know enough about what those settings actually do other than control transmit speed and power consumption.  Basically, using stock values in RO means Skylab transmits very slowly and uses a lot of power to do it.  Using the Apollo values speeds up the transmit speed (especially useful when you have to transmit researched science) and makes the power consumption a bit closer to what other RO transmitters use.

stratochief66:
Yeah, the Apollo one might be a good starting point to go from, but somethings should be changed to reflect their very different ranges and rolls.

@NathanKell any insight into how antenna values should be set for Skylab? I have 0 experience to go from, as a non-RT user.

NathanKell:
A few Mm should be sufficient IMO.

ctiberious:
Well, for the antenna, all I really changed was the packet values.  But maybe I should include a complete RO/RT setup for the Skylab antenna.  Maybe something similar to the Agena Omni antenna?

jwvanderbeck:
On the RCS fuel, forgive me if this is naive, but just converting to the same weight doesn't seem like it would be correct given that the ISP, thrust and possibly storage requirements would be different?

Can we clarify why we have it using different fuel?

> A few Mm should be sufficient IMO.

I think the question was more about power consumption and data rate?

ctiberious:
@jwvanderbeck As for why we're using a different fuel, I'm guessing @raidernick will have to explain that one.  I didn't actually change the fuel source from what we already have.  I just tried to come up with an appropriate quantity of fuel.  That said, my guess is MMH/NTO was put on the RO Skylab because that's the RCS fuel that was used on Apollo.  Why it wasn't used on the actual Skylab I haven't been able to determine.  Maybe the significantly longer mission duration of Skylab meant storing the MMH/NTO was less reliable than using pressurized Nitrogen?

jwvanderbeck:
Yeah I understand your reasoning for not changing it, i'm just more interested in finding out why it is wrong to begin with and if we **can** change it to be historically accurate which would be preferred unless it is for a good reason.

stratochief66:
I'm not sure if raidernick set that fuel combination, it very well could have been me in my initial setting up of this part. I generally focus on the elements that are important to me (mass, scale, consumables and basic function) and in this case you can see I sometimes forget/don't have time to come back and set things right. I probably just copied an RCS profile from an Apollo part to get it worked and never came back and set the right fuels.

Feel free to set the right fuel for the Skylab RCS that your research shows. Apologies I missed this discussion until @jwvanderbeck started calling it out louder :P

ctiberious:
Ok. I'll correct the fuel and quantity for the RCS.  While I'm at it, let me make sure we have the correct thrust rating and other values.

ctiberious:
RCS has been switched to use Nitrogen instead of MMH/NTO.  The thrust value is based on http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19710007043.pdf which says, "approximately 45kg at the beginning of the mission".  45kg thrust = 441.299 N.  I couldn't find anything about the ISP for the thrusters so I used the ISP from the nitrogen thrusters on the Surveyor probe.

Nitrogen initial fuel quantity is set to 700841.95 which is based on the documents and calculations I mentioned before.

I've also updated the ModuleDataTransmitter values to match the values for the FASAAgenaAnt.  Figured that was probably a better antenna to base the Skylab on since they are both Omni and both meant for LEO use.  And I included a RemoteTech setup, also based on the FASAAgenaAnt part.

ctiberious:
A note about the thrust setting on the Skylab.  According to the document I found the thrust was "dependent upon the cold-gas tank pressure and will be approximately 45kg at the beginning of the mission and diminish to 4.5kg at the end of the mission".  I know we have a similar thrust curve system for solid rocket engines.  Could we possibly use that system in ModuleRCS so that thrust gets reduced as fuel is used up?

stratochief66:
Not that I'm aware of. @NathanKell ?

stratochief66:
is this PR good to merge @ctiberious ? or do you want to wait on @NathanKell or try out variable thrust RCS today?

ctiberious:
I think we're good to merge as is.  I'm betting the variable thrust for ModuleRCS isn't going to be available at this point.

kiwivogel:
I'll list cold gas thruster thrust curve in the RCS overhaul topic. Good find!

PR #1092, opened 2016-03-23
BigG adjustments

ctiberious:
I've increased the amount of fuel in the FASABigGeminiRetroModule part.  This gives this part the same delta-V when attached to the Big G that the FASAGeminiUtilitySasRcs gives the standard Gemini.

Because of the animation on the Big G docking port, I needed to increase the acquireRange and captureRange for docking.  These values have to include the length of the extended docking node.

Finally, I've tweeked the position and angle of the RCS thrusters on the FASAGeminiBigGDock to minimize unwanted torque/thrust during attitude control and translation maneuvers.  The new position/angle assumes that the guidance part's RCS is turned off (since it's only supposed to be used during re-entry).  The new position/angle also assumes just stock "fuel".  If you load up the FASAGeminiBigGDock with 39k liters of extra "fuel" (whether that be extra life support, extra fuel, extra lead, extra whatever) you will likely need to add additional small RCS thrusters to account for the extra mass of the BigG.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg
@@ -420,7 +420,7 @@
 	@node_stack_bottom = 0.0, -0.6, 0.0, 0.0, -1.0, 0.0, 4
 	@category = Engine
 	@title = Big Gemini Retro Module
-	@description = The Big Gemini Retro Module that contains 4x solid rockets to de-orbit Gemini. Decouples before re-entry.
+	@description = The Big Gemini Retro Module that contains 4x solid rockets to de-orbit Big Gemini. Decouples before re-entry.  Provides approximately 106dV for re-entry burn.
 	@mass = 1.0978748
 	!MODULE[ModuleRCS]
 	{
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg
@@ -467,7 +467,7 @@
 	MODULE
 	{
 		name = ModuleFuelTanks
-		volume = 112.74
+		volume = 165.4
 		type = PSPC
 		basemass = -1
 	}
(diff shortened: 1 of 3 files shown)


stratochief66:
seems reasonable. Since you targetted the same dV as Gemini, I may be concerned that there may not be enough space on the part to actual support that amount of fuel, but that should be relatively easily calculable.

Did you weld the RCS thrusters to the part or move their welded position(s)?

Could you provide a craft file for these changes so I can see how you put them to together? If that proves difficult, a picture of your assembly may suffice.

ctiberious:
Should be plenty of space but you likely have a better way to calculate it since I've never done that.  Unlike the Gemini part the Big G only had a small passageway (that could be pressurized but my guess is only when the whole BigG was assembled), storage for the flights water and oxygen, and the retro thrusters.  All on a part that was 180" in diameter compared to the Gemini part which was only 90" in diameter.

I adjusted the welded positions I had previously supplied.  Basically, I removed the thrusters and loaded into the game.  Then I placed the thrusters and using "Part Angle Display" mod, I fine tuned the angle and position of the thrusters until RCS Build Aid gave me the smallest unwanted thrust/torque values.  Then made a craft file and pulled the position and placement values and added the thrusters back onto the part.  Reload the game with the thrusters welded back on and verified that I was getting the same values in RCS Build Aid.

And yes, I can provide a craft file of at least the Big Gemini portion.  I'd have to remove some of my custom tweeks to include a craft file with the launch vehicle in place, but since there were a couple different launch vehicle options, I'm guessing most important here is just the Big G itself.  I'll add it to this PR in a few minutes.

stratochief66:
Since Mechjeb doesn't show the dV for some things, like the retrosolid or RCS, it might be worth noting how much dV those provide in the part descriptions?

ctiberious:
Yeah.  To get Mechjeb to display the dV I had to temporarily remove the ModuleDecouple from the part.  The trouble with putting the dV in the description, though, is it would only be accurate if you create the same capsule I use to test with.  If you put extra part (like the Gemini antenna or more/fewer lights) or mess with the quantity of fuel and/or life support, that will change the values.  Also if you jettison parts in a different order.  What I'd be more inclined to do is talk with Sarbian and see if there is any way to get mechjeb to display dV for parts that include their own decoupler.

stratochief66:
Yeah, there are a few RO oriented requests we have in with Sarbian, including that one I think. Those won't come for a few months, until we have 1.1 and mods are stabilizing for now. For now, could you include the dV you found in the description with an "about" or some caveat? If someone takes it as gospel, that is their own fault ;P

ctiberious:
Sure.  I can do that.

stratochief66:
I had a little trouble getting the craft into orbit with a Saturn iB, but that was my own bad trajectory. Carrying less crew or supplies should allow it more free dV, or to each a higher orbit, so everything looks very good! I found that keeping the roll position controlled for a lifting re-entry was basically impossible with the tiny RCS thrust of the standard Gemini, but that is easy to adjust in the future if we want to.

This means it would dock with Skylab, correct?
https://github.com/ctiberious/RealismOverhaul/blob/patch-40/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_BigG.cfg#L672

I will have to try that out sometime, as that how I would like to use this craft :)

ctiberious:
Also, keep in mind that the Saturn 1B was quickly thrown out as a viable launch vehicle for the Big G.  In the end it was planned to fly on the Titan IIIM, Titan IIIG (both with a narrower service module) and the Saturn INT (S-1C + S-IVB).  Put a Saturn INT together and you've got plenty of thrust.

And yes, the Big G docking node will dock with Skylab.  Just keep in mind that you can't really control from the docking port.  I'm assuming it's the texture but when you try to "control from", rather that looking behind (i.e. if you were facing Prograde while controlling from the cockpit, you'd expect to be looking Retrograde when controlling from the docking port) you're sort of looking up or to the side or something.  So the H and N hotkeys don't equate to "forward" and "backward" like you'd expect them to.
I've docked with my Skylab twice now and both times I've lined the approach up as if the docking port was on the nose.  then when I'm about 20m out, I come to as complete a stop as I can and spin around so I'm facing the opposite directly.  Then, while still controlling from the cockpit, slowly back up until the port is in the docking cone.  I also found that I needed to go very slowly (I use Docking Port Alignment Indicator so have me speed down to 0.02m/s) for the last meter or two and just allow KSP to "pull" the parts together with the little bit of "magnetism" the parts have in RO.  That's very different from the more aggressive docking approach I make with the Apollo CSM and LEM.

ctiberious:
Alright.  I've added a note to both the Gemini and BigG retro part to show that they produce approx. 106dV for re-entry.

PR #1096, opened 2016-03-24
Squad fuel tank updates

PhineasFreak:
* Add RO support for the C7 Brand Adapter tanks.
* Add RO support for the large Xenon storage tank.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_FuelTank.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_FuelTank.cfg
@@ -15,6 +15,45 @@
 		type = Structural
 	}
 }
+
+//  ==================================================
+//  C7 Brand Adapter (straight).
+//  ==================================================
+
+@PART[adapterSize2-Size1]:FOR[RealismOverhaul]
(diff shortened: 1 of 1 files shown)


stratochief66:
Look good to me. Not tested in game by me.

Do these parts physically fit any of the other stock parts? I know some other stock parts have been rescaled to represent various things, so I'm curious of the functionality/use case for these parts.

Did you test the TweakScale function in game? We generally don't add TweakScale compatibility to newly configured parts, but if it fully works, it works.

As far as I'm aware nobody who reviews and mergers RO pull requests has much experience with TweakScale (me @stratochief66, @NathanKell , @SirKeplan) might be skin deep, or the other guys might simply suggest to remove the Tweakscale bit. Having better and more consistent tweakscale support for RO could be a good thing (for simple parts it may not mess up, like tanks, physical adapters, structural parts) but we really lack a person to be our TweakScale lead.

PhineasFreak:
These are just generic 2.5 m - 1.25 m adapters for the Rockomax type tanks. As i am aware, no changes have been made to the diameter of these tanks.

To be honest, i hate TweakScale. I included the TS compatiblity patches only because all the other tanks had already been configured for it.

PR #1101, opened 2016-03-25
Improve Tantares Venera descriptions

stratochief66:
I also noticed that the experiments are the opposite of how I would expect (all in the antenna, none in the lander core)

Should this be reversed?

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Tantares/RO_Tantares_Venera.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Tantares/RO_Tantares_Venera.cfg
@@ -13,7 +13,7 @@
 	%mass = 0.56
 	
 	%title = Venera 9-13 Lander
-	%description = Represents the Venera 2MV through Venera 14 (4V1M). First human made object to impact another planetary body.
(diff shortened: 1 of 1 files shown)


SirKeplan:
Description doesn't look quite right. it seems Venera 2MV was a much earlier Venera than Venera 9 (4V-1) onwards

stratochief66:
Shallow information on astronautix & Gunters is ambiguous:

http://www.astronautix.com/craft/mars2mv3.htm

http://space.skyrocket.de/doc_sdat/venera-2mv-1.htm

They refer to 'soft landing' and 'venus lander', which I've only seen this design for.

SirKeplan:
The earlier Venera landers looked more like this http://www.russianspaceweb.com/images/venera7_sa_200.jpg

http://www.russianspaceweb.com/3mv.html

stratochief66:
Ahh! Alright then. Write a description change you would suggest, and I will
modify the PR.

SirKeplan:
"Represents the Venera 9 (4V-1) through Venera 14 (4V-1M) landers. First human made object to successfully return images from another planetary body."

PR #1107, opened 2016-03-27
Attitude control of upper stages to use MMH/NTO

Max-Q:
Modifying the attitude control (RCS thrusters) of both the ICPS and the EUS to use a combination of MMH/NTO instead of "Monopropelant". Also, changing the tank type of the ICPS from standard to cryogenic.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_SLS_Upper_Stages.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_SLS_Upper_Stages.cfg
@@ -24,10 +24,12 @@
 	{
 		%position = 0, -4.1841, 0
 	}
+	
 	@node_stack_bottom = 0,-5.99504,0,0,-1,0,5
 	@node_stack_top = 0,3.32604,0,0,1,0,5
 	@node_stack_top2 = 0,2.34120,0,0,1,0,5
 	@node_stack_interstage = 0,-0.58827,0,0,-1,0,2
+	
 	%RSSROConfig = True
 	@mass = 3.66 //estimate just under 5% heavier than standard 5 m DCSS
 	@title = Interim Cryogenic Propulsion Stage [5m]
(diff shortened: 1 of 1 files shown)


stratochief66:
Cool! 

I added the MMH/NTO RCS to the procedural upper stage (or maybe one of them) as a proof of concept that it worked. In practice, the whole group of RO RCS fuels could be added as well, just like the stock RO pods let you switch and choose.

The ICPS tank should be 'standard'. The ICPS is expected to boil off a substantial portion of its LH2 while coasting, just like the S-IVb did, and the DCSS would if left for a many orbit coast. The ICPS is pretty much a DCSS with a stretched hydrogen tank, to ensure enough LH2 is left for a substantial burn after a long coast. The ICPS, DCSS, and S-IVb have some insulation, but not enough to substantially reduce LH2 boil-off. That is how we classify the difference between 'Default' and 'Cryogenic', apologies as I don't believe that is well documented elsewhere.

Max-Q:
Thank for the clarification about the insulation "thickness" of the ICPS/ DCSS.
Regarding the RCS fuels - I think that today most RCS use anyway MON3 or (MON1) instead of pure NTO because of freezing.

stratochief66:
If you can find an information source that the ICPS intends to use MON3 or MON1, or that the DCSS uses one of those,than makes sense, yeah.

For the configurable upper stages we probably still want to support the whole fuel range used by the regular RO RCS blocks, because at some point they might actually get configured for RP-0 :)

PhineasFreak:
The DCSS uses Hydrazine for it's ACS system (source: [Delta IV User's Guide)](http://www.ulalaunch.com/uploads/docs/Launch_Vehicles/Delta_IV_Users_Guide_June_2013.pdf) but as @stratochief66 mentions, it would be much better if it can support the whole range of possible RCS fuels. You can define multiple RCS fuels by including the following fields (example case):

``%useRcsConfig = RCSBlockDouble``
``%useRcsMass = True``
``%useRcsCostMult = 0.25``

The RCS config depends by the type of the RCS block (single nozzle, dual, quad etc): https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/RO_RCS_Config.cfg

stratochief66:
@PhineasFreak ahh! That is clever, i will try to apply that sometime soon.

Would I want to remove the original RCS config for the part first? I'm a bit unsure if this will work on the SSTU parts, they are relatively complex.

PR #1113, opened 2016-03-29
VSR Radial Avionics Module

PhineasFreak:
* Minor fixes to the Radial Reaction Wheel (thermal & resource consumption).
* Add RO support for the CH-R1 Fly-By-Wire module (very similar to the MechJeb AR202 module).


NathanKell:
Looks good, except we don't use gendered language where possible, so can you change to crewed/uncrewed?

PhineasFreak:
It should be all good now!

PR #1121, opened 2016-04-04
SSTU Realism and playability enhancments

Max-Q:
- Set TAC containers for a 4 person, 21 day mission
- Set battery and solar pannel power to real world values
- Switch CO2 removal to CAMRAS
- Set chute to real world values (only stats, not look)
- Added RemoteTech support
- Added ConnectedLivingSpace support
- Moved docking port to general file

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Orion.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Orion.cfg
@@ -73,7 +73,18 @@
 	}
 	@MODULE[ModuleDockingNode]
 	{
-		%nodeType = NASADock
+		@nodeType = NASADock
+		%acquireForce = 0.5 // 2
+		%acquireMinFwdDot = 0.8 // 0.7
+		%acquireminRollDot = -3.40282347E+38
+		%acquireRange = 0.25 // 0.5
+		%acquireTorque = 0.5 // 2.0
+		%captureMaxRvel = 0.1 // 0.3
+		%captureMinFwdDot = 0.998
+		%captureMinRollDot = -3.40282347E+38
+		%captureRange = 0.05 // 0.06
+		%minDistanceToReEngage = 0.25 // 1.0
+		%undockEjectionForce = 0.1 // 10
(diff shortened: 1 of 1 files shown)


stratochief66:
Is this what they intend to use for Orion? citation if so.
https://github.com/KSP-RO/RealismOverhaul/pull/1121/files#diff-93ca0f9d764d2d72e594ee71fe58d73bR162

It looks like you had a data source from some changes, which is excellent. However you didn't include a link to said source in the code itself or in the PR.

stratochief66:
Nothing in this link leads me to believe they would carry 90kg of oxygen as a compressed gas. I would still expect they would carry it as a cryogen unless a source explicitly states otherwise.
https://github.com/KSP-RO/RealismOverhaul/pull/1121/files#diff-93ca0f9d764d2d72e594ee71fe58d73bR313

Any other life support people have better info or insight into that?

Max-Q:
Now, I am a bit unsure about that, I will try to find more information. The information on this site http://spaceflight101.com/spacecraft/orion/ lead me to believe that it is in compressed form. The drawings are labeled with GN2 and GO2. Guess the G stands for gasseous. Also he gas tanks are described as follows:
"The Gas Delivery System consists of separate tanks and plumbing for the Oxygen and Nitrogen System. A total of four gas tanks are installed in the ESM, each with a capacity of 33 Kilograms of gas stored at 275 bar. The Oxygen Delivery System is comprised of two tanks that interface with a redundant system of two delivery lines that can be selected and isolated via a series of valves.

The tanks and lines include pressure and temperature sensors, heaters and regulators that control the pressure at which Oxygen is delivered to the Crew Module Umbilical. The Nitrogen Delivery System is fed from one tank that interfaces with two delivery lines that include temperature and pressure sensors, on-off valves and pressure regulators. A fourth tank can be connected either to the N2 or O2 system via a jumper based on mission requirements." ...

stratochief66:
Yep, that is definitely sufficient. I agree G is for gaseous, and the
pressure of 275 Bar also indicates gaseous.

Very strange! I wonder if they will have the capability of using liquid
gases to store more for longer mission.

PR #1122, opened 2016-04-04
SSTU: Added variants of docking ports

Max-Q:
iLIDS compatible docking port with and without parachutes


SirKeplan:
Cool. however it looks like something got messed up with the character encoding. an apostrophe has gone very strange in the description field.

stratochief66:
@Max-Q can you fix the borked character encoding in this PR? As @SirKeplan said, something appears to have gone crazy there.

Otherwise, we've got to close (reject) this PR.

Max-Q:
Excuse the delay, I thought the wrong encoding was in the patch description not in the patch itself. I have modified the patch creating a new commit in the process. (Sorry, I forgot to give this commit a describing name). If you are unhappy with that you can reject both commits and I will send in the corrected version again.

PR #1123, opened 2016-04-04
Rescaled FASA IVA's for RO - Attempt #2

cytosine:
FASA IVA's were rescaled to comply with RO external model sizes.

Also fixed a few Gemini Lunar Lander bugs.  The lander pod wasn't being scaled so it was tiny compared to other craft.  However, I later noticed that the lander legs and other related Gemini Lander parts will also need resizing.  I will work on those as part of another PR, unless you guys think otherwise?


cytosine:
Nope, that didn't seem to fix the problem. Going to close this PR and try again...

PR #1126, opened 2016-04-06
Update Centaur tanks

ctiberious:
We previously changed the default Centaur tank from type=ServiceModule to type=Balloon which makes sense as the Centaur was based on the same balloon tank concept as Atlas.  Plus most early Centaur tanks (those used on Atlas) had basically no insulation so high boil off rates.  However, both versions of Centaur used on the Titan (Centaur D-1T and Centaur T) were intended for longer duration flights and therefore included insulation to reduce boil off.  And the latest Centaur III (used on Atlas IIIB/V) included the same kind of insulation for the same reasons.  I've therefore added a new definition specifically for the D-1T, and set the D-1T, D3 and T tanks to all have type=BalloonCryo.  Technically the D3 and T versions should have about half the boil off of the D-1T (1%/day vs just under 2%/day) but I don't know if we currently have the ability to control boil off to that degree.

I also made a number of weight changes.  Unlike in previous updates, this time I specifically aimed for matching the dry weights of the tanks.  And in all cases, I'm including the weight of the payload fairing base, plus rcs thrusters and engines in the weight.  I've included a note on the mass line indicating exactly what I included to come up with the indicated weight and exactly what each extra parts weight was.

This PR is removing the old FASAGeminiLFTCentarCSM_D3.  Previously I had been under the impression that the Atlas IIIB flew a different model Centaur. But after much research and reading, I've come to find that the "Centaur III" (also known as the "Common Centaur") was flown on both the Atlas IIIB and Atlas V.  So there's no reason to include a separate Centaur tank specifically for the Atlas IIIB.

During my research, I discovered that while modern versions of the Centaur use a Hydrazine based RCS system, the original Centaur used a Hydrogen Peroxide based RCS system.  For the Centaur D1 and D-1T tanks, I've added 242 pounds (76.71 L) worth of HTP to the basic Balloon tank.  For the D2, D3 and T models I couldn't include Hydrazine in the existing ModuleFuelTanks because Hydrazine can't go in Balloon or BalloonCryo tanks.  I've therefore added a 320 pound (144.57 L) Hydrazine RESOURCE to these tanks.  Unfortunately that means players can only decrease the amount available (versus completely removing or altering the tank) but it's the best solution I have right now.  If there is a way to add a second ModuleFuelTank module to a part, that would work better but the tests I ran showed the two modules interacting with each other.

For all five of the Centaur tanks, I've been able to track down LOX and LH2 tank capacity information.  I've therefore reset the fuel tanks max volume, and included default LOX and LH2 tanks to match the documentation I've found.  The only concern I've come across is I can't seem to match up the tank capacities found in one set of documents, with the Gross Mass/Burn Times found in another set of documents.  It seems in all cases, the Gross Mass/Burn Times I'm finding assume that less than full tanks are launched.  Once in game, however, if you manually reduce the available fuel to match the documented burn times, the gross mass is within 200kg which I'm thinking is about as accurate as we can get.

Mass and burn time data was pulled off astronautix.com.  Pages for Titan 4, Titan 3E, Atlas V, Atlas IIIB, Atlas IIIA, Atlas IIAS, Atlas II, Atlas I, Atlas Centaur SLV-3D and Atlas Centaur LV-3C.
Other documentation comes from the following:
http://www.ulalaunch.com/uploads/docs/Published_Papers/Extended_Duration/AtlasCentaurExtensibilitytoLongDurationInSpaceApplications20056738.pdf
http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/CentaurUpperstageApplicabilityforSeveralDayMissionDurationswithMinorInsulationModificationsAIAA20075845.pdf
http://history.nasa.gov/SP-4230.pdf
http://www.alternatewars.com/BBOW/Boosters/Centaur/Centaur.htm
http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/HistoryoftheTitanCentaurLaunchVehicle.pdf
http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/TheCentaurUpperStageVehicleHistory.pdf

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Titan.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Titan.cfg
@@ -157,7 +157,7 @@
 	@title = Centaur A/B/C/D/D1
 	@description = Tank and thrust structure for the Centaur A/B/C/D/D1 models. Add two RL10A series engines to complete the stage.
 	@attachRules = 1,0,1,1,0
-	@mass = 1.125	// From Astronautix, based on the Centaur I or Centaur D-1B.  Add leadbalast to duplicate older versions of this model Centaur
+	@mass = 1.003	// From Astronautix, based on the Centaur I or Centaur D-1B.  1.7t total weight comes from 2x RL-10A-3-3A engines (0.274), 4x Centaur RCS thrusters (0.024), 4x 138N attitude jets (0.012), and payload fairing base (0.387).  Add leadbalast to duplicate older versions of this model Centaur
 	!RESOURCE[LiquidFuel]
 	{
 	}
@@ -166,10 +166,39 @@
 	}
 	MODULE
 	{
-		name = ModuleFuelTanks
-		volume = 42511.0
+		name = ModuleFuelTanks	//tank size taken from http://www.alternatewars.com/BBOW/Boosters/Centaur/Centaur.htm for Centaur D-1A
+		volume = 46544.62	//46467.91 + 76.71 HTP
 		type = Balloon
 		basemass = -1
+		TANK
+		{
+			name = LqdHydrogen
+			amount = 35820.81
+			maxAmount = 35820.81
+		}
+		TANK
+		{
+			name = LqdOxygen
+			amount = 10647.1
+			maxAmount = 10647.1
+		}
+		TANK	//Tank size taken from http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/TheCentaurUpperStageVehicleHistory.pdf.  Older, 242lb version using HTP.
+		{
+			name = HTP
+			amount = 76.71
+			maxAmount = 76.71
+		}
+	}
+}
++PART[FASAGeminiLFTCentar]:AFTER[RealismOverhaul]
+{
+	@name = FASAGeminiLFTCentarCSM_D1T
+	@title = Centaur D-1T
+	@description = Tank and thrust structure for the Centaur D1T models. Used on the Titan III-E. Add one or two RL10A series engines to complete the stage.
+	@mass = 1.934	// from Astronautix Titan 3E.  2.631 total weight comes from 2x RL-10A-3-3 engines (0.274), 4x Centaur RCS thrusters (0.024), 4x 138N attitude jets (0.012), and payload fairing base (0.387).
+	@MODULE[ModuleFuelTanks]
+	{
+		@type = BalloonCryo	//Titan version of the Centaur D-1 included radiation shielding specifically meant to increase the stages life by cutting the boil off rate to less than 2%/day.  http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/CentaurUpperstageApplicabilityforSeveralDayMissionDurationswithMinorInsulationModificationsAIAA20075845.pdf
 	}
 }
 +PART[FASAGeminiLFTCentar]:AFTER[RealismOverhaul]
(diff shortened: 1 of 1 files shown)


stratochief66:
Nice source finds :)

Did you compare the Astronautix burnout masses to those on bnumz? ie.
http://www.b14643.de/Spacerockets_2/United_States_3/Atlas-Centaur/Description/Frame.htm

I'm reading through this document. http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/CentaurUpperstageApplicabilityforSeveralDayMissionDurationswithMinorInsulationModificationsAIAA20075845.pdf

Very interesting, I will need to read more. I'm not sure if we can currently match exact boiloff rates, but I believe that is or should be a key goal once we reach 1.1. So finding good sources like these and noting them down is very important for that future work.

On page5, the LO2 boil-off through a number of flights were 1.0 to 2.6% per day (6 flights of data), and LH2 4.1-5.1% per day for 2 flights data. The  LH2 rates was on the G-Prime, with 3 layers of insulation. I'm quite curious what the LH2 boil-off would be estimated per day or hour for an un-insulated early centaur

stratochief66:
@Schnobs differential and variable boil-off suggests more need for a mixing ratio change mod.

NathanKell:
@stratochief66 AFAIK Centaur did not have a propellant utilization system, but S-IV and S-IVB did, and I do plan to add such a thing to RF.

stratochief66:
@NathanKell I wasn't aware that was a planned feature, good to hear!
I also didn't think that centaurs used it, but apparently at least the Titan IV 's centaur did, page 7 of this:
http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/HistoryoftheTitanCentaurLaunchVehicle.pdf

ctiberious:
@stratochief66 I did look at the bnumz (b14643.de, right) dry masses but I can't tell how reliable they are.  According to that site, every version of Centaur D (D - D1AR) has the exact same mass which I figure isn't correct since the D uses RL10A3-3 engines and the D-1AR used RL10A3-3A engines.  It also lists different masses and lengths for the Centaur used on Atlas IIIB versus what was used on Atlas V, but according to the ULA document, "The Centaur Upper Stage Vehicle" (the last link I provided), the Atlas IIIB and V both used the "Common Centaur" or "Centaur III".  In the end I went with the astronautix dry mass data because each version had a distinct mass.
One thing I did notice when I was looking at the alternatewars.com link was this page: http://www.alternatewars.com/BBOW/Boosters/Centaur/Centaur_GIW.htm  If I'm reading this correctly, of the 6 "Centaur D" launches that are listed, each had a different dry mass.  Just looking at the "Basic hardware", there's a 650 pound difference between the lightest launch (AC-11) and the heaviest (AC-9) and they didn't get progressively lighter.  So I don't think we'll ever have an exact mass figure.
My guess is that's also why my fully fueled masses and burn times don't match either astronautix or bnumz.  My guess is the tank capacities I've listed are the maximum each version could store, and what we're seeing on astronautix and bnumz are either taken from what was used on one particular flight, or perhaps an average of what was used on several flights.

Schnobs:
http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19680027035.pdf -- "CENTAUR LAUNCH VEHICLE
PROPELLANT UTILIZATION SYSTEM". The section on history sounds as if even the very first Centaurs were supposed to have a PSU. 
http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19720017275.pdf -- mission report of an Atlas-Centaur that failed to reignite. Launched 10-Aug-1968. According to the text, not only the Centaur but even the Atlas itself had a PSU.

The latter has lots of hard data: trajectory, acceleration, dry masses and consumables down to lubrication oil. A document to squirrel away.

For good measure, http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19760021171.pdf -- post-flight summary of the first Titan-Centaur launch, again with lots of data and schematics.

Schnobs:
http://www.alternatewars.com/BBOW/Boosters/Centaur/Centaur_GIW.htm is a meanie.

The numbers it gives are lifted from post-flight summaries like the two I linked above. The full text of the summaries lays out what made up some of these groups. No surprises here. For example, "Guidance" is gyros, accelerometers and the digital navigation computer.

Most of the weight variance is due to flight instrumentation and miscellaneous equipment for which I can find no further explanation. For all I know this may have been one-offs like extra sensors and cameras overseeing this or that.

But even the propulsion, guidance and control group weights vary by quite a bit. Somehow I'm getting the impression that every Centaur was a unique hand-crafted vessel.

ctiberious:
@Schnobs I think you're right, at least where older Centaur was concerned.  That's why I included not only the website where I got the proposed mass for each version, but also what parts I used and how much those parts weigh.  My guess is those early Centaurs were tweeked with every flight so they could test things.  That would certainly help explain why it's so difficult to get hard data on the older versions of Centaur.  But I'm guessing that the more modern versions tend to be more "cookie cutter" since it's an established vehicle and it's a bit easier to get data on it.

PR #1128, opened 2016-04-09
RN skylab RW fix

raidernick:
-somehow this wasn't addressed in the config. I fixed it now to override
the stock values


stratochief66:
Thanks @raidernick 

Please poke this PR when you've had a chance to test it in game & confirm it applies as expected.

raidernick:
@stratochief66 I tested it and it works

PR #1133, opened 2016-04-10
Vinci Updates

Theysen:
- added 5 restarts to engine config
- adjusted mass to be in line with SSTU (big ? right now as actual mass is not well-sourced / differ)

- added actual Vinci config to StarShineIndustries model


SirKeplan:
@Theysen have you tested this ingame? unless you have another file you haven't included, doing `engineType = Vinci-180` isn't going to do anything.

It would have to be engineType = RL60 and then you remove the non VInci stuff like how the RL10B-2 is treated.

jwvanderbeck:
Or should we maybe add a new generic for the Vinci? How different is it from the RL60? If too much gets offloaded into the part config then it starts to defeat the purpose.

Theysen:
The enginetype error actually remained from that, I had one earlier. SSTU and   AIES  would then need a little update

SirKeplan:
hmmm, what was the original reason for them being on the same part? (with the AIES one anyway).

Theysen:
So basically we could split the RL60_Vinci config into RL60 and Vinci and just edit the according few mod files. SSTU has a combined config for RL10-B2 and Vinci, AIES just offers both variants on a quite incompatible model anyways, to be honest.

jwvanderbeck:
Well really the underlying question is simple - does it make sense to keep them as configs of the same engine, or not.  Is the Vinci a derivative of the RL60? Or a completely separate engine.

Theysen:
No RL60 was actually abandoned by ESA and the development split up. 
[http://www.alternatewars.com/BBOW/Space_Engines/RL60.pdf](url)

They might still have same design structures but aren't a joint project anymore

NathanKell:
So...what's the status here? Have they been split?

Theysen:
I just updated the mass again, didn't split the configs though. It should be fine until there's an unlikely rise of Vinci engine models coming out which I doubt. 

I'm sorry for those others strange commits, I didn't have the time to get into this properly, so I'll just push the updated file. (are they deletable somehow?)

PR #1134, opened 2016-04-10
Add Engine Config for RD-57/RD-57M

SirKeplan:
This is an engine i have modelled in blender, intending to release soon along with 1 or 2 others.
Thought i would put the Engine Config in RO.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD57_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD57_Config.cfg
@@ -1,9 +1,43 @@
-//	RD-57
-//	Vodorod
-//
-//	https://web.archive.org/web/20050319103520/http://home.earthlink.net/~cliched/engines/lyu_eng.html
-//	http://www.lpre.de/resources/articles/11D57.pdf
-//	
+//	==================================================
+//	RD-57 global engine configuration.
+// 
+//	similar design to 11D54 except with gimbaling
+// 
+//	Weight 		840kg
+//	isp			456.5
+//	Thrust		40,000 kgf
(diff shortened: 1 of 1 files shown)


stratochief66:
I look forward to the new model! Much better than my lazy cloning, like I did for the RD-107/108 :)

https://github.com/KSP-RO/RealismOverhaul/pull/1134/commits/1b9743c2522b6280ee4e1c1dfc0c4d26083e0e00#diff-35ead5ad2163114f199c8789d1fd7f8aR16

I can guess what you copied the config from, this should probably be RD-57 ?

I don't think you need this line unless you set the two configs to have differing masses. I could be wrong of course:
https://github.com/KSP-RO/RealismOverhaul/pull/1134/commits/1b9743c2522b6280ee4e1c1dfc0c4d26083e0e00#diff-35ead5ad2163114f199c8789d1fd7f8aR18

I was struck by how the first config appears to throttle very deeply, and the second one only throttles somewhat. Is that correct?

stratochief66:
Example. given your very recent addition, I believe you do want to set them to different masses then:
https://github.com/KSP-RO/RealismOverhaul/blob/b94022f1ef4193a923b0cc2712a56d660b54a4bf/GameData/RealismOverhaul/Engine_Configs/N1_BlockA_Config.cfg#L33

SirKeplan:
ah thanks @stratochief66. Yeah i just noticed the mass thing, already fixed. The throttling though... yeah not sure about that, maybe that should change, those are the values i found from different sources.

stratochief66:
Yeah, fair enough. If you have it sourced that is good enough. It just surprised me, so I figured it was worth mentioning :)

PR #1136, opened 2016-04-11
Fixed RemoteTech for RN-Skylab

cytosine:
Fixing the RemoteTech RO patch for Skylab in an earlier PR was good, but it exposed a different bug.  RemoteTech was (apparently) designed to work with Squad's ModuleAnimateGeneric.  Skylab was designed with the Firespitter equivalent animation modules back in 2014 when Firespitter probably had features the stock module didn't (perhaps).

This patch converts the antennae animation (and just the antennae animation, no other animations) to use ModuleAnimateGeneric so RemoteTech can conveniently avoid this bug.  Current generation (v1.05) ModuleAnimateGeneric looks beautiful with the Skylab antennae animation.  Perhaps v1.05 introduced features similar to Firespitter?  I did not test any of the other animations so there may or may not be a purpose for the Firespitter anim module still.

Added RemoteTech configs to the Skylab ASAS so it functions just like every other probe in RO including a small electric penalty.  Also enabled the ASAS to act as a probe control station when 3 crew members are present (since Skylab crews were typically 3-persons).

Skylab now requires a minimum of one crew member to operate, or an attached ASAS within communications range.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -307,6 +307,13 @@
 {
     !RESOURCE[ElectricCharge] {}  // Let's pretend this is a battery-less computer!  Batteries are stored in the Skylab.
 
+    @MODULE[ModuleCommand]
+    {
+       %RESOURCE[ElectricCharge] {
+           %rate = 0.18
+       }
+    }
+
     %MODULE[ModuleSPU] {
 	%IsRTCommandStation = true
 	%RTCommandMinCrew = 3
@@ -314,8 +321,7 @@
 
     %MODULE[ModuleRTAntennaPassive] {
         %TechRequired = unmannedTech
-        %OmniRange = 3000
-        %EnergyCost = 0.05
+        %OmniRange = 300000
         
         %TRANSMITTER {
             %PacketInterval = 0.3
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -202,7 +202,7 @@
 		@PacketResourceCost = 4.0
 	}
 }
-@PART[skylab]:NEEDS[RemoteTech]:FOR[RealismOverhaul]
+@PART[skylab]:AFTER[RemoteTech]
 {
     !MODULE[ModuleDataTransmitter] {}
 
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -31,7 +31,6 @@
 
 	@MODULE[ModuleCommand]
 	{
-		@minimumCrew = 1
 
 		@RESOURCE[ElectricCharge]
 		{
@@ -206,21 +205,6 @@
 @PART[skylab]:NEEDS[RemoteTech]:FOR[RealismOverhaul]
 {
     !MODULE[ModuleDataTransmitter] {}
-    !MODULE[FSanimateGeneric],0 {}   // It would seem that RemoteTech wasn't intended to work with the Firespitter animation module.  Kept getting an 'index error' when testing from orbit using Kerbal Construction Time.
-
-    MODULE
-    {
-       name = ModuleAnimateGeneric  // Replaced the Firespitter animation module with Squad's stock animation for RemoteTech compatibility.  Looks just as good, didn't notice any difference in antenna animation quality or sequences.
-       animationName = Deploy
-       isOneShot = false
-
-       startEventGUIName = Deploy Antennae
-       endEventGUIName = Retract Antennae
-       // toggleActionName = Toggle Antennae
-       actionGUIName = Toggle Antennae
-       allowAnimationWhileShielded = False
-       allowManualControl = false
-    }
 
     %MODULE[ModuleRTAntenna] {
         %Mode0OmniRange = 0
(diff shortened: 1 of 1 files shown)


raidernick:
I appreciate you fixing the remotetech problems but I have to insist you remove the changes to the animation module and the minimum crew requirements. The animation module uses firespitter in it's entirety for a very specific reason. The stock animation module has trouble with shape key animations and causes part lag when the animation is not deployed. This may not be noticeable on newer computers but I got a LOT of complaints about the parts from people over the years about this.
Also, the station was designed to be launched and operate without a crew, all functions could be controlled remotely from the ground. The ASAS is a crutch to avoid a stock bug regarding having an unmanned and manned module on the same part but the station does NOT need any crew to operate it's major functions.

stratochief66:
@cytosine there appear to still be good reason to use firespitter, and it is up to date (even available for 1.1) even if it was originally created a few years ago.

I don't really see a big difference of the default crew being 1 over being 0. however, adding remotetech to the ASAS seems like a good idea. I'm not a remotetech user, so I'll take your word for it if you say support is non-existent or broken for this part. skylab produces plenty of power to run it when the panels are out, and it has a pretty sizeable battery as well.

Could you modify this PR to just add the RT support to the ASAS?

raidernick:
Also the asas changes should be made to the config which actually modifies the scale of the ASAS, which is not this file so the configs are not split among multiple files for 1 part.

cytosine:
Gents,

I just realized something;  The original RN-Skylab mod by itself has minimum crew = 1 which enforces the proper use of the ASAS unit for remote control of the Skylab.  But the existing RealismOverhaul patch changes minimum crew to 0 which allows us to pilot the Skylab *without* the ASAS or crew.

So you guys are all agreeing that we should set minimum crew back to 1 which forces us to use the ASAS for remote control, right?

As for the other two points (RemoteTech and anim), it's late and I've had too much beer!  Will pick this up in the morning...  :)

Thanks!

stratochief66:
@cytosine if the original config called for 1 crew member, that makes sense. the best thing to do would probably be to just remove the @minimumCrew = 0 line, since with it the original config value will be retained.

cytosine:
Understood, no sense in un-patching a patch when deleting it from RO just makes better logical sense.

As for the RemoteTech problem, I'm going to reach out to the RT crew to see if they can verify and maybe even help with the Firespitter compatibility issue.

Thanks!

raidernick:
The firespitter dependency is non negotiable, please remove the changes

jwvanderbeck:
Ok so sounds like we need to remove the animation module changes, and remove the crew overrides then this should be good to go.

@cytosine can you make those changes?

cytosine:
No @raidernick, this problem is different than what you just described.  No it's not RemoteTech's fault just like it's not your fault either.  I'm already working on a good solution for you, but I need some time, so please sit tight.  Unfortunately I've been away from my comp all-day so I haven't had time to make any progress yet.  Once I understand the root issue and a good solution (if one exists) then I'll bring this team up to speed and you can veto it all day long.  Fair?

cytosine:
Folks,

I finally got around to making these changes since 24 hours ago when I originally proposed them and I properly tested them too.  Unfortunately, removing the Firespitter work-around re-brakes the problem exposed by PR #1130.  But since the world hasn't ended yet I went ahead and reversed that one too so that we're back to a 'stable' state.

The world hasn't ended yet, right?  Just confirming...  :)

cytosine:
Hmmm...  Ok guys, hold-up.  I tested again on another RO install and reverting the 1130 fix didn't revert the Kerbal Construction Time related bug that originally spurred this little endeavour.  Gotta run right now, so please hold-off until I can attack this again tomorrow!

jwvanderbeck:
What is the KCT bug?

cytosine:
Hi @jwvanderbeck,

Well, it's relatively minor and a bit obscure, but I believe we can fix it with little effort.  If you have RO, Skylab, RemoteTech, Firespitter, and KCT installed and you try to 'simulate' from orbit (with or without a crew) then you get a log spamming index error.  The game still runs and you can revert back to the VAB successfully, but you can't access any of the controls, even with the ASAS attached and the minimum crew value has nothing to do with it, it just simply breaks.

I have a couple ideas how to resolve it, but I wanted to test them out first.

Thanks!

ctiberious:
I just noticed a problem I missed when I originally added in the RemoteTech stuff to Skylab ( #1091  ).  When I added the ModuleRTAntenna, I just copied it from another part that I knew worked, then tested it with a Skylab I already had in orbit.  Never even considered trying to launch a new Skylab but if you do, the Antenna will break off during launch (when you reach just over 100m/s).  The problem is the "%MaxQ = 6000" line.  I'm not sure what an appropriate value is but I tried simply setting it as "%MaxQ = 60000" and was able to successfully complete the launch and get into orbit.  I don't think it has anything to do with the KCT issue you're trying to address, but still would be useful to include an updated MaxQ value with this PR.  I could create a separate PR with that change but I think it would conflict with yours.

jwvanderbeck:
You mean it broke off even while retracted?

ctiberious:
@jwvanderbeck Yes.  I don't think the code I used to setup the ModuleRTAntenna module realizes there is an animation or a deployed setting for the antenna.  Obviously, that would be a better solution but I don't know RemoteTech well enough to be sure how to correct that.

cytosine:
Hey, nice find @ctiberious this is starting to make sense...

So the MaxQ value is a measure of pressure.  It simulates if your antenna is extended during flight it should be ripped off.  I am 99% certain that the ModuleAnimateGeneric function is what RemoteTech uses to determine if an antenna is extended or not, but nobody in the RT community has responded to my post yet, and I don't blame them.  They are up to their eyeballs in v1.1 stuff which is perfectly fine by me.

Since this mod uses an alternative animation function it would seem that RT can't properly detect the antenna state so it rips the antenna off at a certain altitude/pressure.

Question for @ctiberious:  Did you notice log spamming "index error" when you launched the craft?  I'm wondering if the log spamming results only from how KCT warps the craft into space during testing.  Not that it really has anything to do with KCT...

This is great progress guys, we are *MUCH* closer to finding the root cause.

Thanks!

NathanKell:
I suggest making an issue on the RT repo, that's where they'll first check, and I can poke with a number.

ctiberious:
@cytosine No, I didn't notice any log spamming during the normal launch but I also wasn't looking for it.

NathanKell:
@cytosine how we doin' here?

PR #1138, opened 2016-04-13
Fix Soyuz Lunar Heatshield

SirKeplan:
Tested.


stratochief66:
So, you fixed the lunar heatshield to actually be lunar rated. Thanks :)

SirKeplan:
Exactly!

PR #1142, opened 2016-04-14
Proton vernier stock gimbal update

stratochief66:
gimbalRangeXP = 11
	gimbalRangeXN = 11
	gimbalRangeYP = 0
	gimbalRangeYN = 0

Are already in the base part config. With this change they will then be contained to the one dimensional motion, although raidernick might just add these to the base part gimbal module once he has a chance to do some 1.1 testing and tweaking as well.


raidernick:
@stratochief66 you will find I actually already did this to all the stock part configs, it's on git already also. I kept the stock values to the correct RO values also.

stratochief66:
@raidernick 

With this PR it should now have good function with just the stock config:
https://github.com/KSP-RO/SovietRockets/commit/1a068382714251bac18f15e3cc0355af8acbf288

PR #1145, opened 2016-04-18
Add CLS to Skylab docking port

ctiberious:
Hopefully this doesn't conflict with #1136  I noticed that the skylab docking node doesn't have a CLS module.  This should correct that.


raidernick:
@ctiberious I forgot to put this in the stock config, I usually add it to all my parts. I am going to add it to the stock config right now for the next mod release as most of the other parts have these modules in the original config already so we won't need this PR. Thanks for reminding me.

ctiberious:
No problem.  :)

PR #1146, opened 2016-04-18
Skylab ECLSS

ctiberious:
So back in #844 we updated the "CO2 Scrubber" on Skylab to be a CO2-O2 converter.  This was done primarily due to a video that was posted (which sadly seems to no longer be available) which implied that CO2 extracted from the air was them separated into C and O2.  This results in a nearly perfect O2 recovery system.  I always thought that was a bit odd because Skylab included over 2 tons of O2, but in the end I just figured NASA was being extra cautious.
In my career game I'm getting to the point where I'm looking at setting up ISS and planning my first mission to Mars so I've been researching more modern life support systems.  And one of the first things I noticed was that ISS currently dumps CO2 overboard.  I found that strange considering what the video about Skylab's ECLSS.  Well, after further research (http://pages.erau.edu/~ericksol/projects/futurspcrft/sp/intro.html), it seems that Skylab's molecular sieve system didn't convert the CO2 back into O2 (with waste C).  It just dumped the CO2 overboard, pretty much just like ISS currently does.
Since the CO2 is just dumped overboard, I've removed the "outputResources" and adjusted the Waste tank accordingly.  Now instead of recycling CO2 back into O2 (and basically never using any of the O2 that comes with Skylab), now all that the extractor does is remove CO2 as it's generated.  I did leave the same 105% efficiency that we previously had so we can recover from time warp inconsistencies.

I've also increased the electrical requirement for the extractor.  I can't find any documentation stating how much electricity was used to run the CO2 extractor, but I have to think it would have been more than 0.01EC/s (10 watts / hour).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Skylab.cfg
@@ -127,7 +127,7 @@
 			{
 				name = Waste
 				amount = 0
-				maxAmount = 391.1	//170 days for 3 crew including CO2-O2 Converter
+				maxAmount = 271.4	//170 days for 3 crew
 			}
 			TANK
 			{
@@ -140,10 +140,9 @@
 	MODULE
 	{
 		name = TacGenericConverter
-		converterName = CO2-O2 Converter
+		converterName = CO2 Extractor
 		conversionRate = 6.0	// # of people - Figures based on per/person
-		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010	// See RO Github #844 for explanation of values
-		outputResources = Oxygen, 0.0071565375, true, Waste, 0.0000027155975, true	// See RO Github #844 for explanation of values
+		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.10	// See RO Github #844 & #1146 for explanation of values
 	}
 
 	MODULE


stratochief66:
Ahh true facts. Found another search with that fact.
https://books.google.ca/books?id=_6hymYAgC6MC&pg=PA264&lpg=PA264&dq=co2+recovery+skylab&source=bl&ots=Nx58ROLJQx&sig=fATfzcZNpQ3mcYvc9IYCAQKIyK8&hl=en&sa=X&ved=0ahUKEwiQ_v3t8pjMAhUpk4MKHdpjAQ4Q6AEIGzAA#v=onepage&q=co2%20recovery%20skylab&f=false

Water is much better recycled on the ISS than on the Skylab, IIRC. For
example, air moisture as well are urine reprocessing. I don't know if we
have any TACLS modules to represent that, or how best it could/should
logically be done.

It will be a bit before I can review this PR, but the physical facts seem
sound.

ctiberious:
Water pulled from the air on Skylab was also just dumped.  It was all part of the CO2 extraction process.  Don't really have a legitimate way to show waste water specifically from exhauling and perspiration and such.  Though we could fake it by having the CO2 extractor also dump a little waste water.  But figured in the end that wasn't worth doing.
As for water recycling, I'm not sure how much recycling (if any) they actually did on Skylab.  Skylab launched with 6000lbs of water and they only estimated they'd need about 2600lbs for three astronaughts for the entire 171 day mission plan.  ISS, on the other hand, does have a pretty good system though it's not perfect.  And because they "create" oxygen by electrolyzing water, I know a lot has to be brought up every year.  Though I'm not sure exactly how much.  I do know they expect to "recover" upwards of 2000lbs/year of water (above and beyond current recycling) once the Sabatier CRA is in place.
As for TACLS, the cool thing about the "TacGenericConverter" module is you can set it up to represent pretty much anything that uses/creates resources.  Though TACLS does come standard with a couple water recycler parts that we can easily use as examples.

ctiberious:
Can't say how accurate this is, but information at this site (http://science.nasa.gov/science-news/science-at-nasa/2000/ast02nov_1/) implies that without recycling, "40,000 pounds per year of water from Earth would be required to resupply a minimum of four crewmembers for the life of the station".  Now if I could find out how much is actually brought up, I could likely figure out what values to use to setup a recycler for ISS.

Schnobs:
For the basic chemistry of CO2 scrubbers, see "soda lime". Never mind the the talk about molecular sieves, that's just a fancy word to describe material with a large active surface area. The chemistry is still the same.

Lime can be regenerated by heating it. Takes hours, but with two units taking turns you can provide 24/7 coverage. All processes I knew until now required an air (or at least gas) flow to carry out the CO2, but apparently that issue has been solved: you can heat directly into vacuum and be done with it.

The lime is hygroscopic and and the chemistry requires some moisture for it to work; venting into space not only wastes CO2 but some water as well. Compared to other places where water is lost, this one seems pretty insignificant.

--

All that said, the basic approach of just extracting CO2 at the expense of electricity is sound. Only question, /how much/ electricity? You need some for the fans though that may already be accounted for in the station modules themselves. Leaves the heating.

This is another place where TACLS isn't really suitable. I'd expect that only a small fraction of the heat is needed for chemistry while most is simply wasted. (Compare boiling eggs: most energy just heats the pot, water, and room, only a small fraction goes into actually congealing the egg. Boiling one egg or five takes nearly the same amount of heat).

My gut says something in the kilowatt range, but not running all the time. An average 200W per person, perhaps? Half as much for a large setup?

For lulz, we might as well have a system that creates lithium hydroxide from waste+power, and leave the scrubbing to the actual scrubbers.

PR #1148, opened 2016-04-22
Update RO_DeadlyReEntry.cfg

Starwaster:
This will bring g-force tolerances closer to real world data. (except that crewGKillChance maybe should be brought down if we don't want these limits to be hard limits for life/death)

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/RO_DeadlyReEntry.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/RO_DeadlyReEntry.cfg
@@ -7,7 +7,7 @@
 	@crewGPower = 5
 	@crewGWarn = 200000000
 	@crewGLimit = 300000000
-	@crewGKillChance = 0.75 // Maybe decrease this?
+	@crewGKillChance = 0.01 // 1% chance per physics frame
 }
 
 // Fix DRE bugs, remove from wings
--- a/GameData/RealismOverhaul/RO_RecommendedMods/RO_DeadlyReEntry.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/RO_DeadlyReEntry.cfg
@@ -1,3 +1,15 @@
+// RO Custom g-force settings.
+// These will peg max durations for 6g at 12 minute and 10g at 1 minute.
+@REENTRY_EFFECTS[Default]:AFTER[DeadlyReentry]
+{
+	@crewGMin = 5
+	@crewGClamp = 40
+	@crewGPower = 5
+	@crewGWarn = 200000000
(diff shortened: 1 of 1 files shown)


Starwaster:
Proposed RO/RSS DRE g-force damage tables
Third column is the time (in seconds) after which crew are eligible for death

crewGMin	5
crewGClamp	40
crewGPower	5
crewGLimit	300000000

6	7776		771.60
7	16807		356.99
8	32768		183.11
9	59049		101.61
10	100000		60.00
11	161051		37.26
12	248832		24.11
13	371293		16.16
14	537824		11.16
15	759375		7.90
16	1048576		5.72
17	1419857		4.23
18	1889568		3.18
19	2476099		2.42
20	3200000		1.88
21	4084101		1.47
22	5153632		1.16
23	6436343		0.93
24	7962624		0.75
25	9765625		0.61
26	11881376	0.50
27	14348907	0.42
28	17210368	0.35
29	20511149	0.29
30	24300000	0.25
31	28629151	0.21
32	33554432	0.18
33	39135393	0.15
34	45435424	0.13
35	52521875	0.11
36	60466176	0.10
37	69343957	0.09
38	79235168	0.08
39	90224199	0.07
40	102400000	0.06

NathanKell:
While we're talking code changes, I'd also suggest multiplying crew kill chance by timewarp.fixedDeltaTime to normalize wrt physwarp.

With those numbers, I think the kill chances will need to be very, very small; IIRC that Soyuz abort hit sustained (i.e. a good few seconds) 19G peaks, and everyone lived. Back problems, but they lived. I think those numbers make sense for GLOC though.

Starwaster:
Probability doesn't work that way though. It doesn't scale over time.

NathanKell:
Discrete steps. If each frame represents 20ms, and you have kill chance a, then if you go to warp and frames become 40ms, then you can either do the check twice or sum the probabilities, no?

Starwaster:
If something has x% chance of happening and you make two checks, that doesn't mean that the probability is x*2%, it just means you made two checks at x%. It doesn't become twice as probable.

NathanKell:
@Starwaster right, got confused re: addition. Let's try the multiplication case: Survival chance x%, make two checks, total survival chance is now x^2 % no? Probabilities can be multiplied. So I guess do (1-killchance)^(fdt / 20) = chance for that frame, if I'm mathing right?

Starwaster:
Something like that, I have to brush up on my statistical probability. But where is 20 coming from in your example?

NathanKell:
Should be 0.02 (20ms).

Starwaster:
ah right, ok.
Except what you actually want is 1-(1-killchance)^(fdt/0.02)

PR #1163, opened 2016-04-24
Add manufacturer, title, description to GEM63 config

alimoncul:
Added title, manufacturer, description lines.


stratochief66:
Possibly a bit long of a description, we will have to see how it looks in game and modify if necessary.

PR #1164, opened 2016-04-24
Add manufacturer, title, description to GEM63XL config

alimoncul:
Added title, manufacturer, description lines.


stratochief66:
@alimoncul 

We have a longer description for this part you should use, here:
https://github.com/KSP-RO/RealismOverhaul/blob/b94022f1ef4193a923b0cc2712a56d660b54a4bf/GameData/RealismOverhaul/RO_SuggestedMods/KWRocketry/RO_KWRocketry_Solids.cfg#L171

In general, I would search the engine type (in this case, 'GEM-63XL') to find existing descriptions.

To make a change on a PR, do the same thing you do to add another commit to it. So, go to your patch-13 branch and edit this file again :)

SirKeplan:
@alimoncul i noticed the same with #1163

PR #1168, opened 2016-04-24
Experiments cooldown & shielding

PhineasFreak:
NOTE: The cooldown time period for the film cameras ("Materials Science") may not be 100% perfect.


NathanKell:
Sorry to have missed this so far! I would prefer that such changes be made in RP-0, since in RO that's not even a film camera yet. Besides, in RP-0 you can do it per experiment not per part...

PR #1170, opened 2016-04-24
Add RF tags to engines

aw1621107:
Should let engines be searchable by the fuel/oxidizer/monoprop they use.

This was generated using a short Python script; if that script is preferred, I can change the PR to that.

Tags definitely aren't final; certainly missing some, and still have to figure out what needs what prefixes, if any.

Info on syntax can be found [here](http://forum.kerbalspaceprogram.com/index.php?/topic/137633-syntax-for-11-part-tagging/)

### DIFF
--- a/GameData/RealismOverhaul/RO_RealFuels.cfg
+++ b/GameData/RealismOverhaul/RO_RealFuels.cfg
@@ -9,225 +9,225 @@
 // Put fuel mixtures, etc. into engine tags
 @PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[*]:HAS[@PROPELLANT[Ethylene]]]]
 {
-	@tags ^=:$: (Ethylene:
+	@tags ^=:$: Ethylene :
 }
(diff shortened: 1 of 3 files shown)


SirKeplan:
just wondering, do we need both EnrichedUranium and DepletedUranium?

aw1621107:
I honestly have no idea, because I don't know what engines if any use them. Now that I think about it, DepletedUranium probably isn't needed anyways, because that's a waste product...

SirKeplan:
AFAIK everything that uses EnrichedUranium generates DepletedUranium, so it shouldn't be neccesary.

aw1621107:
Alright, I'll get rid of that then

PhineasFreak:
@aw1621107 The NERVA engines (and it's derivatives) use Enriched and Depleted Uranium as the nuclear fuel and waste respectively.

aw1621107:
I suppose this should be renamed "Add RF tags", because now I'm tagging things other than engines.

PR #1171, opened 2016-04-25
Created preliminary config for French Vexin engine family

Ravenchant:
WIP!

I cross-checked all parameters except for engine mass, which was nowhere to be found (Astronautix has 175 for the semi-related Vexin A used on Europa, but I don't trust it), 450 kg is a placeholder. For the ignition resource, sources mention "fantol", whatever that is. I just made it use an equivalent amount of Furfuryl alcohol for now.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Vexin_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Vexin_Config.cfg
@@ -1,0 +1,201 @@
+//  ==================================================
+//  Global engine configuration for Vexin/Valois engines.
+
+//  Inert Mass: ?, set to 450 kg
+//  Throttle Range: not throttleable
+//  Burn Time: 93 s Vexin, 118 s Valois
+//  O/F Ratio: 3.57 (Vesta, Vexin), 4.4 (Valois).
+//  Chamber pressure: 19.6 bar
+//  Ignition method: Fantol starter slug (Vexin), represented with Furfuryl
+
+//  Sources:
+
+//  http://www.b14643.de/Spacerockets/Diverse/European_Rocket_engines/engines.htm
+//  http://www.capcomespace.net/dossiers/espace_europeen/ariane/espace_francais/diamant_A.htm
+//  https://books.google.si/books?id=wRujBQAAQBAJ&pg=PR3&lpg=PR3&dq=propulsion+re-entry+physics+proceedings (search for "Valois")
+//  http://nospremieresannees.fr/lanceurs/06_diamant/lafa_developpement/lafa2-caractechniques_perinnelle/texte01.html
+
+//  Not yet used by any mod
+//  ==================================================
+
+@PART[*]:HAS[#engineType[Vexin]]:FOR[RealismOverhaulEngines]
+{
+    %mass = 0.45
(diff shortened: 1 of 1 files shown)


SirKeplan:
Minor thing, engine diameters are now meant to be in the description, not the title.

stratochief66:
@jwvanderbeck does these appear to be reasonable starting TestFlight values, or should it be left w/o Testflight config?

@Ravenchant how did you derive the TestFlight specs?

Ravenchant:
@stratochief66 It's a bit tricky since there isn't that much information available. Vesta had 5 flights, all successful, so a starting reliability of 80% or above seems reasonable to me, but it's still a very early engine, so ReliabilityEnd probably shouldn't be all that high.

Vexin flew 24 times with 5 failures, 4 of which were in the first 12 flights. Looks like I made a mistake, and ReliabilityStart should be lower- maybe around 70%? Valois flew 7 times with no first stage failures.

Ignition reliabilities are pure guesswork, though on second thought they should be higher given that the engines are pressure-fed. Maybe similar numbers as on the Aerobees?

I can remove the TestFlight stuff and just push the rest, but I doubt we'll ever get exact values unless someone goes scavenging through LRBA's archives.

jwvanderbeck:
Can you link the sources you used for the reliability? IE where it indicates flights, failures etc?

Also are those burn times right? They seem incredibly short.

jwvanderbeck:
Also you probably shouldn't be specifying the data rate here.  RO's configer will do that for you and base it on the burn time to normalize it over a number of flights.  With very short burning engines like this it will make the rate higher so that you don't need a million flights to max them out.

Unless it was your intention to make it really, really slow?  If that was your intention, I think there is another property you need to specify as well.  @NathanKell ?

Ravenchant:
@jwvanderbeck Oh, no, I didn't want the data rate to be any slower than usual. Must have forgotten to delete the line since I copied the "form" from another config. But the burn times are correct as far as this site can be trusted: http://www.capcomespace.net/dossiers/espace_europeen/ariane/index.htm (under "CARACTERISTIQUES DES LANCEURS DIAMANT" and "VERONIQUE ET VESTA"). I rounded them up a bit, IRL they were 57s, 93s and 116s respectively.

For flights I used http://space.skyrocket.de/doc_lau_fam/diamant.htm and http://www.b14643.de/Spacerockets_1/West_Europe/Diamant/Description/Frame.htm. On the page of the B-version it's clarified that the two failures were because of upper stage malfunctions.

jwvanderbeck:
Interesting engines, but yeah I see the same burn times on nautix as well.

@stratochief66 I don't have any other specific complaints.  I don't really know the specific engines enough to comment any further really.

Ravenchant:
@jwvanderbeck Mistakes can also be fixed later as no mod currently uses the config. Lionhead Aerospace is outdated, Bluedog Design Bureau doesn't yet have RO compatibility from what I see and my version isn't released yet.

stratochief66:
Hey, sorry, I hadn't checked on this PR in a bit. Yeah, as long as Agathorn doesn't have any complaints I'm good with merging. I'm still learning TF/RO integration, so I figured it was good to consult another opinion.

PR #1178, opened 2016-04-25
1.1 Heatshield updates for FASA

stratochief66:
Updating thermal parametres for FASA Apollo capsule w/built in shield. Pasted same values to the stand alone shield, in case anybody uses it.

FASA Apollo Chutes currently has a glitch that causes it to superheat during re-entry. Temp bandaid until someone smart can figure and fix that.

Leaves a narrow (~2.5km) window of re-entry for the Gemini at 11km/s (fast lunar return), so again barely lunar rated as usual.

If you slow the capsule down a bit before re-entry or use a slower return, still a good lunar return capsule.

https://www.youtube.com/watch?v=fWz0UxuZBZg


NathanKell:
@stratochief66 also needed the outputResource line or CharredAblator isn't generated. ;) I'm fixing that (and lack of charring in Gemini) in a commit--thanks for this! And copying to stock heatshields

PR #1179, opened 2016-04-25
Diameter added to F-1A, F1, F1B, GEM40, GEM60, GEM63XL Config description

alimoncul:


stratochief66:
@alimoncul 

The F-1A description has some stock references in it, ie. werner von kerman. please remove.

PR #1181, opened 2016-04-25
TestFlight configuration for 11D33, LE-7, HM-7 + minor thrust level fix

Ravenchant:
I'm kinda new to this "reliability" thing, please let me know if the values set are lower than intended given real-life track records.


jwvanderbeck:
Some of these numbers have some pretty wide ranges.  Can you cite your reference sources?

Ravenchant:
I used cycleReliabilityEnd = (number of times an engine has failed / number of times it has flown). (the early 11D33 failed a lot). If it didn't have a failure, I took (0.5 / number of times it has flown) rather than 0. The other values are pretty much placeholders. For example, the HM-7 had 32 static tests before being flown, but there's no mention how many of them were successful. The LE-7 had 8 failed static tests (or 4, if you don't consider catching fire a failure) but no mention of how many _were_ successful.

I'm not sure how IgnitionReliability should be calculated, the TestFlight wiki had no info...

NathanKell:
Reliability end should describe the reliability of the engine _after_ all flights. The average reliability, which should be calced as num_success +1 / num_total +2, should be somewhere between start and end. The start reliability should reflect the reliability of the first few launches; the end reliability should reflect the last few launches.

NathanKell:
@Ravenchant ?

Ravenchant:
@NathanKell Right, sorry. I was mostly away from the computer for a few days. Will fix ASAP.

Ravenchant:
I've rechecked the values (also gone through Wikipedia's exhaustive R7 launch log which lists the causes of failure), they should be closer to reality now. One possible issue is the burn time of early S.15400 versions, in that it's hard to know if it was limited by the engine design or simply by the amount of fuel block-L could hold.

NathanKell:
The starting reliability of the 11D33 and RD-58 configs (all of them) still seem shockingly low. Was it really the case that controlling for all other factors, 1/4 of them failed to ignite and of those that did ignite another quarter failed?

Also I'd suggest using a tech transfer of 50 between subvariants (11D33->M and RD-58->other 58 models) and saving the 25 for 11D33M->RD-58.

Ravenchant:
Hey, only wrote the first two, the rest were already there=p But apparently the first few Molniya flights went as follows: stage 3 guidance failure, stage 3 ignition failure, S1.5400 ignition failure, success, S1.5400 failure, S1.5400 failure, stage 3 exploded, S1.5400 stage "failed in orbit", success, S1.5400 stage damaged and couldn't be restarted, TLI failure, S1.5400 stage gyro failure, success, S1.5400 failure...
Discounting the failures caused by guidance and the rest of the rocket, a 1/4 starting failure rate still doesn't seem that high. Maybe I could add a data gathering multiplier so it decreases faster than usual?

NathanKell:
Wow, those are indeed some sobering records. In that case I guess it sounds about right. A lot of those failures aren't engine related so I guess it comes out to 1/4.

PR #1189, opened 2016-04-26
Diameter added to RD57, RD58, RD855, RD856, RL10, RL60_Vinci, RS68 Config description

alimoncul:


SirKeplan:
the RD-855 and RD-856 attach radially(and are much smaller individually than stated IIRC), does it still make sense to state a diameter for parts that attach on the side?

jwvanderbeck:
I think so yes.  Be nice to know that a radially attached part is 5m when you're building a 2m rocket.

SirKeplan:
@jwvanderbeck yes, bit in this case the RD-856 is only about 0.3m maybe it's the main engine that is 3.65m

alimoncul:
@SirKeplan  does all that engines 3.65m ? - http://www.astronautix.com/engines/rd855.htm -

SirKeplan:
OK, It seems the real life RD-855 and RD-856 are a single vernier engine with 4 combustion chambers, where as in RO we have modelled each combustion chamber as an engine (essentially divided the engine into 4)

So here the over 3 metre diameters are not appropriate for how we have these engines modelled. until there are other parts that model this accurately, best to just state the rough diameter of the parts as they are (around 0.3 m IIRC)

alimoncul:
@SirKeplan okey i changed to 0.3 meters

PR #1190, opened 2016-04-26
Diameter added to RS88 Config description

alimoncul:


SirKeplan:
Possibly the model size is wrong, but ingame the part is nearer 1.9m

PR #1191, opened 2016-04-26
Diameter added to RSRMV, RSRM, SSME Config description

alimoncul:


SirKeplan:
is this right? i'd expect the RSRM and RSRMV to both be the same diameter? typo?

stratochief66:
I would also expect the to be the same diametre, but the base nozzle flare could be bigger on one than the other?

alimoncul:
@SirKeplan It was 3.6m on default config. It is 3.7m i changed it.

PR #1199, opened 2016-04-26
Make sure AJ10_138(Titan III stage) title/description not overwritten

SirKeplan:
Description was added twice on AJ10_138, removing one.
Not tested yet as it's 1.1 branch


stratochief66:
Sooo, then no title will be written for the FASA trans-stage, and no other part that might be set to use the AJ10_138_Config gets title, manufacturer, and description set? I'm confused.

SirKeplan:
@stratochief66 title, manufacturer and description are set here https://github.com/SirKeplan/RealismOverhaul/blob/AJ10_138_FIX/GameData/RealismOverhaul/Engine_Configs/AJ10_138_Config.cfg#L5-L7
the FASA trans-stage now overwrites manufacturer and description, as it's a stage not just an engine.

NathanKell:
Now that RF's out, will wait on testing before merging :)

PR #1201, opened 2016-04-28
Mercury heatshield 1.1 tweak

stratochief66:
During a 'standard re-entry' from a 250km x -314km orbit, 52 of 125 ablator remain and a peak skin temp of 1950K is achieved.

In my test, using these specs for the mk1 capsule results in a peak skin temp of 1780K is achieved and 116/200 ablator remains. Given that both of these capsules will start to burn at ~2150K, the mk1 is left with much more temperature overhead, so it has a much easier re-entry.

This PR should wait on confirmation testing that the values achieve the same or very similar results for someone else. At that point, please let me know and I will go adjust all the other values that should be changed as well.


NathanKell:
I redid this from perigee (performing retrofire as of MA-6, rather than at another point) and found the existing 2.9 convection factor far too high. I am testing 1.73 which should yield ~80W/cm^2. This is necessitating retuning ablator pyro flux in order to get ~1800K max shield temp.

NathanKell:
Ok, got the peak flux to be about right, and shield temp is ~1750K, so that's fine. A reentry from apogee (as we were originally testing, oops--turns out all Mercury flights fired retros at perigee) raises peak to ~915kW/m^2, but at least it keeps total heat load in proportion; perigee reentries have a _higher_ heat load than apogee ones. Skin peak temp hit ~1850K, so sounds like a decent ballpark.

stratochief66:
K; just let me know when the values are settled, tested and you're happy with them, and I will move them to the other applicable places.

https://github.com/KSP-RO/RealismOverhaul/commit/7c69aee2664e86d70c2207bd5fa2cd861247d0b8

NathanKell:
Sorry, wasn't clear. That commit has values with which I am comfortable.

stratochief66:
I figured. but 100% certainty is > 95% certainty :)

stratochief66:
Done, I believe. All parts that take the same heatshield values as Mercury now have the new NathanKell approved values.

PR #1204, opened 2016-04-29
Vernier gimbal added to RD855, RD856 - TEST

alimoncul:
@PhineasFreak gave me a tutorial about gimbal config editting


NathanKell:
Uh, do these actually have 1 degree of gimal in the X plane, or did you want to turn gimballing off for the X plane?
Also when you specify the X and Y you don't need the normal `gimbalRange` line up top.

PhineasFreak:
A minimum of 1 degree of gimbal was used just to make sure that nothing will break (expecting any other value than 0) but the KVD-1 test indicates that no such problem exist.

alimoncul:
I was on the camp but im back :). I will change it to 0.

PR #1205, opened 2016-04-29
KVD-1 Vernier Gimbal Fix

alimoncul:
I tried to fix it but im not sure i did. Need more help i think :)
Phineas Freak: KVD-1: Just reverse the X-Y axes (these are used for roll control).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1973,9 +1973,14 @@
 	// The Verniers.
 	@MODULE[ModuleGimbal]
 	{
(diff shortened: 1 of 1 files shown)


SirKeplan:
Shouldn't the gimbalRangeX and gimbalRangeXN be set to 0 not 1, if they only gimbal in 1 plane?

stratochief66:
@SirKeplan possibly. But we will be waiting to be able to test this in-game anyway, and both 0 and 1 should probably be tested. There may be odd or unexpected behaviour with it set to 0, so I think both are worth testing.

PhineasFreak:
From a first look, setting the axis gimbal range to zero has no obvious ill effect (at least for the RD-855/856 engines).

stratochief66:
I haven't seen Alim around in a bit, and 1 ~= 0 in this case. So I think it should be 'good enough' (TM!) for now.

PR #1216, opened 2016-05-06
SSTU fixes/new part support

SirKeplan:
WIth this PR SSTU should load fine in RO. possibly more part tweaking to come.


NathanKell:
@SirKeplan let me know when ready here.

NathanKell:
Ah, it's gotta get retargeted at master anyway, so merging down.

SirKeplan:
@NathanKell That's fine.  I think I just had done everything for this PR anyway.

PR #1218, opened 2016-05-07
CST-100 pack updates/fixes

PhineasFreak:
* Fix a wrong model file path for the heat shield model.
* Update heat shield stats to be compatible with the new changes of the reentry heating system (comparable to the LEO - rated 5 meter heat shield).
* Fix the attachment node position of the NDS protective nose cone.
* Increase a bit the dry mass of the command module.
* Increase a bit the electricity consumption of the command module to be in line with the required keep - alive requierements.
* Decreased the amount of electricity carried by the command module (the service module is always required).
* Decrease a bit the range of the RemoteTech antenna.
* Include all life support resources within the command module so not to have any mass disparencies if the user does not have TACLS installed.
* Keep the electricity consumption of the command module to a fixed value if the user has not installed RemoteTech and/or TACLS.
* Remove the original RS-88 engine configuration (Ethanol & LOX) from the available engine configurations.
* Decrease the dry mass of the service module and increase the carried battery storage (the built - in solar array should barely keep it operational for approximately 60 hours while in autonomous mode).
* Fix the excessive mass of the parachute pack if RealChute is installed.
* Fix the scaling of the active NDS port (model was probably replaced) and it's attachment node.

NOTE 1: The nose cone still has issues with it's collider (nothing we can do on our end).

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CST/RO_CST_Aero.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CST/RO_CST_Aero.cfg
@@ -2,7 +2,7 @@
 //  CST-100 main heat shield.
 
 //  Dimensions: 4.65 m x 0.9 m
-//  Gross Mass: 930 kg
+//  Gross Mass: 1200 kg
 //  ==================================================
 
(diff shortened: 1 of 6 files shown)


stratochief66:
@PhineasFreak thanks!!

PR #1227, opened 2016-05-08
updated existing SSTU plume for KSP 1.1.2

Temeter:
Updated partnames to fit current SSTU version.
Note, these are only the LFO engines that where already patched by RO. There are still a number of engines without patches.


SirKeplan:
Just an observation, as the files where already this way, some use `:FOR[RealPlume]:NEEDS[SmokeScreen]` and others use `:BEFORE[SSTU_ENGINES]:FIRST`

for consistency these should probably all be the same, i think `:FOR[RealPlume]:NEEDS[SmokeScreen]` is preferred?

stratochief66:
There was a time when it mattered (had to be FIRST) in my experience, because of the way SSTU engines were created, then how the RO engines were created. I believe the FOR:NEEDS should work now (But Should Be Tested In Game!) since the SSTU engines are created differently these days.

Temeter:
I tested all modified engines ingame, they seemed to work just fine.

stratochief66:
@Temeter could you please change all the plumes you've modified here to use ":FOR[RealPlume]:NEEDS[SmokeScreen]" instead of ":BEFORE[SSTU_ENGINES]:FIRST" ?

That second bit is only used due to old, odd circumstances and should be replaced. Please confirm they still work with that change, of course.

Temeter:
Ah, so that's how it's supposed to be? Sorry, no clue in what direction stuff is being loaded/modified.

stratochief66:
@Temeter 

No worry, and thanks for the PR. It was one of those many things I knew needed doing, but I didn't have the interest in re-doing, again. :p

There were a few times in the past where SSTU changed how it made clusters of engines in code, and we had to change how they were patched or remade and how the plumes were patched. We left it quite unclear which was the proper way. Thanks @SirKeplan for catching that.

A PR should Always be tested in game by whoever makes it (uber-rare exceptions, like a description spelling mistake). Then, NathanKell, SirKeplan or I will also test it in game as a second pass to ensure nothing was missed in the original before Merging.

PR #1230, opened 2016-05-09
added scaled plume to new SSTU engines, please test if ok

Temeter:
A number of new engines which had not yet RO plumes. Took stock SSTU plumes, adjusted them ingame to RO scale. 
I didn't create RealPlumes before, please check them and give me feedback if they are ok. Orientated myself on existing plumes and to some degree real videos.

For now, the file only contains AJ10-137/190, RD-107a/107x/108, and RD-0110. Gonna create the others too, but I'd like some feedback first.

Note, I'll keep this seperate form the other PR, since this is actually new stuff and not just fixed partnames.

### DIFF
--- a/GameData/RealismOverhaul/RealPlume_Configs/SSTU/New_SSTU_LFEngines.cfg
+++ b/GameData/RealismOverhaul/RealPlume_Configs/SSTU/New_SSTU_LFEngines.cfg
@@ -1,0 +1,142 @@
+@PART[SSTU-SC-ENG-AJ10-137]:NEEDS[RealPlume]:BEFORE[RealPlume]
+{
+    PLUME
+    {
+        name = Hypergolic-Apollo-SM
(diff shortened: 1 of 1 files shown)


Temeter:
So, updated all SSTU plumes of the RO compatible LFO engines. Tuned and checked them ingame, seemed good to me.

SirKeplan:
Looks good. except the RD-0110, the verniers plumes are appearing larger than the main plumes for some reason.

Temeter:
Thanks, I'll take a look at those two!
Good eye on the RS-68, I think i've got a clue what's that about. Wanted to shift them slightly because they weren't lined up perfectly with the model, might not work for every part.
Gotta check RD-0110 tho, all those quad-engines made my head spin.

Temeter:
Yeah, misunderstanding of relative numbers on the RS-68, and confused upper/lower+shifted decimal on RD-0110. It's fixed now.

PR #1232, opened 2016-05-09
Egc ksp11

james3838:
Verified can assign group ids to different engines in VAB
Verified can manage engines by group id in flight


james3838:
No idea why, but diff is showing entire files being changed.  This a windows CRLF issue?  The changes are actually quite smaller.

PR #1237, opened 2016-05-11
Add reasonable solar panels to Vanguard probes

leudaimon:
The Vaguard Probes were with extremely high energy production. I added a production of half the GRAB-1 (as per https://en.wikipedia.org/wiki/Vanguard_1#Spacecraft_design consumption from the transmitters was 15mW - still much lower than that) as a placeholder. If actual values are available, feel free to correct.


leudaimon:
I just noticed that the Explorer 6 also has 160W, which seems wrong. I couldn't find any reasonable values for it though.

PR #1238, opened 2016-05-14
Fixed part ID

Ravenchant:
I like wild cars as much as anyone, but it broke engine support.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/KWRocketry/RO_KWRocketry_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/KWRocketry/RO_KWRocketry_Engines.cfg
@@ -611,7 +611,7 @@
 	massOffset = 1.35
 }
 // ##########################################################################################	KW Rocketry Wildcat-XR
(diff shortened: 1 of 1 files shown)


Temeter:
cats > cars, as every true man knows

NathanKell:
@Ravenchant did you check the base KW files? There are some typos in KW itself, and this may be one of them IIRC.

Ravenchant:
@NathanKell KW did have a typo, but it got fixed by the community patch a day or two ago:)

NathanKell:
Ah, ok. Welp, guess they didn't care about save compatibility?

SirKeplan:
that's crazy... that's not the kind of typo that's meant to be fixed.... :P

NathanKell:
Exactly.

Temeter:
Ohhhhhh so that's why KW's RD107/8 didnt TestFlight correctly! Totally forgot the big one was also called Wildcat.

NathanKell:
Yes, that would explain why that and everything else broke. Jeepers, _why_ do this.

PR #1239, opened 2016-05-14
SSTU RD-107/8 RS-68 fixes

SirKeplan:


Temeter:
So is this why they were so hard to control? Anyway, thanks for the addition!

SirKeplan:
@Temeter probably, the KSP 1.1 update introduced the option of per axis gimbal ranges. which SSTU takes advantage of. as this wasn't set in RO yet the verniers reverted to a low range for gimbal :P

Temeter:
Yeah, was really weird, especially because the animations were sometimes inconsistent. Nice to see it wasn't just my piloting sucking. ^^'

PR #1244, opened 2016-05-19
more SSTU plumes, generic AJ10 configs, M1 name fix

Temeter:
1. Added SSTU plumes for the last few engines.
2. Cloned the AJ10-190 for use of the generic AJ10 configs.
3. Fixed the name of the M1.


Temeter:
Hm, I wonder if I could figure out how to make the new AJ10s start with certain fairings to add some optic diversity (and not 4 times the same engine).

Everything else is done, just needs to be tested and potential issues be fixed.

stratochief66:
I believe by the Ro/RealPlume standard, each plume should have its own .cfg
file named the same as the part itself. So, some of the SSTU engines that
changed names should probably have their .cfg file names changed as well,
but that can be a pain on Github's web interface so if you can figure it
out, that is fair.

For new engines,like some that you add here, could you please modify them?

For example, part name and file match:
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RealPlume_Configs/FASA/FASAApolloLFEF1.cfg

Temeter:
Of course I tested that stuff, countless times! :smile: 
Had to figure out the config stuff mostly by myself. The only thing that's left to look up is really just if a minor adjustment on the M1 plume worked out well.

I originally created a different file, to make sure my efforts at fixing the plumes were consolidated in a single place, in case i'd break stuff.

Temeter:
Moved the engines. Not that hard when using github app, when you figured out the UI. ;)

Temeter:
Everything should be fine now. If you wanna to merge, the stuff is ready.

Gotta figure out attachment modules at a later date and do more with them.

SirKeplan:
@Temeter you might want to scale down some of the AJ10s(using rescaleFactor), i think it makes sense for the AJ10 Early atleast, seems like most of them had a diameter of 0.84 m

http://www.astronautix.com/engines/aj10118.htm
http://www.astronautix.com/engines/aj10101.htm

Temeter:
@SirKeplan I can give it a try. That diameter is the nozzle?

SirKeplan:
@Temeter ah yes, that looks like the right way to rescale. I forgot SSTU was different. 
Normally go with nozzle diameter yes, unless the engines are wierd shapes which these aren't

Temeter:
Everything done! AJ10 variants are their correct size now.

Seems like the SXTAJ10 are actually far too small.

Temeter:
Plox merge? Idk, this pretty much done as far as i can see. ~

PR #1245, opened 2016-05-19
SSTU Pod updates

ctiberious:
I know most of the Lander components have been (at least temporarily) removed from SSTU due to compatibility problems with KSP 1.1, but the three pods do still exist.  I decided not to remove the missing parts from this file as I'm guessing they will eventually be returned.  All I'm doing here is updating the three lander pods so they are usable in RO.  These changes primarily adjust the size/position of the RCS thrusters, docking port and attachment points.  I've also removed the static RESOURCES and replaced the SSTUVolumeContainer with ModuleFuelTanks.  I left the SSTUModelSwitch from the base files in place though they are currently just visual changes as I don't know of any way to have ModuleFuelTanks have it's volume altered on the fly.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Landers.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Landers.cfg
@@ -918,20 +918,45 @@
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


ctiberious:
Creating a new, consolidated PR for all the pod and tank updates I've got.

PR #1246, opened 2016-05-19
Global Configs SXT

SirKeplan:
SXT engines now use the engine configs properly
Change NERVA II stats

Why is the `SXTCastor30` a Star 17? the model ingame isn't even the right shape.
 https://github.com/KSP-RO/RealismOverhaul/pull/1246/files#diff-c67bb5d231d1b37a8e2597a692758360L727


Temeter:
Since you're modifying the files already, can you please normalize the generic AJ10 names? Early and Mid are named 'AJ10 Upper Stage Engine (Early/Mid)', while advanced is called 'AJ10 Series [Advanced]'.

I think they should be the same. For comparision, the generic RL10 is called 'RL10 Series Vacuum Engine'.

SirKeplan:
Good point, what I'll probably do is remove 'upper stage engine' bit etc. The engines names should simply be the designation plus 'series' if its a set of engines, other info should go in the description. It's a standard that was roughly agreed upon, but not yet fully in place.

Sent from my Samsung device

-------- Original message --------

Temeter:
Nice, sounds the best to me.

Also, you're probably planning to do it already, but please change the brackets to round () brackets, those angular ones look a bit out of place in the ingame UI.

NathanKell:
It's a Star 17 because we already had Castor 30s ingame, but nothing in the Star 17 range. (Its SXT ingame name was PALM-D, so there's that too...)

PR #1248, opened 2016-05-20
RO-KSPedia

dxdy-name:
Adds KSPedia to RO. So far three pages: the main page ("Realism Overhaul") has the mod description from the forum OP, a page "False Lessons" copied from the wiki and a page on "Ullage" (figuring that that's going to be one of the first problems people are going to run into)


jwvanderbeck:
I think for the purposes of review it would be helpful to include a screenshot of how each page looks in game.

Also shouldn't there be an .xml file? I can't recall.

dxdy-name:
@jwvanderbeck I didn't think that was neccessary since it's a matter of seconds to copy the file to the game. But you're probably right, [here are the images](http://imgur.com/a/uw6aL)

Re: .xml: As far as I can tell the deployment file (ROKspedia.ksp) is some kind of archive that bundes all the prefabs + the images + whatever xml is needed on top of that.

jwvanderbeck:
> since it's a matter of seconds to copy the file to the game.

Plus 45 minutes to start it up :)

dxdy-name:
To be fair, if you're testing / inspecting wiki pages there is no reason at all to use RO. Not even Stock. Since the files are completely separate from the actual mod, you pretty much needn't even have most of stock, just enough to load the game. So about the load time of stock, or if you happen to have a even more minimized install around about 10 seconds.

SirKeplan:
Image 3: Shouldn't it show an actual ullage motor in that image(like the Saturn V ones)rather than the Altair?

dxdy-name:
@SirKeplan I was more focussed on the "not subject to ullage" part of the description since that's the relevant part of the image for the purposes of ullage. But you're right the Altair wasn't a perfect choice. I've cropped the image (so one doesn't see which engine it is) and added a red rectangle to emphazise the relevant section of the description.

SirKeplan:
@dxdy-name nice, i think that'l make things clearer.

Issue #735, opened 2015-12-07
Heatshield category placement


PhineasFreak:
The stock heatshields are placed in the "Aero" category while the DR heatshields are placed in the "Structural" category. What is the reason behind this? Or is it just an oversight?

stratochief66:
Oversight, I think. I just mentioned it to @NathanKell between yesterday and today. I'm unsure where we want them, @NathanKell ?

NathanKell:
DRE, from the beginning, has placed them in Structural. KSP 1.0, when it added stock shields, placed them in Aero. Pinging @Starwaster as well here.

I don't particularly care which category, but I agree it should be the same for both stock and DRE.

SirKeplan:
Aero makes most sense IMHO, Either that or maybe give heatshields their own category.

stratochief66:
@NathanKell @SirKeplan ping again, now that we're working toward 1.1

Heatshields into Aero or Structural? potentially @alimoncul would be able to go through and check them all if we settle on a choice.

PhineasFreak:
OK then, move them into Aero?

SirKeplan:
@PhineasFreak you are go to move them to Aero if you wish. NathanKell says it makes sense.

PhineasFreak:
Since the PR has been merged there is no reason to keep it open. Closing it now.

PR #1252, opened 2016-05-27
DRE heat shields category

PhineasFreak:
* Move the DeadlyReentry heat shields from the "Structural" category to the "Aero" category for better organization.

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/RO_DeadlyReEntry.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/RO_DeadlyReEntry.cfg
@@ -789,6 +789,18 @@ RESOURCE_DEFINITION:NEEDS[!DeadlyReentry]
 	%maxTemp = 2900
 	%emissiveConstant = 0.85 // don't want too much or too little
 }
+
+//  ==================================================
+//  Heat shield decoupler (0.625 m).
+//  ==================================================
+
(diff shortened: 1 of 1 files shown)


PhineasFreak:
As requested by @SirKeplan: https://github.com/KSP-RO/RealismOverhaul/issues/735

Zarbizaure:
Cool move! I suggested it somewhere on the RO github some days ago, it makes sense to have all the heat shield at the same place.
And: any idea on how we could make the procedural heat shield work with the new charred Ablator system?

PhineasFreak:
Fixing the heat shield parameters for the new heating model is not a difficult task but setting the correct values for the mass curve and the Ablator per area might be.

Zarbizaure:
It would be cool to have a single procedural heat shield with (or three) modes (with a switcher conttrol): LEO-rated, lunar-rated and Interplanetray-rated (a bit more resistant than the lunar one, just to be sure you can survive a mars or venus return).
Yet I have no idea how to do that.

PhineasFreak:
I made some changes and adjusted it to match the DRE LEO - rated heat shields. This fixed the Ablator amount but it seems like it is not possible to actually add the CharredAblator without some PP code changes.

The switcher must be implemented by PP, nothing we can do from the RO side.

PhineasFreak:
Small update to remove the 0.625m heat shield decoupler from the available parts (the DRE heat shield decouplers support TweakScale).

PR #1253, opened 2016-05-29
(Rough) support for Bluedog Design Bureau probe cores and avionics

Ravenchant:
RT values are mostly guesswork, but otherwise the parts are usable.


NathanKell:
Cool! Can you also do the RP-0 side changes so they get the avionics bits?

Ravenchant:
This weekend :) One question- do we have any resources on how to set packet sizes and resource costs for antennas besides eyeballing it?

PR #1256, opened 2016-06-02
Update netkan

aw1621107:


NathanKell:
1. Filter Extensions was moved from rec to suggested based on it consistently causing issues due to caching and due to hiding unbought parts in the VAB/SPH. I got tired of the support headache..
2. VSR is essentially required. Why move it to suggested?

The removal of dead mods is fine, though, as is the url update.

aw1621107:
1. Yep, just heard from @jwvanderbeck 
2. Make the list match the forum post?  Nothing beyond that.

jwvanderbeck:
VSR is required for RP0 but why would it be "essentially required" for RO?

SirKeplan:
VSR being recomended for RO makes sense to me, seriously improves the look of a lot of parts, and adds a few commonly used parts too.

jwvanderbeck:
Sure but it is just a part pack, and those are all suggested not recommended.  What makes this one special?  I'm not saying it can't be, I am just asking.

NathanKell:
Because the scaling and engine choice of the stock engines is designed around the assumption of VSR being installed. Felger and I discussed this back in the mists of time and decided to work from that assumption rather than keeping it as mere suggested.

Zarbizaure:
VSR is good for those who choosed to play with stock engines but they are other people...
Pesonaly I play with SSTU and some other part pakcs, and I make little to no use of VSR engines...so I'll probably delete them.
But, you're right, VSR still improve the look of many other parts, not only engines.

PR #1258, opened 2016-06-02
RO Squad Aero updates

PhineasFreak:
* Add RO support for the inflatable heat shield.


NathanKell:
My understanding is that FAR does not support it, and therefore we do not either...

jwvanderbeck:
Did porkjet not get that collider fixed? I remember talking to him about it.

NathanKell:
Not in 1.1.3 to my knowledge.

PhineasFreak:
I thought that it was supported...well, forget it then.

NathanKell:
Sorry! :(

PhineasFreak:
Hmm, could we include it as a patch for a future version but tag it as WIP or NON RO?

PR #1259, opened 2016-06-02
RO SXT Command updates/fixes

PhineasFreak:
* Add RO support for the new probe core (Probodobodyne QUBC "Warbler").
* Fix some typos.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Command.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Command.cfg
@@ -210,6 +210,46 @@
     }
 }
 
+//  ==================================================
+//  Alouette probe body.
+
+//  Remote Tech compatibility.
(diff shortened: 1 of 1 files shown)


NathanKell:
Is it a nearly spherical solar panel coverage? If so, does it use solar panel type SPHERICAL?

PhineasFreak:
This probe core is a "hacked together" one: it does not have built - in solar panels but it uses instead MODEL{} nodes to add 4 stock surface - attachable ones (zoom inside the part while in the editor(s) and you will see). I had no idea how to correctly set up these panels but i assume that the spherical model could work?

NathanKell:
Ah, yup! A single MODEL'd in solar panel, with SPHERICAL. The transform direction shouldn't matter, just position.

NathanKell:
How is this coming? I'd really like to release a final RO before 1.1.3 comes out.

PhineasFreak:
I did not manage to get it to work...

PR #1263, opened 2016-06-03
RO VSR Science updates

PhineasFreak:
* Add RO support for the dish antennae.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Science.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Science.cfg
@@ -17,9 +17,9 @@
     %scale = 1.0
     @rescaleFactor = 1.0
 
-    @title = Communotron 32
+    @title = Communotron DTS-M2
     @manufacturer = Generic
-    @description = A retractable, short range parabolic antenna for high speed telecommunications. Effective range 145 Gm, maximum data rate 1024 kbit/sec, gain 23 dBi.
+    @description = A retractable, short range parabolic antenna for high speed telecommunications. Effective range 170 Gm, maximum data rate 1024 kbit/sec, gain 23 dBi.
 
     @mass = 0.015
     @crashTolerance = 10
@@ -91,7 +91,7 @@
 
     @title = Comms DTS-M5
     @manufacturer = Generic
-    @description = A fixed Cassegrain parabolic antenna for general applications, featuring a protective cover for the reflector. Effective range 500 Gm, maximum data rate 768 kbit/sec, gain 38.5 dBi.
+    @description = A fixed Cassegrain parabolic antenna for general applications, featuring a protective cover for the reflector. Effective range 380 Gm, maximum data rate 768 kbit/sec, gain 38.5 dBi.
 
     @mass = 0.025
     @crashTolerance = 10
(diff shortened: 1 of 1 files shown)


NathanKell:
@Schnobs @SirKeplan IIRC you're the folks who've done antenna work in the past, these fit well?

Also, @PhineasFreak can you rename the Comm32 to something else? RemoteTech has had a Comm32 for ages, and it'd be confusing if we have two. :)

PhineasFreak:
Missed the name, you are correct! BTW, the balance point for the data rate transmittion is only the "PacketSize" (in kbit/s). I did that in purpose to simplify the balancing against all other antenna parameters (standalone / data transmittion power consumption, range).

Oh and the cone sizes have been calculated as if they were real life antennae (using the X Band on the large ones, the S Band on the smaller and the physical reflector diameter).

Schnobs:
Just looking at the numbers nothing strikes me as truly odd. Whether power demand is alright depends on techlevel I guess -- I made the stock transmitters such that transmitter+probecore demand about as much as the real-life probes did. Incidentally, putting avionics in standby mode was not a standard feature. I can't speak of 1980 and later probes, but up to Voyager transmitters were always active, if only to report that the probe was still alive.

Gameplay-wise, I think we'd need:
a) a high-bandwidth transmitter for LEO. Omnis are low-bandwidth, you don't want to use them with scansat or labs. Dishes in low orbit are a pita. I was thinking of a relatively low-range dish with such a wide angle that it might almost be an omni, perhaps 130-140 degrees. Would any of the new ones suit that role?
b) a less inconvenient solution for Mars and Venus landers, where a large fixed dish looks out of place and retractable ones are unusable during descent. Currently one pretty much has to bring a relay probe. I see no obvious solution, though.

And, hmmm. Shouldn't the LongDeployable be omni?

PhineasFreak:
No, not standby (i might have worded it poorly) but for transmitting just the carrier frequency to know that the probe is "alive". Transmitting data back will require more power because you have to convert the data and modulate that carrier with them. For example Voyager had two transmitters, each one required ~25 W for "standby" operation and another ~25 W while transmitting data.

I think that RemoteTech has a KR parabolic antenna that is configured for high speed, low distance communications. Seeing that VSR is something more than a suggested mod i could suggest a small change: convert the small Communotron DTS-M1 boom dish into a "Ranger-like" dish (lower frequency, lower range, high data rate). Calculating the parameters for an antenna like that gives:

```
Reflector diameter: 0.48 m
Operating frequency: UHF (950 MHz)
Beam width: 20°
Data rate: 4.096 Mbit/s
Gain: 11.6 dBi
```

RemoteTech does not make our lives easier with it's limitations: currently if you want a directional antenna for LEO you have to explicitly point it to a station or use a helical antenna.

BTW, Bevo has a video with a Venus lander equipped with a dish so it is not difficult to achieve that.

@ctiberious the "LongDeployable" is a boom parabolic antenna, just like the stock DTS-M1.

NathanKell:
One other request: can you add the actual maximum range (vs the DSN) to the description, as the existing antennas and dishes have? it's the sqrt thingie, but remember that antennas are clamped to a max of 100x their stated cfg Range, and dishes 1000x (so a 5Mm internal-range antenna would have a max of (the actual calculation, or 500Mm, whichever is smaller).

PhineasFreak:
@NathanKell I have already done this: https://github.com/PhineasFreak/RealismOverhaul/blob/bf370595ad6f46bf037b532481baad215fb1d713/GameData/RealismOverhaul/RO_SuggestedMods/VSR/RO_VSR_Science.cfg#L48

NathanKell:
Ah!!! Didn't see you did it in a later pass. Apologies!

Issue #369, opened 2015-06-23
Maximum Temperatures!


NathanKell:
We need to figure out what appropriate internal and skin maximum temperatures are.

For heatshields, this is simple enough. It's even pretty simple for capsules. But we need to figure it out for other things (engines, tanks, landing legs, solar panels, the works). Because right now RO basically applies a flat 800C (1073K) to all parts, then customizes them a bit (high max temps for capsules and engines and nosecones). But short of the heat shields and capsules we've already done, that really needs a rework, *especially* since 1.0.3 separates between skin and internal temps.

gruneisen:
@NathanKell, I think I can attempt to wade into this and throw out the possibility of ~200°F. Working in the aerospace and the space industry, I can say that many compnents and systems internal to a vehicle are designed to this limit unless otherwise specified by the prime contractor (like its exposed to the engine plume or located in an area exposed to cryogenic temperatures).

This general rule could be applied to various components: take landing legs, for example. If we assume they are actuated electro-mechanically and use COTS components (as most systems do), then the maximum internal temperature the electronics would be qualified to would be ~165°F. This may be an option for any components which consume or store electricity.

Engines, it would seem logical to set the maximum internal temperature to the adiabatic flame temperature (though it would seem to be somewhat conservative, as the combustion chambers are not adiabatic) of the propellans. These can be calculated or found in reference materials.

SRBs are a little bit of a different story. I would start by examining what level of detail would be included. Are they considered to have a thrust termination systems? If so, the maximum operating temperature of these components is usually <300°F and most commonly they are qualified to +200°F. Do they contain a thrust vectoring system? If so, then the maximum temperatures would have to be lower to accomodate the avionics and hydraulic components of the TVC. If none of these subsystems are to be considered, then the maximum internal temperature could be set to the auto-iginition temperature (or activation temperature) of the propellant. For composite propellants like that in the Shuttle and SLS SRBs, that would be ~700°K. I'll do some more digging into this to find better numbers.

Solar panels, being un-insulated, are most likely going to have specialized electronics that can handle the extremes of space temperatures (reference:http://science.ksc.nasa.gov/pao/faq/faqanswers.htm) so it would make sense to set them to +250°F.

[...]

PR #1264, opened 2016-06-03
RO VSR Fuel Tank updates

PhineasFreak:
* Add RO support for the Oscar - A propellant tank.
* Add temperature modifiers.
* Simplify resource removal.


NathanKell:
I'm a bit confused--why set these temps explicitly when other tanks don't and get the usual 1073K (800C) max temp?

PhineasFreak:
I find that 800 °C are higher than most materials can handle (and especially propellant tanks). No reason other than that (you can ask for veto and i will return the values to the default ones).

PhineasFreak:
@NathanKell i found an old issue for the maximum temperatures (https://github.com/KSP-RO/RealismOverhaul/issues/369) and you seemed to support lower temperatures for many parts. Specifically, you mentioned part/skin temperatures in the range of 600/900 and @stratochief66 also wanted it as a 1.1 feature.

NathanKell:
@PhineasFreak I would definitely support such a change. Though I suggest doing so in RO globals to set the base maxtemp to 923 (650 C) and then changing temps where it should be different.

I am unsettled by patching by category; I would be more comfortable if we instead set up a global patcher where we, e.g., said globalTemp = Aluminum or = Titanium or whatever. Or globalTemp=StructuralLow. Whatever, just something that we are sure we are setting rather than doing a category == foo check, which can often be wrong (LES needs high max temp because it's the tip of the rocket).

PhineasFreak:
OK, i started working on a "Global Materials Config" that will include configs for both the thermal (part & skin temperature) and the physical parameters (breaking force & torque). I have some problems getting the values from inside blocks (like the TF configs) but i am sure i will find the way.

NathanKell:
Sounds good! For now I'll merge this in and temps can be tuned later.

PR #1265, opened 2016-06-03
RO SXT Utility updates

PhineasFreak:
* Add RO support for the space station service module tanks. Compatible with the PPD-4 and PPD-6 Crew Cabins.


NathanKell:
Same temperature question as before. Also, uh, hope I didn't miss that in prior PRs I just merged...

PR #1266, opened 2016-06-04
RO Squad Wheel updates

PhineasFreak:
* Fix the electricity consumption of the wheels
* Tweak the electricity consumption of the landing gear lights.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Wheel.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Wheel.cfg
@@ -1,74 +1,167 @@
+//  ==================================================
+//  RoveMax Model M1.
+//  ==================================================
+
 @PART[roverWheel1]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
 	@mass = 0.045
-	@MODULE[ModuleWheel]
-	{
-		@resourceConsumptionRate = 0.000525
-	}
+
+    @MODULE[ModuleWheelMotor]
+    {
+        @idleDrain = 0
+
+        @RESOURCE[ElectricCharge]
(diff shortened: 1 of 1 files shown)


NathanKell:
Uh, maybe I'm confused about what the numbers mean, but is that really saying the TR-2L wheels draw less electricity than most light bulbs?

PhineasFreak:
I could increase it even further, say like 55 - 60 W? It was 1 W before the patching though.

NathanKell:
Yeah, I just find it super odd that a large electric motor driving a wheel draws so little. At, say, 15m/s, what do the TR-2Ls draw? And is that similar to 1/4 what a Prius draws at 15m/s / 33mph / 54kph ?

PhineasFreak:
Testing a 4 - wheeled rover around 10 - 20 m/s i get a consumption of 90 - 100 W (as expected).

Now, to modify the top speed since it can easily exceed 50 m/s! I assume that the "wheelSpeedMax" controls the maximum speed that the wheel can reach, right?

NathanKell:
A Chevy Volt at highway speed (~32m/s) uses ~5kW for its four wheels. So I think we need to increase by an order of magnitude. I would suggest the TR-2Ls and the medium Rovemax should be about 1kW each at ~30m/s, and the large ones probably more like 10? Dunno about them. The tiny wheels, now, for _them_ 90-100W makes sense.

PhineasFreak:
I was thinking about cutting the top speed of the larger wheels down to 10 m/s and set the small ones to 2 - 4 m/s. Right now they do not feel exactly like "rover wheels".

The S2 wheels have similar size to the MSL wheels (0.3 m vs 0.5 m). The MSL RTG provides up to 120 W of power for all systems so 6 of these should not require more than 7 - 9 W.

PR #1270, opened 2016-06-05
RO Squad Engines updates/fixes

PhineasFreak:
* Separate the VSR passes from the main RO configs for better clarity.
* Make sure that ant new parts (+PART) are created before any RO changes are applied to the source part (@PART).
* Remove redundant attachment rules (taken care by the global RO patches).
* Decrease the crash tolerances and the maximum temperatures. Engines now can be destroyed when deorbited.
* Start to move the "ModuleEnginesRF" from the RealPlume pass to the main RO pass.
* Remove some redundant modules (taken care by the global engine configs).
* Remove leftover IGNITOR{} blocks (taken care by the global engine configs).
* Remove some VSR "tankbutt" nodes (not applicable unfortunately for all the engines).
* Comment each patch for easier search/identification.
* Remove some old engine FX configs (taken care by RealPlume).
* Fix some model removals (!MODEL).
* Replace the Altair - II model with the same model as Altair.
* Merge the two RAPIER patches into one.
* Fix the KVD-1 and the RD-0105/0109 models.
* Fix the KVD-1 verniers (ullage, ignitions, gimballing).
* Fix the attachment node sizes and placements of some parts.
* Simplify RESOURCE removal from any parts that require it.
* Add a STAR 37 SRM.
* Fix the size of the STAR 48B SRM.

NOTE 1: Due to an issue with RealPlume the KVD-1 main engine plume will not appear.
NOTE 2: The NSTAR ion engine still uses the old workaround for setting the ElectricCharge consumption. This will be fixed in a future update.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1,5 +1,5 @@
 //  ==================================================
-//  RD-253
+//  RD-253 (booster engine).
 //  ==================================================
 
 +PART[engineLargeSkipper]:BEFORE[RealismOverhaul]:NEEDS[!RftS,!RealFuels_StockEngines]
@@ -21,11 +21,6 @@
 		scale = 1.82, 1.82, 1.82
 	}
 
-	@MODEL:NEEDS[VenStockRevamp]
-	{
-		@model = VenStockRevamp/Squad/Parts/Propulsion/SkipperB
-	}
-
     %scale = 1.0
 	@rescaleFactor = 1.0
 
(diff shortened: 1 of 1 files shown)


NathanKell:
Uh...please test your changes in future. This killed gimbal on the LR79, LR89, and H-1. Those gimbal MODULE patches do need to be there, otherwise the transform is not set properly.

PhineasFreak:
The engines had gimbal modules but VSR changes the transform name to be "Nozzle". This is a not valid gimbal transform though with the latest versions.

NathanKell:
All three engines use our own model, not VSR's new model, and therefore they need to have their gimbal transform set to what our model uses...

PhineasFreak:
Damn, i went with the part module name and forgot the actual part model...mea culpa.

NathanKell:
It happens, no worries. :) Just do please try to test a change like "removing the gimbal patch" ;)

NathanKell:
Also we leave minThrust at 0 for solids, so that you can configure the thrust limiter in the VAB (to represent the different shapings and grains that could be used).

PR #1272, opened 2016-06-06
Support for roughly half of BDB solid motors

Ravenchant:
Titan boosters are still missing...and some part descriptions.


NathanKell:
Quick question, are you using global engine configs where available?

Ravenchant:
Sure. I didn't yet bother making the missing global configs though, as there is no real information about thrust curves.

PR #1254, opened 2016-05-31
SSTU engines

Zarbizaure:
Support for:

- Merlin 1B (stats from wikipedia.fr)

- Merlin 1B vacuum (a bit guessed stats/calcultated by a ratio between Merlin 1C/1CVac stats)

- XLR81 (Agena engine, the SSTU name was just LR81 but design seems to match)

- Add a correct title for the fictive RD-107X engine (RD-107 withouth verniers)

- Plumes for Merlin 1B and Merlin1B Vacuum (copy/pasted from Merlin 1C/1C vac because same bell size)

Also, two files in this SSTU folder (SSTU-RD-0110 and SSTU-RD-107X) have - instead of _ between "SSTU" and "EngineName"
By the way Github failed to merge because I created a new file; you could delete my own SSTU-RD-107X, the only thing to add in this file is a `@title = RD-107X`

PR #1276, opened 2016-06-06
SSTU Pods and Tanks

ctiberious:
Update consists of a number of fixes for all of the SSTU pods (series A, B & C plus lander core), and fuel tanks.  I will include a more detailed comment with all the changes/fixes once I have everything posted.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Apollo.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Apollo.cfg
@@ -1,4 +1,4 @@
-@PART[SSTU-SC-B-BPC]:FOR[RealismOverhaul]
+@PART[SSTU-SC-B-BPC]:FOR[RealismOverhaul]:NEEDS[SSTU]
 {
 	%category = Engine
 	%RSSROConfig = True
@@ -14,13 +14,60 @@
 	@mass = 2.574
 	@maxTemp = 1973.15
 	%stagingIcon = DECOUPLER_VERT
-	@RESOURCE,*
+	!RESOURCE,* {}
+	!MODULE[SSTUAutoDepletionDecoupler]{}
+	!MODULE[ModuleEnginesFX],1{}
+	!MODULE[MultiModeEngine]{}
+	@MODULE[ModuleEngines*]
 	{
-		@amount *= 6.4
-		@maxAmount *= 6.4
+		@minThrust = 689.5
+		@maxThrust = 689.5
+		@heatProduction = 100
+		@atmosphereCurve
+		{
+			@key,0 = 0 190
+			@key,1 = 1 176
+		}
+	}
+	MODULE
+	{
+		name = ModuleDecouple
+		ejectionForce = 100
+		explosiveNodeID = bottom
+		staged = true
+	}
+	MODULE
+	{
+		name = ModuleFuelTanks
+		type = HTPB
+		volume = 903.38
+		basemass = -1
+	}
+	MODULE
+	{
+		name = ModuleEngineConfigs
+		type = ModuleEngines
+		modded = False
+		configuration = Solid
+		CONFIG
+		{
+			name = Solid
+			minThrust = 0
+			maxThrust = 689.5
+			PROPELLANT
+			{
+				name = HTPB
+				ratio = 1.0
+			}
+			atmosphereCurve
+			{
+				key = 0 190
+				key = 1 176
+			}
+		}
 	}
 }
-@PART[SSTU-SC-B-CM]:FOR[RealismOverhaul]
+@PART[SSTU-SC-B-CM]:FOR[RealismOverhaul]:NEEDS[SSTU]
 {
 	%RSSROConfig = True
 	@MODEL,0
@@ -41,11 +88,93 @@
 	@title = Apollo Command Module
 	%manufacturer = North American
 	%description = Apollo Command Module. Contains three astronauts.
-	@mass = 4.6715
+	@mass = 4.6505
 	%skinMaxTemp = 3200
 	@maxTemp = 500
 	@CoMOffset = 0.0, -0.5, 0.0
+	%skinMaxTemp = 3600 // 3000 for lifting reentries, but with welded shield use shield max
+	%maxTemp = 900
+	%emissiveConstant = 0.6 // not too absorptive for reentry
+	%thermalMassModifier = 1.0
+	%skinMassPerArea = 4
 	
+	!RESOURCE,* {}
+	!MODULE[ModuleReactionWheel]{}
+	!MODULE[SSTUModularHeatShield]{}
+	!MODULE[ModuleResourceConverter]{}
+
+	@MODULE[SSTUModularParachute]
+	{
+		@drogueCapName = SC-A-CM-ParachuteCap
+		@drogueRetractedScale = 0.0078, 0.0078, 0.0078
+		@drogueSemiDeployedScale = 0.312, 1.56, 0.312
+		@drogueFullDeployedScale = 0.936, 1.248, 0.936
+		@mainRetractedScale = 0.0078, 0.0078, 0.0078
+		@mainSemiDeployedScale = 0.468, 2.34, 0.468
+		@mainFullDeployedScale = 2.34, 2.34, 2.34
+		@drogueSemiDeployArea = 10
+		@drogueFullDeployArea = 103
+		@mainSemiDeployArea = 258
+		@mainFullDeployArea = 2586
+		@DROGUECHUTE[Generic],0
+		{
+			@localPosition = 0.2886, 0.8268, 0.7644
+			@retractedUpVector = 3.7206, 27.3, 0.7644
+			@semiDeployedUpVector = 3.7206, 27.3, 0.7644
+			@fullDeployedUpVector = 8.368152, 26.052, 0.7644
+		}
+		@DROGUECHUTE[Generic],1
+		{
+			@localPosition = -0.2886, 0.8268, 0.7644
+			@retractedUpVector = -3.7206, 27.3, 0.7644
+			@semiDeployedUpVector = -3.7206, 27.3, 0.7644
+			@fullDeployedUpVector = -8.368152, 26.052, 0.7644
+		}
+		@MAINCHUTE[Generic],0
+		{
+			@localPosition = 0.603876, 0.766896, 0.34866
+			@retractedUpVector = 4.214496, 40.34316, 2.434848
+			@semiDeployedUpVector = 4.214496, 40.34316, 2.434848
+			@fullDeployedUpVector = 20.853144, 39.331188, 12.048504
+		}
+		@MAINCHUTE[Generic],1
+		{
+			@localPosition = -0.603876, 0.766896, 0.34866
+			@retractedUpVector = -4.214496, 40.34316, 2.434848
+			@semiDeployedUpVector = -4.214496, 40.34316, 2.434848
+			@fullDeployedUpVector = -20.853144, 39.331188, 12.048504
+		}
+		@MAINCHUTE[Generic],2
+		{
+			@localPosition = 0, 0.766896, -0.69732
+			@retractedUpVector = 0, 40.34316, -4.867356
+			@semiDeployedUpVector = 0, 40.34316, -4.867356
+			@fullDeployedUpVector = 0, 39.331188, -24.083592
+		}
+	}
+	MODULE
+	{
+		name = ModuleAblator
+		ablativeResource = Ablator
+		lossExp = -8000
+		lossConst = 0.06
+		pyrolysisLossFactor = 26000
+		ablationTempThresh = 500
+		reentryConductivity = 0.01
+		charMax = 1
+		charMin = 1
+	}
+	@MODULE[ModuleAblator]:NEEDS[DeadlyReentry]
+	{
+		@name = ModuleHeatShield
+		depletedMaxTemp = 1200
+	}
+	RESOURCE
+	{
+		name = Ablator
+		amount = 848
+		maxAmount = 848
+	}
 	@MODULE[ModuleCommand]
 	{
 		@minimumCrew = 0
@@ -54,19 +183,58 @@
 			@rate = 1.895
 		}
 	}
+	@MODULE[ModuleRCS]
+	{
+		@thrusterPower = 0.4136846
+		!resourceFlowMode = DELETE
+		!resourceName = DELETE
+		@PROPELLANT[*],0
+		{
+			%name = MMH
+			%ratio = 0.456
+			%resourceFlowMode = NO_FLOW
+		}
+		@PROPELLANT[*],1
+		{
+			%name = NTO
+			%ratio = 0.544
+			%resourceFlowMode = NO_FLOW
+		}
+		@atmosphereCurve
+		{
+			@key,0 = 0 260
+			@key,1 = 1 100
+		}
+	}
+	@MODULE[ModuleDockingNode]
+	{
+		@nodeType = Apollo
+		%acquireForce = 0.5 // 2
+		%acquireMinFwdDot = 0.8 // 0.7
+		%acquireminRollDot = -3.40282347E+38
+		%acquireRange = 0.25 // 0.5
+		%acquireTorque = 0.5 // 2.0
+		%captureMaxRvel = 0.1 // 0.3
+		%captureMinFwdDot = 0.998
+		%captureMinRollDot = -3.40282347E+38
+		%captureRange = 0.05 // 0.06
+		%minDistanceToReEngage = 0.25 // 1.0
+		%undockEjectionForce = 0.1 // 10
+	}
+	@MODULE[SSTUNodeFairing]
+	{
+		@FAIRING,0
+		{
+			@topY *= 1.56
+			@bottomY *= 1.56
+			@topRadius *= 1.56
+			@bottomRadius *= 1.56
+		}
+	}
 	MODULE
 	{
 		name = CoMShifter
 		DescentModeCoM = 0, 0, -0.192
-	}
-	!RESOURCE[ElectricCharge]
-	{
-	}
-	!RESOURCE[MonoPropellant]
-	{
-	}
-	!MODULE[ModuleReactionWheel]
-	{
 	}
 	MODULE
 	{
@@ -126,7 +294,7 @@
 		{
 			name = Waste
 			amount = 0
-			maxAmount = 162
+			maxAmount = 166
 		}
 		TANK
 		{
@@ -141,10 +309,80 @@
 		converterName = CO2 Scrubber
 		conversionRate = 3.0	// # of people - Figures based on per/person
 		inputResources = CarbonDioxide, 0.006216, ElectricCharge, 0.010, LithiumHydroxide, 0.00001189
-		outputResources = Waste, 0.00003847, false
+		outputResources = Waste, 0.00003932, false
+	}
+	MODULE:NEEDS[RemoteTech]
+	{
+		name = ModuleSPU
+	}
+    MODULE:NEEDS[RemoteTech]
+    {
+        name = ModuleRTAntennaPassive
+        TechRequired = start
+
+        OmniRange = 2000000
+
+        TRANSMITTER
+        {
+            PacketInterval = 0.4
+            PacketSize = 0.27
+            PacketResourceCost = 0.025
+        }
+    }
+	!MODULE[ModuleDataTransmitter]:NEEDS[RemoteTech]{}
+	MODULE:HAS[!MODULE[ModuleConnectedLivingSpace]]:NEEDS[ConnectedLivingSpace]
+	{
+		name = ModuleConnectedLivingSpace
+		passable = true
+		impassablenodes = bottom
+	}
+}
+@PART[SSTU-SC-B-CMX]:FOR[RealismOverhaul]:NEEDS[SSTU]
+{
+	%RSSROConfig = True
+	@MODEL,0
+	{
+		%scale = 1.56,1.56,1.56
+	}
+	@MODEL,1
+	{
+		@position = 0, 1.3507416, 0
+		@scale = 0.78, 0.78, 0.78
+	}
+	//%rescaleFactor = 1.25
+	@node_stack_top = 0,1.400412,0,0,1,0,1
+	@node_stack_bottom = 0,-0.2723916,0,0,-1,0,2
+	@node_stack_bottom2 = 0,-0.91494,0,0,-1,0,2
+	@CoPOffset = 0.0, 0.936, 0.0
+	@CoLOffset = 0.0, -0.702, 0.0
+	@title = Apollo Command Module - Orbital Module
+	%manufacturer = North American
+	%description = Apollo Command Module - Orbital Module variant. Contains three astronauts. DO NOT USE FOR RE-ENTRY!
+	@mass = 4.6715
+	%skinMaxTemp = 3200
+	@maxTemp = 500
+	@CoMOffset = 0.0, -0.5, 0.0
+	%skinMaxTemp = 3600 // 3000 for lifting reentries, but with welded shield use shield max
+	%maxTemp = 900
+	%emissiveConstant = 0.6 // not too absorptive for reentry
+	%thermalMassModifier = 1.0
+	%skinMassPerArea = 4
+
+	!RESOURCE,* {}
+	!MODULE[ModuleReactionWheel]{}
+	!MODULE[SSTUModularHeatShield]{}
+	!MODULE[ModuleResourceConverter]{}
+
+	@MODULE[ModuleCommand]
+	{
+		@minimumCrew = 0
+		@RESOURCE[ElectricCharge]
+		{
+			@rate = 1.895
+		}
 	}
 	@MODULE[ModuleRCS]
-	{
+	{		
 		@thrusterPower = 0.4136846
 		!resourceFlowMode = DELETE
 		!resourceName = DELETE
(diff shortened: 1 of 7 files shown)


ctiberious:
Apollo:
	Updated BPC to use correct HTPB fuel.  Setup now matches other versions of Apollo LES.
	Removed all inappropriate "RESOURCE" entries and moved then to TANK where appropriate
	Adjusted CM, CMX & SM weights to match weight from similar parts used in FASA
	Added CM parachute to CMX part.
	Adjusted CM, CMX & SM parts have supplies and life support updated to match FASA
Landers:
	I've removed all the parts that no longer exist in SSTU
	All three pods have had RESOURCE removed and/or converted to TANK where appropriate
	Life support for LC2 has been setup similar to FASA LEM, but assumes the bulk of O2, water and waste water would be in a separate descent module
	Life support for LC3 & LC5 have been setup similar to SSTU Orion, but assumes the bulk of O2, water and waste water would be in a separate descent module
	These changes supercede my earlier proposed changes from #1254
Orion:
	BPC setup to use HTPB fuel similar to the Apollo LES.
	Removed all inappropriate "RESOURCE" entries and moved then to TANK where appropriate
	Added CM parachute to CMX part.
	Updated RCS information to match what I could find on the web.  References included in the file changes
	Updated life support resources to match what I could find on the web.  References included in the file changes where I could find them.
SLS Upper Stages:
	Removed all inappropriate "RESOURCE" entries and moved then to TANK where appropriate
	Reset tank type for both to "BalloonCryo".  I don't currently have a reference for this but I'm under the impression that the Delta Cryogenic Second Stage (which is all the ICPS really is) uses balloon tanks similar to the newest Common Centaur uses on Altas V
	Based on what I can find on the web, ICPS is just a reinforced DCSS using the same RL10B-2 engine from DCSS.  I've reset the engine on ICPS to match.
	RCS on both stages has been revamped to use Hydrazine.
	Since Hydrazine can't be included in a BalloonCryo tank, I've added a Hydrazine RESOURCE to both parts.  The amount included is a close approximation to the rcs fuel mass I could find listed on the web.
Soyuz:
	BPC setup to use HTPB fuel similar to the Apollo LES.
	Removed all inappropriate "RESOURCE" entries and moved then to TANK where appropriate
	Masses for OM, DM and SM parts have been adjusted to match Soyux 7K-0K.  Reference included in changes.
	Ablator has been added to the DM.  For lack of better values, I used the same Ablator setup from Apollo.
	Life support and fuel resources updated to match Soyuz and Prospect information I could find on the web.  Refrences included in changes.
	Soyuz life support module changed to reflect the Potassium Supoeroxide system actually in use.  I've setup the system to most (about 99.5%) use Potassium Superoxide with about 0.5% CO2 production being handled by standard Lithium Hydroxide candisters.
Tanks:
	Adjusted minimum and maximum diameter to match what other procedural tanks allow
	Adjusted the diameter increment to match what other procedural tanks allow.  The recuded value also give better control over final diameter.  Unfortunately, it seems that there is now equivalent to "diameterLargeStep" in SSTU so the large step is only twice as much as the small step.
	Built-in RCS on the MUS tanks has been switched to use Hydrazine.  RCS thrust scales with tank size.

One issue I still can't seem to resolve.  Since it's possible that you might use one of the MUS tanks in a non-ServiceModule configuration, there is no way to include Hydrazine for the built in RCS system.  I tried to include a RESOURCE but when I test the parts in game, the resource is not available (even though it's there when I look at the debug information).
Other than the above issue, I've tested everything else in game and it all appears to work correctly.

SirKeplan:
the DCSS is not a balloon tank, i don't know how much difference that makes here, but the DCSS is a structurally rigid, aluminium isogrid tank, without a common bulkead. also has a slightly worse mass fraction than the Centaur.

Thomot512:
Hi,
I'm knew to all this and i didn't knew how to propose change to a file in a pull request that's why i'm doing it so:
the Apollo CM has a space between his top and his docking port because they aren't at the same scale.
These modifications should do the trick:

@MODEL,1
{
	@position = 0, 1.3312416, 0
	@scale = 0.78, 0.78, 0.78
}
@node_stack_top = 0,1.3856076,0,0,1,0,1

the scale that you use for the docking nodes of the three landing pods is not the same as they are used in Apollo or Orion or in the docking port themselves. To correct that i suggest:

for LC2:

@MODEL,1
{
	%scale = 0.78,0.78,0.78
	%position = 0, 0.9916375, 0
}
@node_stack_top = 0,1.0602463,0,0,1,0,2

for LC3

@MODEL,1
{
	%scale = 1.33333,1.33333,1.33333
	%position = 0, 1.31323, 0
}
@node_stack_top = 0,1.43784,0,0,1,0,2

for LC5

@MODEL,1
{
	%scale = 1.33333,1.33333,1.33333
	%position = 0, 1.52673, 0
}
@node_stack_top = 0,1.65134,0,0,1,0,2

I don't know if that's what you wanted but i see that you deleted the ascent tanks of the three landers. it you want to put them back without using SSTUModelSwitch i suggest these line with instead of LC2, LC3 and LC5 when required:

MODEL
{
	model = SSTU/Assets/LC2-ASCF
	position = 0, 0, 0
	rotation = 0, 0, 0
	scale = 1.25, 1.25, 1.25
}

At last, i'm not sure how you ended up with the volume of the tanks for the 3 landing pod. If i use the same logic for the volume of the fuel tanks as for the mass of the pod i get:

for LC2

volume = 2246 l

for LC3

volume = 5664 l

for LC5

volume = 7617 l

Once again i'm sorry if i did it wrong, or if i misunderstood your work.

Zarbizaure:
I think you forgot to add Charred Ablator ressource to the soyuz Orbital module (and Apollo too). Otherwise, great changes!

ctiberious:
@SirKeplan I'll change both the EUS and ICPS to be Cryo tanks instead of BalloonCryo.  Still have the same issue of not being able to support Hydrazine, though.

@Thomot512 I didn't actually mess with any of the sizing.  That's all from the original RO configurations.  But I'll spend some time fine tuning the rescaling.
I temporarily removed the SSTUModelSwitch because it was causing problems with the ModularFuelTank volumes.  As soon as I figure out what the issue is and get it resolved, I intended to restore the ability to switch models.
As for the volumes, keeping in mind that these have no external tanks right now, I figured most of the pods space would be set aside for living area.  I therefore kept the storage capacities limited to what was needed for life support.  And then only the life support that couldn't be offloaded to the descent stage.

@Zarbizaure Thanks for that.  I knew something was missing when I copied the heat shield info from the FASA Apollo CM.

Zarbizaure:
@ctiberious The soyuz descent module is smaller (2.2m wide) than the Apollo capsule. Plus, it is at least twice lighter. You should use this config instead:

> 	MODULE
	{
		name = ModuleAblator
		ablativeResource = Ablator
		outputResource = CharredAblator
		outputMult = 0.75
		lossExp = -40000
		lossConst = 15000
		pyrolysisLossFactor = 40000
		ablationTempThresh = 500
		reentryConductivity = 0.01
		//reentryConductivity = 0.12
		//@reentryConductivity = #$../heatConductivity$ // if it exists, use it
	}
	@MODULE[ModuleAblator]:NEEDS[DeadlyReentry]
	{
		@name = ModuleHeatShield
		depletedMaxTemp = 1200
	}
	RESOURCE
	{
		name = Ablator
		amount = 200
		maxAmount = 200
	}
	RESOURCE
	{
		name = CharredAblator
		amount = 0
		maxAmount = 150
	}
}

Zarbizaure:
After some testing my resizing is ok (applying the 1.173333 factor everywhere) but the soyouz LAS has a p^roblem: the abort FX are always active. Is that because you addded a "1" just after `!MODULE[ModuleEnginesFX],1{}` ?

ctiberious:
@Zarbizaure I'll add those heat shield values to Soyuz once I've done another run through of all the parts.  I want to run a few more tests before adding more commits.
What resizing did you do?  For the most part, the scale values I'm using are the same ones shadowmage45 used in his last batch of updates, though I know he got those from someone else.
The LAS for Soyuz (and for Apollo & Orion) has two ModuleEnginesFX tags: One for abort mode and one for jettison mode.  "!MODULE[ModuleEnginesFX],1{}" means, delete the second module of that name (in this case, the jettison mode one).  I'm deleting one ModuleEnginesFX tag and leaving just the one.  Basically they are setup like most of the other LAS I've seen in various mods.  Are you seeing the FX play even when the Soyuz is on the pad or just from the VAB?  I've noticed most of the RCS FX seem to be displaying from the VAB but I haven't noticed them displaying incorrectly once I'm on the pad.  I'll take a closer look at what is going on here to see if something I've done is causing an FX issue.

Zarbizaure:
The current soyouz model (non-resized) is 1.875m and the real one is 2.2 so you have to multiply every value by 1.173333333. I'll send you my version of the file soon.

About the LAS: I see nothing in the VAB but once I go on the launchpad, one of the effect start to play (I suppose it's the jettison effect, because the running effect work fine once you activate the engine), even during timewarp.

I did noticed nothing about the RCS though.

ctiberious:
Actually, I just posted some changes to Soyuz that adjusts the size.  Now it's docking port and the one on Apollo match up correctly.  I also discovered why the fx was happening constantly and fixed that by restoring the two separate motors on the LES.  I suppose I could simply have deleted the "jettison" effect but I decided to leave it since the model is built for them.  But I did have to use a different fuel type for the jettison motor.  I used PSPC pretty much at random so hopefully no one has an issue with it.

Zarbizaure:
yes I also though deleting the jettison effect was a solution but it makes sense to keep it. Do you have any source for the Soyouz LES propellant? I'll search and check if it is PSPC.

Zarbizaure:
after some testings (re-entering the descent module from LEO, 85km periapsis) the heat shield may need a bit more skin heat tolerance. I'll open the debug thermal menu and give you a better value.

Zarbizaure:
using the correct paremeter I still explode during re-entry: my ablator abalte way too fast and the spacecraft explode when there is no more ablator. I suspect that DRE LEO shield settings are a bit outdated. Any idea @NathanKell ? How could we get something that survive a LEO re-entry but not a lunar one?

SirKeplan:
is the `outputMult = 0.75 ` field making any difference here? it seems only Lunar shields have this.

Zarbizaure:
the outputMult only generate a bit less charredAblator; which seems logical because with a high energy reentry, more particles will be ablated and thrown away from the heat shield.
I still have no idea why it's impossible to reenter with this config.

ctiberious:
@Zarbizaure I think you misunderstood what I meant about the PSPC fuel.  The AbortMotor on the all three LES uses HTPB which is the same fuel we have the FASA Apollo LES using.  However, the SSTU parts have a separate JettisonMotor.  And a SSTUAutoDepletionDecoupler module which keys on a fuel source.  If you setup both the AbortMotor and JettisonMotor to use the same fuel, then there will be no fuel left when the LES is auto decoupled after an abort.  Therefore both motors have to have a different type of fuel.  I setup the JettisonMotor to use PSPC so that it would be different from the AbortMotor.  My guess is that IRL, both motors use the same type of fuel but we can't do that in KSP (at least not when both motors and fuel sources are on the same part).  If you can find a source for the fuel used on the actual Soyuz LES, I'll correct the AbortMotor, but I'm guessing that we will still have to use a random fuel for the JettisonMotor.

As for the heat shield settings, I have only done one test flight so far.  I put my Soyuz in a 180km orbit (simulated flight using KCT), then de-orbited with an initial periapsis less then 0 (started off at 0, then got impatient and dropped it to -95km for an even steeper re-entry).  My Ablator used about 150 units and the DM landed just fine.  Is the Soyuz supposed to make a very shallow LEO re-entry (85km is about twice the periapsis I'd use on a lunar re-entry) or is it supposed to make a steep re-entry like Mercury did?

Zarbizaure:
Yes, I understood that. But on real life soyuz, I think there is no reason why the LAS won't consume all of its fuel for abort. Is it really how it work? Is it not just decoupled by the drag of the grid fins after an abort? I would like to see a video of a soyouz abort, I'll search a bit.

That's strange...with my reentry, it took about 2 minutes to burn up all of the ablator at a really high altitude, with a really slight deceleration (max G was 0.9G); then it exploded. Maybe a steeper angle will help in a faster descent which means less ablation. I'll try again and see what happen.

SirKeplan:
it seems regular Soyuz does a de-orbit burn of around 120m/s (remember this is from ISS altitude)

ctiberious:
Honestly, I don't know if there are really separate abort and jettison motors on any of these parts IRL.  I just know the way SSTU is setup, there are.  If you look at all three LES parts, even the models have clearly separate nozels for the abort and jettison motors.  Of course, I could always just delete one of the EFFECTS section if there really should only be one kind of motor.

Zarbizaure:
it makes sense to have just one kind of propellant. I did not find any soyuz footage but here's an orion one: https://www.youtube.com/watch?v=5w5p4X6rdjE We can clearly see a second engine igniton in order to jettison the LAS.
I suppose Soyouz works the same way so the way you set them up is right.

ctiberious:
Like I said, we can't have both motors use the same fuel because we can't restrict the quantity of fuel each motor gets too use.  If they both use the same fuel then the AbortMotor will use all the fuel leaving nothing for the JettisonMotor when the SSTUAutoDepletionDecoupler forces the part to decouple.

ctiberious:
I think I've got all the fine tuning done unless someone sees anything that I've missed or need to take another look at.

Zarbizaure:
oh, if we consider this is no longer WIP could you remove the (WIP) tag from the soyuz parts and set %RSSROconfig = true?

ctiberious:
Soyuz parts have been marked as RSS/RO and WIP tag removed.

Zarbizaure:
nice! I think it's ready to be merged :-)

Thomot512:
For the Landers, with your configuration the LC2 has still unrealistic volume since you let the SSTUModelSwitch activ.
As a temporary solution to this problem i suggest to make 2 separate part per lander. (See attached file)
I think personally that the LC-2's docking port should be cpmpatible with the Apollo's docking port and not the Orion. (See attached file)
I tried to match the weight of the LC-2 with the one of the Apollo lunar module and keep the same mass coefficient for the oder Pods. (See attached file)

[SSTU_Landers.txt](https://github.com/KSP-RO/RealismOverhaul/files/307885/SSTU_Landers.txt)
( was not able to attach the file as .cfg so just change extension if you want to try)

ctiberious:
Woops.  The SSTUModelSwitch shouldn't be active.  I accidentally left it on from a test I was running.  I'm still trying to figure out how to use that model without it messing with the ModuleFuelTanks volume.  I'll get that corrected shortly.
As for the docking port, I assumed that the LC2 was intended to be a smaller version of Altair.  Not a different version of the Apollo LEM.  If I'm mistaken in that assumption, then I don't have an issue adjusting the size and mass of the LC2.

Thomot512:
The docking node in SSTU are both (small and big) compatible. But with RO patch this is not the case anymore, the LC2 pods is unlocked (5th lvl) before the Apollo CM (6th lvl) in the tech tree and after the Soyuz(4th lvl).The Orion CM in the other hand is unlocked much later (8th lvl).
Meaning, for me at least, the LC-2 should work with the Apollo and Soyuz which both have an apollo, small sized Docking port.
And what do you think of the workaround for the model switch : Having two different part once with ascent tanks and once without directly in editor.

ctiberious:
Based on the original sizes, all three landers appear to have been designed to dock with the Orion.  Not the Soyuz or Apollo.  I'm not saying whether the original code would allow Apollo to Dock to LC2.  I'm just saying that the docking port size on LC2, LC3 and LC5 matches the size of the docking port on Orion.  That tells me that LC2 was not intended to be a replacement of the Apollo LEM.
I can't really comment on the current tech placement of the pods.  They could have been placed where they are completely at random for all I know.
Having extra parts so we can have the landers with and without descent tanks is a possible solution.  But it does mean if we do manager to get SSTUModuleSwitch to work in RO, we'll be losing parts with causes compatibility problems in the future.  Hate to add a part I intend to remove down the road.

I'm thinking about resetting the LC3 pod to be a 4-man pod instead of a 3-man.  That would give us options for 2-man, 4-man and 6-man.  It would also allow the LC3 to somewhat accurate represent the Altair (which I believe was supposed to hold 4 astronaughts).

Thomot512:
You're probably right, about LC2, however i would prefer to have an sstu pod that works with the sstu Apollo CM. 
If you find a way to make SSTUModuleSwitch work with Real Fuel, it would be great.

Thomot512:
I noticed today that you set the RCS thruster of the MUS fuel tanks to Hydrazine. Problem is we can't fill these tanks with hydrazine and your lines:
``
RESOURCE
	{
		name = Hydrazine
		amount = 250
		maxAmount = 250
	}
``
doesn't seem to work.

ctiberious:
@shadowmage45 posted a new version of SSTU that includes a fix for the volume problem we were seeing with the landers.  I'm in the process of updating the configs a bit more to take advantage of the latest SSTU version.

Best case for the MUS tanks would be to figure out a way to use the RO RCS config system.  Until I can figure that out, though, I switched the MUS RCS to match what I'm guessing is (more or less) the standard for upper stage RCS: Hydrazine.  I know it's used on all the American upper stages (Centaur, DCSS and ICSP).  I know you can't include Hydrazine in non-pressurized tanks but it's easy to work around the issue if you're also using Procedural Parts.  I just include a few small, round pressurized tanks for the Hydrazine storage.  I'm also still trying to figure out why the RESOURCE isn't working because it's definitely there according to the database.

ctiberious:
With the latest 116 update to SSTU, I was able to add SSTUModelSwitch back to the Landers.  For some reason the volume is only 86% of whatever you set it to be, so to hit my target volumes, the configs are set with higher volumes (LC2 has 1162.8 which results in 1000 in game).  I've also adjusted the model sizes for the ascent tanks so they appear correctly.

I set the MUS tanks and the RCS thrusters so they can now have their fuel type set like all the other RCS thrusters in RO.

I reset LC3 to be a 4-man lander so it can be used to create something similar to the Altair lander.

The volume values I selected for the Landers were basically pulled from thin air so if someone has a good suggestion, I'm all ears.  For now the LC2 should have 1kL/3kL (standard/ascent), LC3 should have 1.5kL/4.5kL, and LC5 should have 3kL/9kL.

Thomot512:
All of this are grat news!
For the volume i suggest you use a factor 1.33333^3 aprox 2.37 because of the stretching 1.333333 in 3 axis. It is for LC2 590 / 2725. 
Then set the mass to have a dV 2100 with ascent tank and a lunar ascent engine. Because it's approximatly what's needed to leave the Moon. and designing a Ascent module that cant leave the Moon wouldn't be logic. Then use what mass factor it gives you  to calculate mass of LC3 and 5.
With a mass of 2.05 t and this volume i have a dV of 2098 m/s with ascent engine.
This is a mass 1.687243 times bigger than the original.

ctiberious:
Found and (I think) corrected a problem with a bracket that caused issue with LC5

ctiberious:
Wouldn't the ascent stage mass be dependent on the ascent engine which we currently have no data on.

Thomot512:
And a small detail: the vesselType of these Landers is ship it may make sense to set it as Lander.

Thomot512:
And a small detail: the vesselType of these Landers is ship it may make sense to set it as Lander.
The volume corrected for the 86%: 
LC2 vol= 686 / 3200
LC3 vol= 1105 / 7995
LC5 vol= 2070 / 12790

Thomot512:
When i use your lander.cfg file i still have odd value : 

LC-2 1.0 kl (OK)
LC-2 Ascent 5.74 kl (not OK)

I looked a bit myself and discovered that if you take this formula it works aproximately:

Volume in game for Ascent = (Target volume - (Value entered as baseVolume * 0.86) ) / 2.04
                                                        or
Volume in game for Ascent = (Target volume - Volume in game of nonAscent ) / 2.04

the 2.04 is not 100% exact i'm still trying to be more precise.

Temeter:
I've gave this a try, I'm quite impressed how well most of the things seem to work. Mass is very slightly off the proctanks and raw cost a bit higher, but its in a very playable realm.

One issue I did notice, the upper stage tanks have avionics with unlimited control integrated.

ctiberious:
@Thomot512 So it seems that the scale value used to properly resize the ascent tanks for the larger lander model is affecting the volume.  I'm guessing that's something SSTU is doing intentionally but I'll ask and see if there is another way to rescale the tanks that won't mess with the volume.

@Temeter It's not surprising that the upper stages have unlimited avionics.  Avionics is part of RP-0 and all I'm doing here is working on making the parts functional in RO.  We'll have to do a PR in RP-0 to update their avionics at some point.  But figured I should start with making sure the parts work right in RO.

Temeter:
Ah right, I constantly forget that avionics are RP-0 only.^^'

Actually kinda weird, considering how important they are to real rocketry.

ctiberious:
I added the RO RCS setup to the landers.  I've used the RCSBlockQuarter for LC2 (69/111N), RCSBlockHalf for LC3 (138/223N) and RCSBlock for LC5 (275/445 N).  May have to mess with those but figured this is a start.

As for the volume, I figured out what is causing the issue.  SSTU takes the cube of the scale value and multiplies it against the listed volume.  I've adjusted the volumes to account for this.  We now should be back to the volumes I originally tried to use.  As for the volumes @Thomot512 suggested, my first concern is that the Apollo LEM ascent stage has 2685L which is only slightly less than your 2725L suggestion.  I would hope that 40 years later, we'd be able to carry more.  I will research Altair more and see if I can find anything that suggests what its carry volume might have been.

NathanKell:
Given Altair was lifting rather more than 2 crew I find it highly unlikely
it had similar tankage to the LM ascent stage,  even with a few more s of
Isp.

Sent by my thumbs, slowly.

ctiberious:
@NathanKell The LC2 is a 2-man lander shaped more or less like what Altair would look like.  Right now I have it set with a 3kL volume (ascent tanks active) because I felt it should have a bit more capacity than the old LM even though it has the same crew capacity.  It seems like it's got more living area than LM had, and going on the idea of a 2-man version of Altair, it would be intended to remain on the Moon longer than LM was.
LC3, on the other hand, has been setup as a 4-man lander (SSTU default is only 3man) so it can be used to create Altair.  It currently has a 4.5kL tank setup (ascent tanks active) but that will change if I can get more accurate info on what Altair was supposed to have.

ctiberious:
@NathanKell We've got a pair of effectively procedural tanks in SSTU_Tanks.cfg that I have been trying to give a Hydrazine RESOURCE.  I've got the code in the cfg and if you look at the database in game, the RESOURCE is definitely there.  But when you actually use either of the tanks (SC-MUS-CB & SC-MUS-ST) the RESOURCE is missing.  I talked with Shadowmage and he verified that it's not the SSTU code (at least not the specific module used to create these procedural parts) that's blocking the RESOURCE.  He thought that it might be something in the Real Fuels system that was causing the problem (linked this code https://github.com/NathanKell/ModularFuelSystem/blob/master/Source/Tanks/ModuleFuelTanks.cs#L144-L162 )  Do you know any reason that the RESOURCE might be blocked?

jwvanderbeck:
Is the tank TYPE set properly?  Hydrazine can only be added to Service Module tanks I think?

ctiberious:
@jwvanderbeck It's not a TANK inside a ModuleFuelTanks module.  It's a RESOURCE.  We do the exact same thing for the ICPS and HUS/EUS.  But on the resizable tanks, the RESOURCE doesn't seem to work.

ctiberious:
Made a number of adjustments to the mass and storage capacity for the landers.  For all three landers I've moved all the life support into the lander itself.  So LC2 and LC3 have 7day capacity while LC5 has 21day.  Without ascent tanks, there's enough storage for the life support, a small amount of rcs fuel (NTO/MMH by default) and 100k EC, with a little extra capacity for customization.  With the ascent tanks there's enough space for the above plus enough MMH/MON3 (for an AJ-190 used as ascent engine) for around 2800dV which lines up with what the LM needed to return to orbit.  Have to test if that's enough to get into orbit from a polar landing site but so far these numbers look good.  Only drawback I've seen so far is when you switch models, it automatically adjusts the resources for the new volume so while I've included default values for life support, rcs and ec, they are only valid when the ascent tank is off.  As soon as you turn it on, you'll have to manually adjust everything.  Not sure if there's a way to get around that.

Thomot512:
Hum, sorry to tell you, but the volume are still wrong:
LC-2 = 2.55 kl
LC-3 = 3.86 kl
LC-5 = 7.25 kl
may i insist that you try the formula i gave you 2 days ago ? It seems to work.

Other problem the masses:
LC-2 dry = 2.155 t (even if in .cfg you wrote 1.915)
LC-3 dry = 1.803 t
LC-5 dry = 1.68 t
The bigger the lighter I assume it's a mistake.

ctiberious:
I guess I need to include a note that explains, the ascent tank volume is added to the base volume.  It's not used in place of the base volume.  As such, you should find that the volume targets are as follows:
LC2: 550L / 2550L
LC3: 850L / 3861.5L
LC5: 2750L / 7250L
As for the formula you previously suggested, I decided not to use it because I found the actual equation that was altering the volume in the SSTU code.  SSTUModelSwitch We lose 14% of the volume for some reason that neither shadowmage or myself can determine.  Then the cube of the scale value indicated in SSTUModelSwitch is multiplied against the volume.  Hence the equation I used is:
<targetVolume> / .86 / <scale>^3
For LC3, I wanted the ascent tanks to add 3011.5L (or 3.0115m3) therefore:
3.0115 / .86 / 1.333333^3 = 1.477299436462672446406336671317 which I rounded up to 1.477295.

The volume of each pod (without ascent tanks) is such that 50 units MMH/NTO (for rcs), 100k EC, and 7 days (21 on LC5) life support can be stored.  The LC2 includes storage for LiOH and recycler waste.  Totals I came up with are as follows:
LC2: 320.225 life support / 520.225 total
LC3: 518.6 life support / 718.6 total
LC5: 2301.1 life support / 2501.1 total
I added a bit to each to allow for player customization and to give round numbers.  I suppose I could go with 550, 750, & 2550 to cut down on capacity a bit while still meeting the criteria I was working with.

The volume targets I have right now for the ascent tanks assumes the only thing they are used for is fuel for the ascent engine.  That volume was set to give each pod approximately 2700dV at the mass I targeted each with.

As for the masses, no I didn't make a mistake.  The masses I posted are not final since we are still messing around with volume capacity.  Yes, it does currently stand that the larger pod has the lighter mass.  That was not done intentionally.  It's just how it worked out with this first pass.  I will continue to fine tune things as I continue to work out the best values for the tank volumes.
The target masses I am using for LC2 and LC3 come from the pdf that is posted in the cfg file: https://www.nasa.gov/pdf/289914main_fs_altair_lunar_lander.pdf
That pdf includes a comparison of the 2-man Apollo LEM vs the 4-man Altair.  Including ascent stage mass.  I'm therefore setting LC3 (including supplies and engine) to 6.141t.  For LC2 I'm aiming for 4.8t.  I don't actually think LC2 should have an identical mass to LM.  It presumably (being just a smaller version of Altair) has more crew area and an airlock, but is also made using modern parts so I don't know if it should be a little more or a little less than LM.  But I figured as long as it's close, we're probably good.
LC5 is a bit harder since I don't really have a target.  Right now I think I set it around 8.5t but I'm open to suggestions.
In all three cases, I assumed the "ascent stage" was the lander pod plus an AJ10-190 which appears to be what was planned for Altair (at least according to https://en.wikipedia.org/wiki/Altair_(spacecraft) in the "Engines" section).

ctiberious:
As a note, it appears that it's RealFuels that is altering the amounts when the ascent tanks are turned on.  Not SSTU.  @NathanKell can you possibly verify that Real Fuels is altering the cfg set TANK amounts when the volume is altered in game?  And I don't suppose there is a flag or something we can add to the TANK so that the amounts can remain static (unless intentionally changed by the player)?

ctiberious:
Here's a new set of masses and volumes which should look a bit better.  Target masses, standard volume and total volume (with ascent tanks) are as follows:
LC2: 4.2t / 550L / 2720L - with 2170L fuel, 2803dV
LC3: 6.141t / 750L / 310L - with 3160L fuel, 2785dV
LC5: 9.2t / 2550L / 7280L - with 4730L fuel, 2781dV
Like before, the base volume includes enough storage for life support, electricity and rcs fuel.  The ascent tanks add the volume for the fuel.  Target mass assumes the inclusion of a single AJ10-190 engine.
Apollo LM gets something like 2450dV so I figured the 2780-2800 I set these pods up with should be enough to include up to 500kg extra science components, while still giving enough energy to launch from a polar orbit (still needs to be tested).
I tried setting the default tanks to include everything (life support, ec, rcs, and ascent fuel) but it seems like Real Fuels is still altering the amounts.  At this point I'm wondering if I should just remove the default tanks completely.

jwvanderbeck:
First off @ctiberious I want to say **thank you so much for this**.  It is very much appreciated!

Secondly, moving forward can we maybe tag the SSTU release that the configs are for in the PRs? Due to rapid changes on both sides it is getting difficult to make sure everyone is using the right things.

ctiberious:
@jwvanderbeck I'm not sure how to add tags to PRs.  :(  These changes are meant for 0.4.31.116 which included the fix needed for the landers to use their switchable models.

jwvanderbeck:
Sorry the use of the term tag was more formal than it should have been.  I just meant indicate in the PR either title or text just so we know.

Thomot512:
@ctiberious Great! looks realy good now! 
Thanks a lot for your work.

I tried to set a second group of model in  SSTUModelSwitch wich could handle a change in docking port, but i got stuck when i wanted to change the top node.
I had the idea to create 2 top node and that each model would supress 1 of the node wich would allow us to use 2 ModuleDockingNode with 2 nodeType. but i don't see how to make SSTUModelSwitch delete a node.
Do you know if it's possible?

shadowmage45:
@Thomot512 
Switching docking ports generally won't be possible with the ModelSwitch, due to the docking port needing the backing MODULE updated with new transform names for each model.

With that said, yes, it can handle node switching, but it is a fairly complex setup... that I don't remember off the top of my head how to put into the config :)   I'll try and come up with some examples/instructions for it.

Thomot512:
@shadowmage45 Thanks that for the help. And of course thanks for your amazing mod!

ctiberious:
Adjusted base volume to more closely match procedural parts.
Added ability to adjust tank utilization.

Zarbizaure:
while you're on it, could you adjust the increment procedural diameter parameter on every SSTU procedural part to 0.1 instead of the stock-KSP 0.625 setting? (on petal adaptaters, fairings, interstages, procedural decouplers...)

Zarbizaure:
good change :)
Also, I still struggle to re-enter the soyouz DM. I tried various descent path, from a positive +50km Pe to a negative -500km one. And it still burns up durning reentry.

ctiberious:
I've updated a number of misc parts so they increment at 0.1 where possible.  
Avionics was mentioned previously but that's an RP-0 thing.  Not an RO thing.  At some point someone (maybe me) will have to do a PR for SSTU in RP-0 but I figure first thing we need to do is get the parts working in RO.
I'll run another test with the Soyuz.  I've only done one re-entry test from a 200km orbit with a perapsis of -95km and it landed just fine.

Zarbizaure:
yeah, maybe it's just me. It always explode of overheating. I just added a standard 2m lunar heat shield under it for my RP-0 campagin, but it does not fix the SSTU part.

Zarbizaure:
I also noticed the Apollo docking port has the wrong description. It is described as a NASA docking port (Orion one) instead of Apollo, but it is correctly set up (name and docking type) as Apollo.

ctiberious:
I'll correct the description in a bit.

Zarbizaure:
oh, and while playing I acyually found that there is a soyouz heat shield (part name = SoyuzHeatShieldLEO ) but I have no idea from where it comes.

Zarbizaure:
I actually found these value in Tantares:
`	%emissiveConstant = 0.6 // not too absorptive for reentry
	%thermalMassModifier = 1.0
	MODULE
	{
		name = ModuleAblator
		ablativeResource = Ablator
		lossExp = -8000
		lossConst = 0.09
		pyrolysisLossFactor = 26000
		ablationTempThresh = 500
		reentryConductivity = 0.01
		charMax = 1
		charMin = 1
	}
	@MODULE[ModuleAblator]:NEEDS[DeadlyReentry]
	{
		@name = ModuleHeatShield
		depletedMaxTemp = 1200
	}
	RESOURCE
	{
		name = Ablator
		amount = 266
		maxAmount = 266
	}`

Maybe we could try, we'll just have to add an output ressource = charredAblator

SirKeplan:
Yep `SoyuzheatShieldLEO` is for the Tantares Soyuz, i set that up based on the values the DeadlyReentry shields had at the time. should still work but seems things may have changed a little in the meantime. if there's issues probably best to just go by how the actual Deadly Reentry heatshields are configured.

ctiberious:
I've corrected the description for the Apollo docking system, though its pretty generic.

NathanKell:
I'm going to go ahead and merge this, since it's obviously better support than what we have but I want to get a release out tonight. Please keep working on this stuff however! :)

PR #1280, opened 2016-06-20
BobCat Soviet Engines 1.1.2 Update

Theysen:
brought inline with recent 1.1.2 Update , available via CKAN

 -> changed .netkan to make new version show up under Suggested Mods in exchange for old CKAN name


NathanKell:
Thanks!

PR #1287, opened 2016-06-28
RN KK fairing fixes

raidernick:
-Fix long term standing issue with FAR and the 400 series fairings, only
solvable by making new fairings from scratch. Tested and works in game.


NathanKell:
Um, there's a png?

PR #1289, opened 2016-07-01
better ISP/Thrust for the B9 Sabre engine

Zarbizaure:
Also, are the AJE parameters still okay?
Maybe we could create an engine config file for the S.A.B.R.E. too.
Sea level thrust in air breathing mode is still really low; I think there's something to change in the AJE settings (I get around 400kN whereas wikipedia and REL state "1960kN SL"). This make liftoff possible but it is incredibly hard to gain speed; my skylon stucks at mach 1.2 at 6km.

### DIFF
--- a/GameData/RealismOverhaul/REWORK/RO_B9.cfg
+++ b/GameData/RealismOverhaul/REWORK/RO_B9.cfg
@@ -3105,9 +3105,25 @@
 @PART[B9_Engine_SABRE_Intake_M]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
+
+	!mesh = DELETE
+
+	!MODEL,*{}
+
+	MODEL
+	{
+		model = B9_Aerospace/Parts/Engine_SABRE_Intake/model
+		scale = 1.52, 1.52, 1.52
+	}
+	
+	@node_stack_bottom01 = 0.0, -0.0, 0.0, 0.0, -1.0, 0.0, 2
 	//%maxTemp = 2273.15
 	%skinMaxTemp = 2500
 	@mass = 0.4
+	@MODULE[ModuleResourceIntake]
+    	{
+        	@area = 0.0739
+    	}
 }
 @PART[B9_Engine_SABRE_S_Body]:FOR[RealismOverhaul]
 {
@@ -3195,8 +3211,23 @@
 {
 	%RSSROConfig = True
 	
+	!mesh = DELETE
+
+	!MODEL,*{}
+
+	MODEL
+	{
+		model = B9_Aerospace/Parts/Engine_SABRE_M/model
+		scale = 2, 2, 2
+	}
+	
 	@title = SABRE M
-	@mass = 8
+	@mass = 5
+	
+	@node_stack_top = 0.0, 2.5, 0.0, 0.0, 1.0, 0.0, 2
+    	@node_attach = 0.0, 1.25, 2.5, 0.0, 0.0, -1.0
+	@CoMOffset = 0.0, -1.25, 0.0
+
 	@MODULE[ModuleEngines*]:HAS[#engineID[AirBreathing]]
 	{
 		@name = ModuleEnginesAJEJet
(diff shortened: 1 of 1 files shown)


Zarbizaure:
oh, I just noticed something in the SPH while building a SKylon: the length in a sandbox game is limited to 40m (or length is tagged as heigh...) and in order to launch it, I need to turn it 90 degree. Maybe we could just the upper limit to 500m for lenght, height and width?

Zarbizaure:
DO NOT MERGE NOW. More change are coming.

blowfishpro:
I think there are still some issues

* 1960 kN static in air-breathing mode is almost certainly wrong.  REL's numbers are inconsistent, but the best data I can find suggests 600-800 kN/engine on the runway
* If you want to modify the engine's thrust, change `Area`.  Changing `thrustUpperLimit` will just change when thrust growth begins to fall off
* The intakes won't have `ModuleResourceIntake` - it is replaced by `AJEInlet` by AJE.  And you'll want to modify `Area`, not `area` (it should match `Area` on the engine module)
* Mass is still unknown

Zarbizaure:
thanks for your help!
I'll do some changes and test tomorrow. I tested with a bigger engine area and indeed the thrust increased. However we need a setting so that the thust increase with altitude and speed. Maybe it'll be better with bigger intake area (these were still 3.6m^2, they should be around 12m^2...), maybe we'll have to change some settings to CRP and/or engine inlet temperature.

NathanKell:
...that's not how gas turbines _work_ though. The higher you go, the less dense the air is, so the _less_ thrust you get. Now, increasing speed will increase the flow rate and bring thrust back up, but there is absolutely no way going higher will ever do anything, on its own, other than decrease thrust.

Zarbizaure:
of course less air density means less thurts. But the SABRE, thanks to its precooler, is able to keep a decent amount of thrust even at high altitude and speed. Its TWR does not drop as a conventionnal turbojet. Going higher will mainly reduce the drag on the plane, which will help to gain a bit of speed (and generate a bit less heat); it won't improve engine thurst or performance.

The SABRE precooler is still something really specific and I wonder if there's really a way to emulate it with AJE. What I meant is, there shouldn't be any brutal loss of thrust due to heat at high speed (Mach 2+), which happened during my testing.

NathanKell:
"and speeds" is the key there. You can't create matter from nothingness--you can't get more air mass into your engine than you intake from your intakes, no matter how much you cool it. Its TWR does not drop because it is expected to be at Mach 5 at 25-30km, right?

blowfishpro:
The SABRE uses the turbojet jet model at the moment, which isn't really accurate, but regardless seems to provide a reasonable approximation.  Camlost came up with the original engine parameters for the SABRE, and they seem to work so I'm not really going to mess with them (other than adjusting area since that only affects thrust, and maybe seeing if I can push the Isp up a bit).  I've tried coming up with more accurate models of the SABRE, but there are just too many unknowns.

Zarbizaure:
so I tried with these settings:
		@name = ModuleEnginesAJEJet
		Area = 8.9
		BPR = 0
		CPR = 140
		FPR = 0
		Mdes = 1
		Tdes = 280
		eta_c = 0.95
		eta_t = 0.98
		eta_n = 0.9
		FHV = 92000000
		TIT = 2000
		TAB = 3800
		exhaustMixer = True
		thrustUpperLimit = 1600
		maxT3 = 2000

I have about 850kN of static thrust, but the engines are still very weak at high altitude and speed. I barely reach mach 2.75...after 25 minutes of flight! Meanwhile 75% of LH2 has been burned. Maybe drag forces are not correctly represented too, but I'm almost sure that the SABRE need to gain a bit more thrust at higher speed. At 10km & Mach 2 It stucks under 650kN and there's almost no more acceleration; it's also impossible to pitch up withouth loosing speed.

blowfishpro:
Well it's possible that you have inadvertently reduced thrust growth, or that the aerodynamics are just not as good as they need to be.  It's really hard to say anything just looking at the engine parameters.

Zarbizaure:
probably. I don't know which parameters change how the thrust grows with speed, so...

PR #1290, opened 2016-07-01
RO Squad Utility Updates/Fixes

PhineasFreak:
* Add RT range and power descriptions for all antennae.
* Fix the size of the Communotron HG-55.

NOTE: the power consumption entry is a temporary measure until RT implements the stock value display system.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
@@ -32,7 +32,7 @@
 
 @PART[HighGainAntenna]:FOR[RealismOverhaul]:NEEDS[VenStockRevamp]
 {
-    %rescaleFactor = 0.5
+    %rescaleFactor = 1.0
 }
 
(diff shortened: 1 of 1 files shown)


NathanKell:
What is the size reasoning for the HG-55, out of curiosity?

PhineasFreak:
Someone had tried to rescale but because of VSR it failed. Not a personal reason, just because i saw it: https://github.com/KSP-RO/RealismOverhaul/pull/1290/files#diff-b293827331c2c4b24d3c7f33e6c552eaL9

NathanKell:
Ah, gotcha. In that case I'm inclined to leave it at the existing size then. That size seems to fit well.

PhineasFreak:
OK, give me a sec to make the required changes.

NathanKell:
Ah, this may be a 1.9.2 / 1.9.4 thing. I see in 1.9.2 it's the right size with rescale 0.5, is it different in 1.9.4?

NathanKell:
OH I get this now. It's because there were two HighGainAntenna nodes from the AsteroidDay mess. 0.5 does appear to be the correct scale in the end. Sorry for the mess!

PR #1292, opened 2016-07-01
RO DECQ Angara pack

PhineasFreak:
* Add RO support for the DECQ Angara pack.


SirKeplan:
`@MODULE[ModuleGimbal]` shouldn't be necessary on the engines, gimbal settings are already set in the Engine Configs.

PhineasFreak:
Consider it a leftover from the initial testing round. I will remove it in a sec.

jwvanderbeck:
Um , wasn't it removing the gimbals last time that broke tons of engines?  Unless i'm just misunderstanding, @NathanKell  said the gimbals need to be there didn't he? See https://github.com/KSP-RO/RealismOverhaul/pull/1270 for reference.

SirKeplan:
@jwvanderbeck this should be different. The transform names aren't being changed

PR #1299, opened 2016-07-03
Gimbal range edited for per axis

alimoncul:
Gimbal range edited for per axis


SirKeplan:
Sorry maybe I am missing something here, but what's the point of this change?

NathanKell:
Already spoke to @alimoncul in IRC, all taken care of. :)

PR #1313, opened 2016-07-21
Update RO_UniversalStorage.cfg

EagleEyeLogic:
since UniversalStorage has patch for RealFuels, we better do this :FINAL and keep an eye on for later if the patch will be deleted.

TANK_DEFINITION can be off since RealFuels define Hydrogen


NathanKell:
In general, it's better to use a FOR[zzz(whatever)] pass, i.e. FOR[zzzAfterUniversalStorage] for example. That still lets end users make FINAL patches.

EagleEyeLogic:
@NathanKell You are right, sometimes it's better too think twice

PR #1314, opened 2016-07-22
Add Schnobs/NK suggested fairing changes

stratochief66:
Add Schnobs/NK suggested fairing changes

Makes a shorter version of the fairing base.

Makes fairing mass closer to 8mm aluminum instead of 20mm

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
@@ -101,6 +101,27 @@
 	{
 	}
 }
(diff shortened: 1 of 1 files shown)


stratochief66:
Don't merge yet, I'm adding allthechanges.

PhineasFreak:
Would be possible to change the pass from final to a more "mod a mod" friendly one? Like "zzzRO" or something similar?

NathanKell:
I agree on the pass. Also we need to target the PFFE fairing/interstage bases and fairing sides. Might be worth targeting by module rather than partname.

NathanKell:
@stratochief66 I think it's safe to go by module, although we then need to vary it for the flatter interstage I guess.

PR #1316, opened 2016-07-25
Change XLR81 for latest SSTU version

Zarbizaure:
8048 and 8096 models

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_XLR81.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_XLR81.cfg
@@ -1,16 +1,84 @@
-@PART[SSTU-SC-ENG-LR81]:NEEDS[SSTU]:FOR[RealismOverhaul]
+@PART[SSTU-SC-ENG-LR81-8048]:NEEDS[SSTU]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
 	%engineType = Agena
+	@title = XLR81 (Agena) Vacuum Engine (Early)
 	@mass = 0.132
 	@maxTemp = 1973.15
 	@MODULE[ModuleEngines*],0
 	{
-		@heatProduction = 100
 		//engine stats set by RO engineType patch
-	}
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
Eh I was going to leave this out for a while.  I made the change a while back but intentionally didn't apply it to RO yet because the model isn't done, in fact the old model other than being wrong actually looks better.  Given that splitting the engine up is going to break crafts, I figured it was better to at least wait until the new engine was done.

jwvanderbeck:
Also why the change in stats?

jwvanderbeck:
I think if we are going to split them, then we need to do more than this.  The engineType probably needs to be unique and the TF configs handled a bit differently.  Not positive.  Have you tested this and looked at the resulting MM cache? I feel like things aren't going to be exactly correct.

Zarbizaure:
yeah I configured it a bit fast. I just checked if the parts appeared correctly in the editor but I didn't test them; I apologize (however titles are correct).

I set the stats to match the first config avalable in the list, similar to what was done with the RL-10 series.

I'm almost sure that shadowmage planned to improve the model of these parts so that's not the main problem. Plus he added that the old LR81 will be removed in future update (indeed it's just a streched F1-B model).

jwvanderbeck:
Yeah but my point is we really just went from one placeholder model, to one correct but unfinished model and it seems a bit fast to jump on such a breaking change when it doesn't really improve the experience much.  I was holding off hoping he was going to at least get the thing textured, or heck at least an AO bake like his other models.

Zarbizaure:
ok so maybe we could just keep the old seetting too so we'll have support for both? Then we could just choose what to do after the next SSTU update.

jwvanderbeck:
Yeah that is probably a good idea.

stratochief66:
@Zarbizaure should this PR still be considered? I'm not sure if something in SSTU may have changed since this was proposed.

TestFlight integration doesn't have to be included/perfect at first. That can be left for a later PR.

SirKeplan:
Woops, i must have forgot/not noticed this PR when i did my recent XLR81 bits. sorry.

think this engine is all finished in SSTU now. unless i missed anything.

PR #1320, opened 2016-07-26
RO DECQ Soyuz (launch vehicle)

PhineasFreak:
* Add RO support for the DECQ Soyuz pack (launch vehicle).


SirKeplan:
Why is this setting the RD-0124 gimbal range to 45 degrees? does this mod use the RD-0110 model for it or something? real life RD-0124 has no verniers and about 7 degrees gimbal.

PhineasFreak:
After all that testing...

It's nothing, i probably forgot to remove the gimbal modules after copy & pasting the RD-0110 config (and i did a pretty good job of not noticing it considering that i also made a sub - patch "Engine Compatibility" for it).

Anyway, it's fixed now!

PR #1321, opened 2016-07-26
RO DECQ Soyuz (spacecraft)

PhineasFreak:
* Add RO support for the DECQ Soyuz pack (spacecraft).


Theysen:
Hey @PhineasFreak, just a quick question, what Docking port is this compatible with from "stock" RO?

PhineasFreak:
The docking port type is compatible with the Soviet Pack from Raidernick (and with Tantares).

I do not think that stock has a compatible one.

PR #1332, opened 2016-08-06
RN not install craft files by default

raidernick:
-Because KSP doesn't hide craft files like subassemblies when parts are
missing, these should not be installed by default as it clutters the
list.

-This should be made into a separate netkan file as a separate mod that
is recommended to install. For manual install, users should be
intelligent enough to do remove it themselves.


raidernick:
@NathanKell looking for you to take a look at this

NathanKell:
Yep, seems good to me. :)

stratochief66:
@NathanKell I assume this is the bit that requires special attention before release?

Issue #1322, opened 2016-07-30
RD-275 tech and cost rebalance.


shimmythejj:
RD-253/RD-275 Engine. All configurations unlock at the same time, and cost appears to be unbalanced (lower than it should be)

Vens part unlocks at staged combustion

SirKeplan:
Initial cost looks [ok to me](https://github.com/KSP-RO/RP-0/blob/master/tree.yml#L1968) , the later configurations could do with placing though.

shimmythejj:
I might not understand the intended pricing methodology, though comparing the RD-253 to an engine of similar class (thrust) like the LR87 base config, The RD-253 has 1/10 more thrust, 2/3 the mass, significantly better SL and Vac ISP, but has 1/3 the cost of the LR87, 515 vs 1550. The only downside of the RD-253 is that it is a single chambered engine with no built in roll authority.

SirKeplan:
Yeah, the pricing is one of those problems that appear when we have both US and Soviet engines. Soviet engines generally have a much lower price, and a much greater performance. This is partly because of the soviet economy, and the fact that they often mass produced parts much more than the USA did.

see https://github.com/KSP-RO/RP-0/issues/490 for some of our discussion on this.

NathanKell:
For now we can add some nominal extra cost for the later configs, and at least place them correctly.

It's also worth noting that the balancing only comes to the fore with TestFlight, since, well the early Proton was not exactly known for reliability.

PR #1333, opened 2016-08-06
Place/price later variants of RD-253

SirKeplan:
costs are more or less based on NK-33
historically the RD-275 should be TL6, but RD-253s flew in the mid 80s(mir core launch) with a similar thrust upgrade to the 275, so seems reasonable.
EDIT: #1322


SirKeplan:
of course, this does make it a big gap in the tree between the RD-253 and the RD-275, when there is a smallish performance gap

PR #1339, opened 2016-08-12
Tantares configs overhaul, part 1

Niemand303:
I leave for a week, so that what I've done would already be in the mod in case a release happens.


stratochief66:
@Niemand303 hey, sorry for the delay. There are a lot of changes in here. Some I find a little hard to follow, but nothing objectionable. Do you want to take another quick look at this before I merge it?

Niemand303:
Sorry, I won't have time since I leave for France tonight (in 5 hours) :c
You can merge it, however, since I haven't made much progress since my return back to Moscow.

PR #1352, opened 2016-08-25
RO DECQ's SLS+Orion

Asduffo:
RO configs for DECQ's awesomely detailed SLS+Orion craft.
Masses and Delta V are accurate with a precision of a couple of kilos/meters per second.
TAC support guarantees life for 6 kerbals for 21 days just like the real pod. CO2 scrubber included.

DISCLAIMER:
This is a very unlucky mod since apparently DECQ accidentally losed all of his files and so he can't fix it. There are a couple of errors I couldn't manage to fix (or better: WIP i don't have the experience to fix but they don't stop the mod to work). If you happen to know how to fix them, please do it since it's a really beautiful mod. in particular:

0)(not really an issue): the EDS has the same masses and specs of the EUS stage, simply because NASA scrapped it before releasing any important data. Tested it with a block II version of the SLS (RO pyrios) and, strange enough, it managed to send without any problem 150 tons into LEO as it was supposed to.
1)The service module RCS simply didn't work with MMH/MON3 like it was supposed to, so i had to let it work with MonoPropellant.
2)The SPS backup stopped working entirely in RO for some reason
3)There is no RCS on the ICPS and the EDS since they lacked of RCS ports (the exploration upper stage does however and that's the really important stage)
4)The hatch is obstructed, so no EVA. Couldn't find any way to fix it.
5)The chute sometimes works, sometimes it causes the craft to spin terribly and it causes really bad stuff (i took it from another user who contribuited to the mod, anyway)
6)mechjeb won't show delta v statistics for some reason. I can guarantee however that they are the right ones.
7)the j-2x are too long, so when stacked above the core stage they collides with it. BUT this doesn't cause any issue excluding a weird glitch at staging with no explosion/bad consequences. This could be solved by making a stretched interstage (don't have the experience to do so)/just use procedural parts

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/Dummy
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/Dummy
@@ -1,1 +1,0 @@
-Dunny file
--- a/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/Dummy
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/Dummy
@@ -1,0 +1,1 @@
+Dunny file
--- a/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/RO_SLS_ENGINES.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/RO_SLS_ENGINES.cfg
@@ -2,84 +2,12 @@
 {
 	%RSSROConfig = True
 	
-	@title = SLS's core stage
-	@manufacturer = Rocketdyne
-	@description = SLS's core stage with 4 RS-25D/E
-	
 	@rescaleFactor = 1
 	
-	MODULE
-	{
-		name = ModuleEngineConfigs
-		type = ModuleEngines
-		configuration = RS-25D
-		origMass = 26.73
-		modded = false
-		CONFIG
-		{
-			name = RS-25D
-			description = Recycled Shuttle's engines used in SLS till NASA won't run out of them. Max 109% thrust.
-			minThrust = 5912.4
-			maxThrust = 9096
-			PROPELLANT
-			{
-				name = LqdHydrogen
-				ratio = 0.7276
-				DrawGauge = true
-			}
-			PROPELLANT
-			{
-				name = LqdOxygen
-				ratio = 0.2724
-			}
-			atmosphereCurve
-			{
-				key = 0 452.3
-				key = 1 363
-			}
-			
-			%ullage = True
-			%pressureFed = False
-			%ignitions = 1
-			IGNITOR_RESOURCE
-			{
-				name = ElectricCharge
-				amount = 0.5
-			}
-		}
-		CONFIG
-		{
-			name = RS-25E 
-			description = SSMEs optimized in thrust and costs for a single launch. Max 111% thrust.
-			minThrust = 5964.4
-			maxThrust = 9279.6
-			PROPELLANT
-			{
-				name = LqdHydrogen
-				ratio = 0.7276
-				DrawGauge = true
-			}
-			PROPELLANT
-			{
-				name = LqdOxygen
-				ratio = 0.2724
-			}
-			atmosphereCurve
-			{
-				key = 0 450.8
-				key = 1 363
-			}
-			
-			%ullage = True
-			%pressureFed = False
-			%ignitions = 1
-			IGNITOR_RESOURCE
-			{
-				name = ElectricCharge
-				amount = 0.5
-			}
-		}
-	}
+	engineType = SSME
+	engineTypeMult = 4
+	ignoreMass = true
+	
 	!MODULE[ModuleAlternator]
 	{
 	}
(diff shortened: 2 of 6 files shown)


PhineasFreak:
@Asduffo:

* You are missing the critical "RSSROConfig = True" tag required to make the parts compatible with RO.
* The Module Manager pass must be "FOR[RealismOverhaul]" since these are patches that will be shipped with RO.
* Some parts with modular fuel tank definitions indicate negative volume figures (example: Orion CM & SM).
* The RS-25 and RL10 engines do not use the global engine configuration like the AJ10, the J-2X and the RSRM.
* There is an RCS module definition for the Orion SM but the part module does not have one: https://github.com/Asduffo/RealismOverhaul/blob/d04a194e2357626e766cf51aac5ae4f6a3f79b2d/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/RO_SLS_ORION.cfg#L74-L98
* There is an RCS module definition for the Orion LAS but the part module does not include any RCS transforms (the Chaka Monkey LAS includes them as MODEL{} nodes but these were not copied over): https://github.com/Asduffo/RealismOverhaul/blob/d04a194e2357626e766cf51aac5ae4f6a3f79b2d/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/RO_SLS_ENGINES.cfg#L274-L306
* The Orion SM and RCS ports must have fuel cross - feed enabled in order to work with RF (flow definitions are different than the stock "MonoPropellant").

Some other minor points:

* Orion cannot support 6 crew for 21 days but only 4 (i guess NASA never actually made this point clear).
* The Orion SM includes a TACLS "Waste" resource but it is physically impossible to store solid waste inside it (no crew access).
* Some typos with the LF & OX removal from the Orion SM: https://github.com/Asduffo/RealismOverhaul/blob/d04a194e2357626e766cf51aac5ae4f6a3f79b2d/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/SLS/RO_SLS_ORION.cfg#L15-L16
* You can use a MODEL{} node to fix the Z axis scaling of the EUS interstage:

```
@MODEL
{
    %scale = 1.0, Z, 1.0
}
```

And play around with the Z value until the engine does not clip the propellant tanks.

Asduffo:
Hi

You're right about the first 3 points. Sorry but that's my first ever RO work excluding tiktaalik's MEM which was published alongside the mod itself and never branched with RO (ask him if he wants to add it to the official RO too)

As for the other points:

- I used custom data for the SLS core engine and the RL-10 because NASA would use only 2 configs for the RS-25 (D and E respectively) and 2 for the RL-10 (B-2 and C-2) and not all of them, so i thought that it could have been more realistic to give the player some limitations in terms of configurations. Also i know there are configs about the RS-25, but is that the same about the SLS core stage?

- The RCS of the service module is the SPS backup (these 8 engines on the bottom around the main SPS/OMS) and it used to work in the stock game with mono, but not with RO and MMH/MON3. Same for the RCS' quads. But unlike the RCS which i managed to make them work by leaving them use mono, it just stopped working with RO. I left the configs commented in case someone in the future could figure it out (same in the case someone with more experience could have managed to add RCS for the LES). I swear I tried litherally everything. 

- I can't really understand the last point. I think I'm just too noobish.

As for the minor points:

- Do you want me to scale it down to 4 people or just leave it be until NASA doesn't clarify it? Personally I think it's really meaningless :\

- I know I'm going to look stupid, but I swear I always thought that for "wastes" it meant... liquid stuff... and only now that I think of it I realized it's really stupid (language barriers strikes again). Anyway I will move it to the command module.

- Will fix it

- Tried but I just gave up, Since I also had to move the node inside of it to attach the upper stage, every time I had to play with 2 variables and I just couldn't figure it out how to do. Anyway as I said you won't notice anything in the final work (since the j-2x are inside the stage till staging), excluding that weird glitch I said I can't figure out the cause.

Sorry but this was a summer party day so I'm pretty much out of order for the rest of the day. Will start fixing everything either tomorrow or the day after XD

PhineasFreak:
* Removing extra engine configs can be done on a separate MM pass (example: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Engines.cfg#L62-L74). SLS will use the RS-25D for the first flights until the stock is expended. The plan then is for Aerojet Rocketdyne to provide the RS-25E (E from "Expendable - overall same stats as the D variant) that it is easier to manufacture and field. Similarly, the RL10C-2 variant is a man - rated RL10B-2 (used on the DCSS but with extra sensors for detecting anomalies early).
* The SM RCS seems to be a weird case. The part model itself specifies the 8 different thrust transforms but the part config shipped with the mod (SERVICE.cfg) does not have a ModuleRCS{} (or even a ModuleEngines{}). Can you confirm that?
* The LES RCS is a workaround for the CMES LAS and not really applicable directly for this one but ignore it for now since it can be fixed later.
* The RCS blocks require to be "crossfeed - enabled" in order for the propellants to reach them (otherwise they are just blocking the propellants like a valve). This can be done by adding an extra "%fuelCrossFeed = True" definition in the RO patch to enable their fuel cross - feed capability.
* For rescaling parts on one vector: you have to adjust the scaling first and after that to fix the nodes (trying to do both as you found out it is impossible). But is is also an easy thing to do later on.

Nice work on the configs BTW! These mods are sometimes hard to make them compatible with RO.

Asduffo:
Ok I fixed everything you told me (I think), excluding the RL-10 engine for the simple reason that the code you gave me is either not working or I wrote it wrongly.

Anyway I pushed  the SLS core stage engine's configs: can you tell me from that if I'm doing something wrong? Cause the editor simply shows me all the configs (RS-25A/C) even when I told the game to remove them. The only thing it keeps is the mass.

By the way I must have been high or something cause the SPS backup is not working again even in stock and I swear I remember it cleary working (and no it wasn't the rcs).

SirKeplan:
to sort out the RS-25s you have to do it in `:AFTER[RealismOverhaulEngines] `

Asduffo:
Ok done. I hope everything is fine now.
PS: you guys should add an RL-10-C2 config too

SirKeplan:
Yes, we probably should have an RL10C-2. i understand there are ignition system and turbomachinery changes in this engine, though from the performance perspective I’m not sure if there's anything different between the C-2 and B-2.

PhineasFreak:
@Asduffo taking a closer look at the model of the Orion SM i noticed that:

* It is missing any kind of engine or RCS thrust transforms for the auxiliary (backup) engines. This would not work even in stock.
* It includes 16 RCS transforms at exactly the same positions as if the "ORION_RCS" model was at first part of the SM model and later separated from it (probably for modularity).

We can easily add working transforms for the auxiliary engines in the future.

@SirKeplan ULA lists the RL10C-2 upgrades as:

``
The RL10C-2 engine will incorporate all improvements from the RL10C-1, including an upgraded  redundant ignition system to improve reliability, changes to the engine plumbing to improve starting  operations, a propellant valve design update, and a number of improvements previously qualified under the Assured Access to Space program including a revised gear train and seal improvements.
``

And also:

``
Additionally, the RL10C-2 is intended to be qualified to operate with active Mixture Ratio control, a capability available on Atlas/Centaur missions dating back to 1965. This feature, enabled on Delta IV   by the addition of Common Avionics could  result  in  a performance  improvement  of  up  to  200  lb for  certain  Delta  missions. The RL10C-2 will continue to use the 3-segment extendible nozzle currently used on the RL10B-2. The C-2 will look virtually the same as an RL10B-2, with slight changes to the Ignition and Engine Instrumentation Boxes and realignment of some of the large plumbing. 
``

Source: [Delta IV User's Guide (June 2013)](http://www.ulalaunch.com/uploads/docs/Launch_Vehicles/Delta_IV_Users_Guide_June_2013.pdf)

In sort nothing new regarding improvements that can be easily seen or used in RO.

SirKeplan:
Yeah, then as it stands adding a C-2 will just be adding an exact clone of a B-2(apart from maybe price?), if we simulated more things like mixture ratio control and ignition mechanisms, then it would be worth it.

maybe we should rename the RL10B-2 to "RL10B-2/C-2" ?

Asduffo:
http://www.spacepropulsion.org/uploads/2/5/3/9/25392309/spaceaccess2014-25.pdf

according to this 2014 document (from which I took the EUS data) it will have a thrust of 110Kn and an Isp of 462 in contrast of 465.5 of the RL-10-C1-B

Anyway I don't understand: are my configs ready ready to go or is there something more to do? Will they be published with the next RO update and I'll become famous? XD

PhineasFreak:
@Asduffo these changes are minor at best (delta VACIsp of 3.5 sec, delta VAC Thrust of 93 N) and can be easily be ommited (as it was done with some of the RS-25 versions). Granted, if Real Fuels ever implement a way to change the propellant utilization on the fly then it would matter for the C-2 to have this capability versus the B-2 (that it would not).

Your configs might be ready but this does not mean that they have been checked 100%. It takes some time for the RO collaborators to test the parts in - game and make sure that nothing is broken or gets broken. Be patient.

stratochief66:
@Asduffo hey, I'm finally getting around to reviewing this, sorry for the delay.

The <.> parts are in fuel tanks, when they should probably go in the <> category.
The solar panels appear to have 2 toggles, Open and Extend. One of them caused it to not generate power,
It still appears to have a reaction wheel on the crew module, that will definitely need to be removed.
I also see it not generating dV data through mechjeb, very strange. We will put some -top minds- on it.
There are a few other oddities that can probably be ironed out later, particularly if you are interested in continuing to help @Asduffo.

It took me about 20 minutes to assemble a Block 1 SLS with Orion, which unfortunately is a side effect of the mod being pretty and having many many parts.
Would it be possible for you to create some craft files for this mod in RO and share them via Dropbox or something? It would be much appreciated and will make things a lot easier to test and review (and play with!)

stratochief66:
Certainly an improvement from the status quo; this PR can be looked to if someone wants feedback for further tuning these configs in the future.

PR #1355, opened 2016-08-28
Move TAC LS LifeSupport.cfg file to new location. 

JPLRepo:
Can be turned into an MM config file if so desired.


JPLRepo:
I have moved the LifeSupport.cfg file out of PluginData so that it can be MM'd to your hearts content.
I have removed settings values from this file into the save file persistent.sfs file so it should ONLY change when the user changes the usage settings, etc. NOT window settings etc. These are no longer in this file.

raidernick:
merging this to make RO compatible with current tacls, pending future changes, RO will need to be updated again.

PR #1361, opened 2016-09-01
CMES AerojetKerbodyne updates/fixes

PhineasFreak:
Continuation of the CMES updates, this time for the "AerojetKerbodyne" parts.

Changelog:
- Add sources for some of the part configs.
- Round the mass of the DCSS 4 truss adapter.
- Adjust the inert mass of Orion MPCV linear thruster blocks.
- Add Global Engine compatibility for the AJ10-190 engine.
- Make the RESOURCE removal a bit more bulletproof.
- Update the part titles and the descriptions.
- Fix formatting (tabs, false spacing).

[Alternate diff ignoring whitespace changes.](https://github.com/KSP-RO/RealismOverhaul/pull/1361/files?w=1)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CMES-WIP/RO_CMES_AerojetKerbodyne.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CMES-WIP/RO_CMES_AerojetKerbodyne.cfg
@@ -112,7 +112,7 @@
 }
 
 //  ==================================================
-//  Delta Cryogenic Second Stage (DCSS) engine truss.
+//  Delta Cryogenic Second Stage (DCSS) equipment bay.
 
 //  Dimensions: 3.1 m x 0.8 m
 //  Inert Mass: 100 Kg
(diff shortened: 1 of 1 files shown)


SirKeplan:
Something that is needed on command parts is a mechjeb core, since the baskettruss is not usually the command part, it seems it needs re-adding

PhineasFreak:
@SirKeplan i have made a PR for the RO global config to now apply AFTER[RealismOverhaul] instead of BEFORE[RealismOverhaul]: https://github.com/KSP-RO/RealismOverhaul/pull/1423

SirKeplan:
Ah, yes it does. my bad for not testing with latest master.

SirKeplan:
Only other thing is the 4 metre DCSS truss doesn't really fit properly, but probably nothing can be done about that?

PhineasFreak:
Yes, unfortunately there is not a proper truss for the 4 meter variant. This is something that it is bothering me since i ever started doing the configs but i never found an overall good solution.

SirKeplan:
Only option is to pester YANFRET then :stuck_out_tongue_closed_eyes:

PR #1363, opened 2016-09-02
Update NK-19 to be similar to the NK-9V rather than NK-39.

ahmedcharles:
The NK-19 was changed as part of: https://github.com/KSP-RO/RealismOverhaul/commit/26f1c285a8d4c64b1bb53af8aec5a6c0f313c772

I've spent some time researching and isn't any evidence to suggest that the NK-19 is anything but a more reliable NK-9V, so the thrust dropping drastically doesn't make sense and I can't find a reference to it's thrust being any different than the NK-9V. Though, there are sources that suggest that both of them have lower thrust. I also haven't found any information suggesting that it throttled at all or that it had a burn time of 450 as opposed to 443.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/NK9V.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/NK9V.cfg
@@ -68,7 +68,7 @@
 		{
 			name = NK-21
 			description = NK-9V rerated for N1 Block V use. No gimbal, differential throttle.
-			maxThrust = 400 // b14643
+			maxThrust = 400 // b14643.de
 			minThrust = 240 // assume 60% throttle.
 			heatProduction = 205
 			// 2.5 O/F mass ratio (b14643.de)
@@ -106,8 +106,8 @@
 		{
 			name = NK-19
 			description = NK-9V rerated for N1 Block G use. Gimbal, throttle.
-			maxThrust = 400
-			minThrust = 240
+			maxThrust = 451.1 // b14643.de
+			minThrust = 260
 			heatProduction = 205
 			massMult = 1 // I...guess?
 			// 2.5 O/F mass ratio (b14643.de)
@@ -124,7 +124,7 @@
 			}
 			atmosphereCurve
 			{
-				key = 0 353 //assuming same as NK-31
+				key = 0 345 // b14643.de
 				key = 1 240 // astronautix, GR-1 entry
 			}
 
@@ -143,8 +143,8 @@
 		{
 			name = NK-39
 			description = Improved for N1F Block V. No gimbal, differential throttle.
-			maxThrust = 400
-			minThrust = 240 //FIXME guessing 60% min throttle
+			maxThrust = 400 // b14643.de
+			minThrust = 240 // assume 60% throttle.
 			heatProduction = 205
 			massMult = 1.09375 // 700kg http://www.scribd.com/doc/7362263/Russian-Liquid-Propellant-Engines#scribd
 			// 2.6 O/F mass ratio (b14643.de)
@@ -180,7 +180,7 @@
 		{
 			name = NK-31
 			description = Improved for N1F Block G. Relightable. Gimbal, throttle.
-			maxThrust = 402 // b14643
+			maxThrust = 402 // b14643.de
 			minThrust = 240 // assume some level of throttle, engine design was capable, and LH2 replacement engine for this stage had throttle.
 			heatProduction = 205
 			massMult = 1.128125 // 722kg http://www.scribd.com/doc/7362263/Russian-Liquid-Propellant-Engines#scribd
@@ -284,6 +284,6 @@
 		ignitionReliabilityEnd = 0.99
 		cycleReliabilityStart = 0.94
 		cycleReliabilityEnd = 0.985
-		techTransfer = NK-9V,NK-9,NK-19,NK-21,NK-39:50
+		techTransfer = NK-9,NK-9V,NK-19,NK-21,NK-39:50
 	}
 }


SirKeplan:
I gave the NK-19 that thrust as it seemed to make sense that it would not have more thrust than it's later evolution the NK-31. Also i just assumed it would have the same thrust is the variant used on the N-1 3rd stage(NK-21)

This document has performance for the NK-39 and 31, and mentions that it was based on the NK-9V, but unfortunately doesn't mention the NK-19.
https://www.scribd.com/document/7362263/Russian-Liquid-Propellant-Engines#scribd
However this section would seem to suggest there was thrust downrating specificly for the N-1
>After development works on GR-1 were
discontinued, the NK-9V version with extended service life was used in the upper (the
3rd and the 4th) stages of the super heavy Lunar
N-1 carrier rocket. In order to suit for the new
launch vehicle, the engine power was degraded
through a reduction of the turbopump operation intensity.

SirKeplan:
However this source I have here would seem to agree with you.
http://www.k204.ru/books/Aviadvigatel2.pdf (page 211)
it lists the НК-19(NK-19) as having significantly more thrust than the НК-21

It could be that you are correct, it's often awkward to get good sources when looking at old soviet engines.
In any case, it would seem a good idea to add an NK-21 engine to this as well.

SirKeplan:
Also the throttling seemed reasonable assuming the 3rd stage uses differential throttling like the first 2 stages, and the 3rd and 4th stage engines are very similar.

Out of interest what are your sources for this?

ahmedcharles:
I agree that it's hard, since as far as I can tell, the upper stage engines were never tested/flown, since they all blew up before being ignited on the test flights. So, my primary concern is mostly game play, since it's sort of hard to justify realism when there aren't clear, hard numbers involved.

As for throttling the NK-19, I couldn't find a source, so I just assumed it didn't.

I watched a documentary about the N1 and mostly NK-33 first stage engines and while it didn't talk about stats at all, it did talk a lot about their mindset and methodology for building the engines.

If you consider that the NK-19 seems to have been active between 1962-1967 (http://www.lpre.de/sntk/ I don't read Russian, but I'm assuming that's what that column is) and the first N1 launch was 1969 (https://en.wikipedia.org/wiki/N1_(rocket)), the NK-19 would never have flown on the N1 at all.

Based on the dates and rockets listed http://www.lpre.de/sntk/, it would make the most sense that the original N1 designs were using:

A: NK-15
B: NK-15V
V: NK-21
G: NK-19

But before the first flight they were switched to:

A: NK-33
B: NK-43
V: NK-31
G: NK-39

So, I'm thinking that the reference to the turbopump operation intensity decrease could've come with teh NK-31 and NK-39, which every source I've seen agrees with. The NK-31 definitely throttles though the NK-39 seems to gimbal. I have no idea if it throttles too.

As far as suggestions go, I might suggest:

NK-9V -> (NK-19, NK-21) -> (NK-31, NK-39)

The only issue here is that the NK-19 is better in every way than the NK-21 other than size, probably. Since you have to fit 4 of them, it would make sense that they were smaller. The other difference could be that the NK-21 throttled but the NK-19 did not.

As an aside, the fuel ratios don't seem to be correct either. Every source listed above seems to suggest that the NK-9V/19/21 have a O/F of 2.5 and the NK-31/39 have an O/F of 2.6. That would result in 0.71429 vs 0.28571 and 0.72222 vs 0.27778, respectively. Note, this would really make them play different than American kerolox engines (which they should, since the Americans viewed high oxidizer mixtures to be too risky to use).

I'm updating my pull request with my suggested changes.

NathanKell:
KSP ratios are volume ratios not mass ratios. That PR claims the engines use 2.5 *liters* of LOX per liter of kero, not 2.5 kg of LOX per liter of kero. ;)

See https://github.com/BobPalmer/CommunityResourcePack/blob/master/FOR_RELEASE/GameData/CommunityResourcePack/CommonResources.cfg#L1307 and https://github.com/BobPalmer/CommunityResourcePack/blob/master/FOR_RELEASE/GameData/CommunityResourcePack/CommonResources.cfg#L1255 for densities per liter.

ahmedcharles:
@NathanKell I'll update based on volume... though the documents don't say that the ratio is kg either... but I'd imagine that to be the case. (Why does rocket science have to be so hard...)

SirKeplan:
@ahmedcharles on the lpre.de site those years are "Years of development" not the years it was active.

A little correction, the NK-15 NK-15V NK-21 and NK-19 were for the first 4 flights of the N-1,

ahmedcharles:
@SirKeplan Sometimes (actually just today), I wish I could read Russian.

SirKeplan:
@ahmedcharles on the lpre.de site those years are "Years of development" not the years it was active.

A little correction, the NK-15 NK-15V NK-21 and NK-19 were for the first 4 flights of the N-1, the NK-33s etc were improved versions of these engines designed to be more reliable and durable, i have a document that lists the improvements if i find it.

about volume vs mass ratios, you'l find whenever the talk is about the rocket engines the ratio is a ratio of mass, ksp is just a little awkward using volume ratios when you usually talk about mass ratios.

SirKeplan:
@ahmedcharles same, though for little things like that google translate is enough for me to work out the meaning.

ahmedcharles:
Ok, that should be better, in that it uses volume ratios (they are still slightly different than before, not sure why).

SirKeplan:
From what I understand
N-1
Blok V: 	NK-21
Blok G:	NK-19

N-1F
Blok V: 	NK-39
Blok G:	NK-31

Now this does make the numbers a little counter-intuitive, but i think this is right.

This is why the NK-31 has more mass, it's for the gimbal assembly, unless https://www.scribd.com/document/7362263/Russian-Liquid-Propellant-Engines#scribd is wrong of course

ahmedcharles:
I noticed this too. Wikipedia says that the NK-31 was used on the Blok V, though I don't think it's perfect. I agree that the stats listed in that link mean that the NK-31 is for the Blok G or they just switched the names. I think going from NK-21 -> NK-31 and NK-19 -> NK-39 makes more sense (and I doubt the Russian makers of the N1F are going to complain much about us getting the names swapped).

Anyways, other than the names potentially being swapped, I think the stats are better this way?

On a side note, I just noticed that the NK-15/33 both gimbal and throttle, though they both have a config that doesn't gimbal. This seems to make them somewhat 'OP' compared to the originals. Shouldn't the version with the gimbal either not throttle or weight more?

NathanKell:
To the best of my knowledge, yes, 19->31 and 21->39.

SirKeplan:
About the NK15/33, probably should give gimbaling variant slightly more mass. Maybe baseing of AJ26 massis good.

NathanKell:
seems reasonable

ahmedcharles:
How do we like the changes as they stand for this PR?

stratochief66:
@ahmedcharles , You broke this PR up/included the changes elsewhere now, yes? Should we close this PR, or is there still something you want to do with it?

ahmedcharles:
I'll comment more when it needs more review or I'll close it. I'm still going through the changes to make sure I got everything.

PR #1364, opened 2016-09-02
Fix RO engine categorization

PhineasFreak:
The engine categorization patch (aimed to fix the placement of some of the RSB parts) did not work as expected as it broke some SXT and VSR solids so:

* We revert the RO_Engines to the previous filtering logic.
* ~~We create some "workaround" patches to force the RSB parts to their correct categories.~~


PhineasFreak:
Reverted the RSB changes since they will be done as part of an RSB "updates/fixes" PR.

PR #1365, opened 2016-09-02
RSB generic updates/fixes

PhineasFreak:
Some generic (i.e. not tied to a particular release) updates and fixes for Real Scale Boosters.

Changelog:

* Improve RT and RealFuels patching and handling.
* Make the RCS patches compatible with the new RCS FX module.
* Fix/update the manufacturers and descriptions of some parts.
* Fix various cases where the power consumption of the avionics would be incorrect if RT is installed.
* Fix various wrong tank definitions (Ares I, Ariane 5 ESC-A, Centaur, DCSS, Saturn S-II, S-IVB).
* Fix the category placement of some of the parts that include ModuleEngines{} but are not engines per se.
* Update the RCS thrust values of the Ares I interstages.
* Fix the RCS thrusters of the Ariane 5 EPC and ESC-A stages.
* Improve the Carrack interstage RCS thruster values (same to the ones used by Ares I).
* Remove redundant decoupler FX definitions (now part of RSB itself).
* Improve the operation of the Saturn APS and Centaur RCS thrusters by having different thrust values for each operation (ullage and attitude control).
* Make the PSLV PS1 stage propellant tweakable.
* Convert tabs to spaces.
* Remove trailing spaces.
* Group the source links together to eliminate duplicates and simplify searches.
* Add descriptions at the sources for quick reference
* Other typo additions and fixes.


GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
```diff
@@ -268,34 +270,29 @@
...
-    @mass = 4.375
+    @mass = 4.45
     @crashTolerance = 14
     @breakingForce = 250
     @breakingTorque = 250
     @maxTemp = 773.15
     %skinMaxTemp = 873.15
 
-    @MODULE[ModuleRCS]
+    //  56 N Gas Hydrogen (GH2) attitude control thrusters for each axis (using liquid hydrogen directly for now).
```
stratochief66: Does it really? Very interesting. As long as you make the proper conversion between LH2 & GH2 densities and masses, usingn LH2 should be fine.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
```diff
@@ -268,34 +270,29 @@
...
-        {
-            name = Nitrogen
-            ratio = 10.0
-            ignoreForIsp = True
-        }
-
         @atmosphereCurve
         {
-            @key,0 = 0 216
+            @key,0 = 0 180
```
stratochief66: You may have to do something funny with this to account for the fact that it actually consumes GH2, while you have it consuming LH2. I'll have to think about that.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Atlas.cfg
```diff
@@ -403,22 +422,11 @@
     MODULE
     {
         name = ModuleFuelTanks
-        type = ServiceModule
+        type = BalloonCryo
```
stratochief66: I'm not sure if this tank qualifies as a BalloonCryo. IIRC, BC has better insulation than cryo, which I don't know if centaur /dcss really ever went for. Did you check how much boiloff you get in practice with this tank type selection vs just Balloon?

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
```diff
@@ -714,15 +697,45 @@
...
-    !MODULE[ModuleRTAntenna*]{}
+    !MODULE[ModuleRTAntenna*],*{}
 
     !RESOURCE,*{}
+
+    //  Avionics batteries 1.4 kWh.
+    //  Can support the DCSS for flights up to 7 hours in duration.
+    //  Assumes that the electricity consumption of the DCSS avionics is 200 W.
+
+    RESOURCE
```
stratochief66: I believe it is preferred that resources like these reside in a module Tank { }, but I'm not sure if that would render this unacceptable. Without being nested in a Tank, the volume couldn't be cleared and reassigned by the user.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
```diff
@@ -746,25 +759,16 @@
...
     %skinMaxTemp = 873.15
 
+    !MODULE[ModuleFuelTanks],*{}
+
     MODULE
     {
         name = ModuleFuelTanks
-        type = ServiceModule
-        volume = 58782.54
+        type = Cryogenic
```
stratochief66: Same sort of question as above about appropriate tank choice. One way to tell is usage, how long did this stage ever coast and what kind of boiloff did it experience then? Also, a stage with a more insulated tank generally has a higher dry mass than the alternative.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
PhineasFreak: Yep! From the Ariane 5 User's Manual (page 204):

```
During the powered flight, the attitude control in pitch and yaw is ensured by the gimballing of the nozzle, and 4 GH2 thrusters are used for roll control. During the ballistic phase, roll, pitch and yaw control uses 4 clusters of 3 GH2 thrusters. 2 GO2 thrusters are also implemented for longitudinal boosts.
```

So even now it is not accurate enough (we cannot use different propellants for different RCS transforms on the same part).

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
PhineasFreak: I am awaiting the RealFuels release for KSP 1.2. @Starwaster mentioned that he will add the capability for cryogenics themselves to produce gas products (LqdHydrogen --> Hydrogen) without the need of any converter modules.

This could also come handy with CMES to create cryocoolers for the Mars Transfer Vehicle.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Atlas.cfg
PhineasFreak: The FASA RO configs have a link for the Centaur V boiloff and long duration missions: http://www.ulalaunch.com/uploads/docs/Published_Papers/Upper_Stages/CentaurUpperstageApplicabilityforSeveralDayMissionDurationswithMinorInsulationModificationsAIAA20075845.pdf

And the config for it mentions a "BalloonCryo" tank type: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Titan.cfg#L270

Apart from that, no, i did not check (and it is probably my fault since i the sources that i used might be inaccurate and i should have validated the results).

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
PhineasFreak: I had them originally like that but theoretiacally this requires a "ServiceModule" tank and because the parts are not modular then the DCSS tanks were also the same type (really bad!).

Other thoughts on that:

* A user might remove the tanks, rebuild them but forget the batteries (not so easy to check because of the nature of the part assembly).
* No actual reason to remove the EC since it is 100% required to operate the stage (apart from adding separate battery modules).

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
PhineasFreak: Considering that the DCSS has almost the same capabilities as the Centaur V then:

* We exclude any "ServiceModule", "Structural" or "Fuselage" tank types.
* Being able for long duration flights indicates that the tank type cannot be "Default".
* Being a structurally stable stage indicates that the tank type cannot be "Balloon" or "BalloonCryo".
* It is also insulated by the same insulation type as the STS ET and the SLS core stage (using foam applied by a spray machine) and i think that is what "Cryogenic" is modelled after.

The LiquidHype Delta III configs also use the same tank definition but again i did not validate if it is actually 100% correct: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/Liquidhype/RO_LqdH_Delta3.cfg#L28

Starwaster:
FYI, STS ET couldn't possibly be cryo tank since its insulation was only SOFI. It would have to be SOFI + MLI.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
stratochief66: Good idea. I was not aware of that upcoming feature. I look forward to that as well :)

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Atlas.cfg
stratochief66: Ahh, right, I remember reading those docs long ago. Yeah, BalloonCryo should be good then. Possibly a future review to compare the on orbit staying time of our configs to the real ones will be in order to confirm or lead us to fine tune such things.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
stratochief66: Yeah, good point. That seems like probably the best current way to deal with that issue then. Being able to contain 2 tank types would be a good future feature for ModuleFuelTanks, but is definitely not currently possible.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Delta.cfg
stratochief66: You are correct.

PhineasFreak:
@Starwaster oh, i thought that "Cryogenic" only referred to the type of tanks that had foam insulation. Is "Default" the correct tank definition then?

Speaking about tank definitions, is there any guide for the various cryo capabilities or intended usage of the tanks?

Starwaster:
@PhineasFreak Crap, let me clarify what I said. I typed it on my phone and that gets me terse out of necessity and I left part of what I meant to communicate out. I was primarily referring to differentiating between hydrogen tanks on the different tank types.

Historically, LH2 tanks have ALWAYS been insulated. You just can't NOT insulate those tanks. LOX tanks in the old days were not insulated and that was less of a problem because 1) higher boiling temp and 2) the outside of the tank would actually form ice that had an insulating effect. LH2 tanks didn't get ice formation because the air would liquefy and run off which increased the heat leakage into the tank.

So there's a little discrepancy in the Default and Cryogenic types where Default was supposed to be Saturn tanks but LOX got insulation and Cryogenic got MLI simulated by having the skin-internal conductivity reduced in addition to the insulation applied. 

So that's what I meant by what I said. Ultimately, just ignore it if you want :)

PhineasFreak:
@Starwaster ok, thank you very much for clarifying it!

PR #1366, opened 2016-09-03
Update NK-15/15V/33/43 to not have free gimballing.

ahmedcharles:


stratochief66:
This PR does not do what your title says. You add figures to mixing ratios with no citation for the source.
You change the price of unlocking a config without explaining or citing why.

IIRC, the N-1 was steered through differential throttling, since the engines were in fixed positions. Since no mechanic exists in any mod to achieve this, we made a gimballing version, similar to how we gave the A-4 engine gimballing, even though it uses carbon thrust vanes for steering.

At some point, IIRC we agreed that mixing ratios should have 3 or 4 sig figs, I don't recall which. Pinging a few who might recall?
@Schnobs . @SirKeplan , @raidernick  ?

SirKeplan:
Its 4 sig figs I'm pretty sure. Anything more is pretty pointless. 

I would suggest dialing back some of these changes in here. And i suggest both variants should have throttle.  As stratochief says some of these changes need more justification and sources.

SirKeplan:
Also you switched round which one gimbals and which doesn't. From i historical perspective i understand this makes sense,  but it just seems like it could be slightly annoying from a game play perspective.

ahmedcharles:
I've updated the change.

I'd like to split feedback into two categories (assuming there's interest in doing this at all).

1. Let me know if I'm missing something factual or if something's not sourced or not accurate, based on the assumptions stated in comments.

Note: About the precision of the propellant ratios, I noticed one that had 5 digits and just made them all 5 digits. I'd rather have something consistent than not. And I didn't want to reduce accuracy for the one that was already 5. If you want, I can consistently make the entire codebase use 4 or 5, but if it's not already consistent, I don't see the point in debating it as part of this change. Things aren't getting worse due to it.

2. Assuming the numbers for the 'real' engines are fine, the other choices are purely game design tradeoffs, so comments on those lines should be separate. Everyone has different game design sense, so the choices here are more subjective.

Note: I found it amusing that this PR got comments that mention both game design and sources to real engines. It seems like you can't really get both at a philosophical level, since they have different ultimate goals. And since I'm relatively new, I can't possibly know where the existing line is and it's not obvious that the project has a consistent stance here either. Though, if it does, I'd like to know, since it would make proposing changes easier.

Temeter:
Hey, it's cool just to have the discussion! This is also what forms these kinds of mod.

Historically it's a bit tricky, here everything depends on freedom. RO does have precedent of following rather logical developments measured on tech level, rather than purely realistic spaceflight. That's why we got the ambitious M1 engine and nuclear engines, under the assumption those projects weren't cancelled.

There is precedent of upgrades in this fashion, if decades late: The Aj26-62 of the antares was a modified NK33, and they included new gimbaling system. Same goes for the Soyuz 2-1, using a gimbaled NK33A, later variants are going to use an uprated variant called NK33-1. To my understanding, all of these are take from the stockpiles that were originally built for N1 rockets that never flew. I think they also were throttled during flight, but that might be wrong and I can't find a source...

SirKeplan:
OK to close this issue?

stratochief66:
@SirKeplan, yeah I think so. I think it is worth checking if there is an open 'Issue' regarding this set of engines, and if not, creating one. Some of the stuff discussed here is worth hashing out into as 'official' a policy as we can muster, or at least having a proper static location to discuss it.

@ahmedcharles We generally discuss the trade-off & balance on a case by case basis in the RO IRC chat, because it can be complex and go a different way depending on specific circumstances. For example, an engine without any gimballing at all is basically useless in KSP. So, if an engine is to have no gimballing, there better be a damn good reason for it (history sometimes isn't enough), like with the Proton stages the non-gimballing engines have verniers that go on the same stage to provide control.

As I stated previously, some historical engines didn't strickly gimbal, but they had differential throttling which we currently cannot simulate, and so we sub in gimballing to match the historical capability as best we can.

PR #1367, opened 2016-09-03
Gimbals for Puff / Aerospike (Shuttle OMS / J-2T)

Schnobs:
Shuttle OMS should have 7 degrees IIRC.

J-2T... according to Ferram there's no reason why spikes can't gimbal. I don't know what the proper/real value would be, but certainly not zero.

ETA:  http://www.alternatewars.com/BBOW/Space_Engines/ADP_Vol_I_JAN-1968.pdf, p. 40 (PDF pagination)
The document doesn't call it J2T, but by all appearances that's the engine it's about.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1962,7 +1962,7 @@
 	{
 		name = ModuleGimbal
 		gimbalTransformName = thrustTransform
-		gimbalRange = 3		//FIXME, real value unknown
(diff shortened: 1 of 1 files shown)


SirKeplan:
For the AJ-10-190 the gimbal declaration should really be in in the `:FOR[RealismOverhaul]` bit, then you can also take out the `gimbalRange = 7` and response speed lines and put the range in the `AJ10_190_Config.cfg` file

SirKeplan:
@Schnobs poke? gimbal ranges for both engines should be set in the global configs.

Schnobs:
Closing this one, new PR:
https://github.com/KSP-RO/RealismOverhaul/pull/1389

PR #1382, opened 2016-09-16
Update mixture ratios for the NK family of engines.

ahmedcharles:


SirKeplan:
nice, useful having the ratios in comments.

PR #1383, opened 2016-09-17
Update the fuel ratio for the AJ26-62.

ahmedcharles:


SirKeplan:
Something felt off about the b14643.de number, so i found the users guide for the rocket, according to that it's 2.7, which is probably a more trustworthy source. also notice the other stats like isp are slightly higher. which is interesting.

https://web.archive.org/web/20140115161002/http://www.orbital.com/NewsInfo/Publications/Antares%20UG.pdf

ahmedcharles:
@SirKeplan There you go. Updated.

PR #1385, opened 2016-09-17
Update NK-19 burn time.

ahmedcharles:


SirKeplan:
@jwvanderbeck @NathanKell what's the guide here with setting TF burn times? it doesn't feel right setting the `ratedBurnTime`  to the same as a stages burn duration. is there some kind of general margin we add?

jwvanderbeck:
Ideally you find the real manufacturer stated value (which is always lower than how long it COULD burn for) but that is often extremely hard to find, so in general most engines were simply set up by finding the longest example of continuous burn time (IE if the engine was used in more than one configuration or LV find the longest) and then adding some X value of padding on top of that.  It isn't really much of a science sadly, more of a gameplay tweak so you take your best swing then refine it if needed.

SirKeplan:
@jwvanderbeck makes sense, closing this PR, unless someone has a better figure.

ahmedcharles:
I guess I assumed it was more of a science than it actually is.

jwvanderbeck:
We add some extra buffer to avoid the problem of forcing recreations on the player.

ahmedcharles:
That makes sense.

PR #1386, opened 2016-09-18
Move RD-58 to propulsionSystems node

SirKeplan:
This makes the RD-58 unlockable at the same time as the NK-15(in real life RD-58 flew before, in 1967), and it will no longer be unlockable at the same time as the improved RD-58M. unfortunately this makes it come along at the same time as the 11D33M now.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD58_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD58_Config.cfg
@@ -134,10 +134,10 @@
 			name = RD-58
 			maxThrust = 83.36
 			minThrust = 83.36
-            		heatProduction = 100
+			heatProduction = 100
 			massMult = 1.534
 
-			techRequired = precisionPropulsion      //  Tier 3 of staged combustion, TL4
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/Engine_Configs/RD58_Config.cfg
```diff
@@ -134,10 +134,10 @@
 			name = RD-58
 			maxThrust = 83.36
 			minThrust = 83.36
-            		heatProduction = 100
+			heatProduction = 100
```
stratochief66: indenting error makes the Git panda sad

stratochief66:
Why is that unfortunate? I need to cuddle up with bnumz in order to get russian engines straight every time. Also, is it best to have those tech changes here, or should they be in RP-0? It is unclear to me, what you proposed may make the best use of the centralized engines config system.

GameData/RealismOverhaul/Engine_Configs/RD58_Config.cfg
SirKeplan: we have lots of cases where there's a mix of spaces and tabs an a file :/

SirKeplan:
well, maybe not unfortunate, just a means people will often skip using the 11D33M config

PR #1389, opened 2016-09-18
Gimbals for Puff & Aerospike (Shuttle OMS and J-2T).

Schnobs:
Both have no gimbals in stock, needed a MODULE stanza.
Gimbal range added for J-2T (+source);
OMS already had a range it set, but it went unused until now.

New attempt at https://github.com/KSP-RO/RealismOverhaul/pull/1367

source for J2:  http://www.alternatewars.com/BBOW/Space_Engines/ADP_Vol_I_JAN-1968.pdf, p. 40 (PDF pagination)
The document doesn't call it J2T, but by all appearances that's the engine it's about.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
@@ -30,9 +30,9 @@
 
     @MODULE[ModuleGimbal],*
     {
-        @gimbalRange = 6.0
-        %useGimbalResponseSpeed = True
-        %gimbalResponseSpeed = 16
+        gimbalRange = 6.0
+        useGimbalResponseSpeed = True
+        gimbalResponseSpeed = 16
     }
 
     MODULE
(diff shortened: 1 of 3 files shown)


GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
```diff
@@ -30,9 +30,9 @@
 
     @MODULE[ModuleGimbal],*
     {
-        @gimbalRange = 6.0
-        %useGimbalResponseSpeed = True
-        %gimbalResponseSpeed = 16
+        gimbalRange = 6.0
```
PhineasFreak: Why did you remove the MM modifiers? If an engine module already has them then the final output will include duplicates of the parameters.

PhineasFreak: Yep, that would be the perfect solution. You could even use a value of 0 for them (since it is only a placeholder).

SirKeplan: did you try using the `%` modifier? that will edit or create, depending on if it already exists.

Schnobs:
Percent sign works on AJ10 (tested). Added them to J-2 as well (untested).

SirKeplan:
Are you sure you tested the AJ10? had to remove the line that was disabling the gimbal.

PR #1392, opened 2016-09-20
NK-33/43 reliability, NK-9V family tweaks.

SirKeplan:
Feedback on the TF stuff would be good, to make sure i have applied the reliability numbers in the correct way.

Other things are generally minor like fixing recently noticed inconsistencies.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/NK33_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/NK33_Config.cfg
@@ -1,7 +1,8 @@
 // Nk15/33 Series
 // Bobcat
 // ------Sources--------
-// http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20000088626.pdf, http://www.scribd.com/doc/46957813/Aviadvigatel
+// http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20000088626.pdf
+// http://www.scribd.com/doc/46957813/Aviadvigatel
 // http://www.b14643.de/Spacerockets_1/East_Europe_2/N-1/NK/index.htm - dubious
 // https://web.archive.org/web/20130703154050/http://www.spaceandtech.com/spacedata/engines/nk33_specs.shtml
 // http://www.lpre.de/sntk/NK-33/index.htm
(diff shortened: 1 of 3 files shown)


GameData/RealismOverhaul/Engine_Configs/NK33_Config.cfg
```diff
@@ -298,11 +301,11 @@
 	TESTFLIGHT
 	{
 		name = NK-33
-		ratedBurnTime = 240
+		ratedBurnTime = 365 // [10]
```
stratochief66: All uses of this engine I could find burned from roughly 113 seconds (N1 1964) to 139 seconds (Kistler K-1), what is the 365 from?

IIRC, Agathorn's standard is to add ~5% to the longest burn time you find to give it a little leeway?

Same issue exists with the ratedBurnTime on other versions.

Ahh, you cite source [10]. Can you help me find a link to the full source? It appears to be from 1991, which may not be the best source to lean on when we unlock it in the 60s, but I understand we take what we can get for sources. I'm definitely willing to read the source and give 365s a fair shot :)

GameData/RealismOverhaul/Engine_Configs/NK33_Config.cfg
```diff
@@ -311,11 +314,11 @@
 	TESTFLIGHT
 	{
 		name = NK-33-Original-NoGimbal
-		ratedBurnTime = 240
+		ratedBurnTime = 365
```
stratochief66: This burn time appears high, what is it from?

SirKeplan:
@stratochief66  burn time is in this doc https://www.dropbox.com/s/cr8zzil9c8lxsx6/EVOLUTION_OF_THE_NK-33_AND_NK-43-anisimov1997.pdf?dl=0

365 seconds is mentioned as the specification requirement in certification tests. though that is actually over 3 ignitions total. there are other longer tests they were subjected to as well.

stratochief66:
Yeah, but by that logic we could also 625 or 635 seconds of testing across various burn lengths on the same engine, but I don't think we should do that either.

We should probably discuss a realistic and meaningful way to spec engines intended for re-ignition and re-use, but first stage engines like this shouldn't be rated beyond Agathorn's standard of "longest used/intended burn time in a flight stage +5%" in my opinion, as it wasn't intended for re-use or re-ignition, so the performance across 3 burns on the ground isn't a good basis.

From my quick research, that would come to about 120 seconds for this engine, you may know a more precise value from your research.

365 seconds is crazy long. Like, a Delta sustainer doesn't exceed 300 seconds!

PR #1395, opened 2016-09-22
Add Fastrac config to Merlins and update description

Ravenchant:
It still defaults to 1A though

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Merlin1_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Merlin1_Config.cfg
@@ -14,7 +14,8 @@
 {
 	@title = Merlin Series
 	%manufacturer = SpaceX
-	@description = The whole family of Merlin rocket engines, made by SpaceX. Diameter: [1.75 m].
+	@description = The whole family of Merlin rocket engines, made by SpaceX. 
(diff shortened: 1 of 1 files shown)


SirKeplan:
that sea level isp is more than just suspicious, it has to be totally wrong. no way a fairly low performance engine like that has got this performance, what's your source for this?

Another thing, unless this is quite similar to the Merlin, this should be a separate  engine.

SirKeplan:
ok, now this is not perfect, but gives an idea of the isp possible. the sea level isp of this with a 15:1 nozzle is around 263 s

http://www.lpre.de/upload/AIAA-1998-3365-Fastrac.pdf

Ravenchant:
Ah, yes. I figured something was fishy, but...well, doing research right before dozing off was a bad idea. The source for specific impulse was:
http://www.b14643.de/Spacerockets/Specials/U.S._Rocket_engines/engines.htm
since Astronautix' article is lacking at best. It gives a different mass flow rate than the wiki article, so I recalculated the thrust according to that. There was also some NASA testing report that I can't find again which gave rough thrust figures and the expansion ratio, namely 30:1.

What I *think* happened is that the 310s Isp was calculated at the altitude X-34 was supposed to be dropped from. Whoops.

PR #1396, opened 2016-09-23
RN rcs fixes

raidernick:
@Nathankell these are the fixes I have come up with for a few parts.

All of these parts needs RCS enabled by default because they fit a
certain criteria:
1) They have a combined solid stage/decoupler and rcs module
2) They are the first(or only) part on the rocket stack that needs rcs enabled,
so enabling it by default makes sense
3) They need RCS for ullage of a liquid stage and the rcs and liquid
engine are combined on one part

every one of these parts fits the 2nd criteria so having rcs enabled by
default should be perfectly acceptable, but that alone was not enough to accept it as part of the compromise.

I omitted a bunch of parts as I decided they need to fit at least 2 of
those options to be modified to be on by default.


stratochief66:
What is with the removal of the 3 ModuleRCS's ? The PR description doesn't seem to address those.

raidernick:
@stratochief66 they are leftovers from when the rcs pods were attached to the main engine mesh. They have been separate parts for years and I just forgot to remove them from the original mesh.

NathanKell:
Looks good to me too :)

PR #1398, opened 2016-09-23
RN fix all part volumes

raidernick:
-For some reason there were a ton of total part volumes calculated
incorrectly for almost all of my parts. This will fix that.

-Craft files will need to be redone AGAIN...FML


raidernick:
All max volumes now conform to the added value of all contained resources.
The errors were leftovers from whoever originally did the configs for some of my oldest parts. 

Since then, many changes have been made to make fuel values and tacls values more correct. For some reason I forgot to also update the total volume.

PR #1399, opened 2016-09-24
Rn solar and battery fixes

raidernick:
-Fix battery capacity and charge rates for panels based on the best info
I could find for older probes


raidernick:
craft files will need to be redone AGAIN for these changes

raidernick:
-these fixes have been tested and work in game. All craft files also work correctly.

-Data was taken from nasa.gov first and if not available anywhere I could find it. If no data was available at all(which was the case often), I made a best guess based on the mass of the probe and the probable weight of the battery and types used based on the time period. 

Knowing the battery type and probable mass of the batteries I could figure out the kJ using the average watt hours per pound for that battery type, which I know thanks to NASA documentation on transit IB a probe made at around the same time as these other early probes.

PR #1400, opened 2016-09-24
Update RO_KOSMOS_Engines.cfg

Moonatick:
I was getting errors in Module Manager with KOSMOS installed so I tried my hand at editing the MM patch. This seems to fix the errors on my test install. Please review and confirm that everything is correct and working as intended.


raidernick:
@Moonatick are you using the latest version of KOSMOS? I added some new engines to the station pack in the latest release. Where are you downloading it from

Moonatick:
I thought I had downloaded from your github page, but just to be sure I downloaded it again from your link with the original cfg, and it loaded with six errors related to the aforementioned cfg. The engine specs seem to be working, but the title for some reason displays the "non ro" tag on all engines.

SirKeplan:
what is with the closes, i assume this PR is no longer relevant?

PR #1402, opened 2016-09-28
Reduce the minimum diameter of the proceduralCrewTube part

tschieggm:
The procedural crew tube part is very useful for creating adapters between docking ports and stations. Unfortunately most docking ports are much smaller than 2m. Reduce the minimum diameter to 800mm.

[Pressurized Mating Adapters (PMAs) specifications](https://www.nasa.gov/externalflash/ISSRG/pdfs/pma.pdf)
[APAS Development](https://en.wikipedia.org/wiki/Androgynous_Peripheral_Attach_System#Development)
[Soyuz docking probe specifications](http://www.russianspaceweb.com/docking.html)


SirKeplan:
Was originally added in #500 , seems like this lower minimum diameter makes more sense. though

PR #1408, opened 2016-10-24
RD-701 & RD-704 engines

PhineasFreak:
Adds two new engines to the Soviet Engine Pack: the tripropellant dual - mode RD-701 and it's single chamber version, the RD-704. Global engine configs are available for integration with other mods.

Notes:

* The RD-701 is a copy of the RD-180 (same format).
* The RD-704 is a copy of the RD-191 (same format).
* The inert mass values might not be 100% accurate.
* Mode toggling is done via the part module interface and not via RF.
* The "Automatic" mode switch option does not seem to work.
* RF cannot handle the automatic tank fill for both modes simultaneously (but works fine for either of them separately).
* The engines were only ground - tested in reality (never flown) so TestFlight configs are not available.


SirKeplan:
Great! i'd thought about adding these engines, but never got round to it. haven't tested this ingame yet, but what i have noticed is the max thrust for RD-701 on LH2 mode looks wrong, i think max for LH2 is 1588 kN (357,000 lbf)

also for dry mass, give figures in a NTRS document higher priority to astronautix

PhineasFreak:
I rounded a bit some values (since the different sources gave different results) but i will fix them. Going by the NTRS document the thrust values should be:

* Mode#1 thrust: 900000 lbf = 4003.39 kN (rounded to 4000 kN?)
* Mode#2 thrust: 357000 lbf = 1588.01 kN (rounded to 1588 kN?)

The inert mass value should be:

* RD-701: 9742 lb = 4418.89 kg (rounded to 4420 Kg?)
* RD-704: Half of the RD-701 = 2210 Kg (other sources indicate that is lower than that, at 1923 Kg, due to the different size & construction)

The NPO Molniya also mentions a throttle range between 32% and 100% for Mode#1 but seems far too low for a hydrocarbon cycle (compared to say with RD-180 with a range of 47% to 100%). Astronautix has a somewhat better min throttle value of 40%.

(I also seem to have set the throttle range wrong as i have added it to the Mode#2 instead of the Mode#1)

PhineasFreak:
Hopefully the latest corrections are good! The inert mass of RD-704 is probably off so we should probably take a second look it it.

SirKeplan:
Works nicely as is, second look at some stats might be good, but merging for now.

PhineasFreak:
OK, i will do some more research and try to find some better sources!

PR #1410, opened 2016-10-25
DECQ Soyuz DM mass fix

PhineasFreak:
Because 3 crew really weight 300 Kg...


stratochief66:
If this is for the case where the user has USI installed, shouldn't the MM logic be looking for a USI installation?

PhineasFreak:
USI is not considered at all here. This patch did two thing:

* If USI is installed then the USI conversion module is going to be outright removed (since USI is not supported at all by RO), The USI removal patch is there because a) the OLDD Soyuz MM patches are poorly written and do not check for dependencies before adding any mod modules and b) the OLDD Soyuz folder name comes after RO so any MM patches will run after RO. So, i am forced to remove it in a semi-final patch.

* The mass fix is because i forgot something important: RO adds mass to the crew members (100 Kg per crew member). I thought it was a bug with MFT/RealFuels and so i "fixed" a bug that was not there (and thus the small comment that i had left).

In short, USI has nothing to do with that bugfix. It is there because of the various reasons listed above.

PR #1416, opened 2016-11-07
RCS TestFlight configs

PhineasFreak:
(Re)Uses the TF configs of the stock 1 kN Generic Thruster (operates like an RCS thruster) so operational burn times might need some tweaking for the actual RCS thrusters.

All RCS & engines that use the global RCS config are automagically supported.

Also supported is the new MMH/MON3 propellant combo.

[Alternate diff ignoring whitespace changes.](https://github.com/KSP-RO/RealismOverhaul/pull/1416/files?w=1)


PhineasFreak:
Small update to fix some situations where the "useRcsCostMult" and the "useRcsConfig" would not pass the cleanup checks.

PR #1430, opened 2016-11-30
Titan Engine updates

ctiberious:
For some reason we currently have the LR87-AJ-7 and LR91-AJ-7 engines used on the Titan GLV as being weaker than the LR87-NA-5 and LR91-NA-5 engines used on the Titan II ICBM and Titan 23G (Titan II SLV).  I'm not sure where the numbers we currently use came from.  The first reference address, while full of great information, doesn't seem to have any thrust or ISP information for the Titan.  And Astronautix no longer seems to have usable data.  I was able to find a NASA Technical Document from the Gemini 7 mission which does list the thrust for both engines and those values appear to line up with what can be found on b14643.de so I'm suggesting we update.

Also, I've found in game that I can't seem to get the FASA Gemini into orbit using the values as we have them currently.  I'm always around 200-300dV shy.  But by using the numbers from b14643.de, I'm able to get into orbit with about 100-200dV to spare.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/LR87_X2_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/LR87_X2_Config.cfg
@@ -3,6 +3,8 @@
 // References:
 // [1] http://crgis.ndc.nasa.gov/crgis/images/c/cd/A_Historical_Look_at_US_Launch_Vehicles_1967-Present_%281988%29.pdf
 // [2] Astronautix
+// [3] https://ia802608.us.archive.org/29/items/gemini7nasamissi00godw/gemini7nasamissi00godw.pdf
+// [4] http://www.b14643.de/Spacerockets/Specials/U.S._Rocket_engines/engines.htm
 @PART[*]:HAS[#engineType[2XLR87]]:FOR[RealismOverhaulEngines]
 {
 	%title = LR87 Booster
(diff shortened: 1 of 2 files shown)


SirKeplan:
this is often a good site for US rocket engines, http://www.alternatewars.com/BBOW/Space_Engines/Aerojet_Engines.htm

ISP values seem to differ from b14643, what do you think?

ctiberious:
The more I research, the more I wonder if there was any functional difference between the AJ-5 and AJ-7 engines other than one was man rated and the other wasn't.  Maybe we should simply combine the two engine configs into a single config: LR87-AJ-5/7 and LR91-AJ-5/7?

I like that alternatewars has a link with the NASA tech doc they used to get their numbers (http://www.alternatewars.com/BBOW/Space_Engines/SP-8120_Excerpt.pdf).  Though I'm not sure why they list the thrust for the LR87 slightly higher than the tech doc lists (214400*2 vs 430000 sl) (236400*2 vs 473800 vac/alt).  Either way, what about if we used:
LR87-AJ-5/7
1907.4 @ 263 s.l.
2103.1 @ 289 vac/alt
LR91-AJ-5/7
444.8 @ 321 vac/alt

SirKeplan:
Combining would make sense, except we already have TF configs for the different variants, and presumably they had different reliabilities in real life too.

I think for right now just keep them separate and give them accurate (and nearly identical) stats.

raidernick:
Is this PR ready to merge?

PR #1433, opened 2016-12-02
Vernier Gimbal edited on LR_101_Config

alimoncul:
I think verniers only need 1 axis but this is 1.5 axis i think. Check image i can still disable XP Axis. https://gyazo.com/8de6f6febf75ff3c36ce8a4f20f4fb91


alimoncul:
Mechjeb Stable Gif : https://gyazo.com/7c1f90d6414487014a4554fe12e8c455

stratochief66:
I ~think~ that they should be able to pitch in somewhat, I recall NK and I discussing that a bit and me watching a bunch of videos when this cool feature was finally added to KSP.

I'll leave it as it is here in your PR for now. If anybody complains or it seems odd, I reserve the right to take a second look and possibly make them able to pitch in again. :)

PR #1438, opened 2016-12-04
Changing old input-output with new input-output on TacAirFilter Part

alimoncul:
First test of new sytstem.


PhineasFreak:
Please note that the original config ("AirFilter.cfg") already includes the INPUT and OUTPUT nodes. Currently you are duplicating them instead of simply patching them.

alimoncul:
I think i am not enough good to fix it i will close this and #1439 pr's i didnt understand issue :D sorry guys.

PR #1439, opened 2016-12-04
Changing old input-output with new IO on TACLS CO2 Scrubber (Lithium Peroxide)

alimoncul:
… (Lithium Peroxide)

Lithium Peroxide CO2 Scrubber Module updated.


PhineasFreak:
Check the note that i left in PR #1438.

PR #1440, opened 2016-12-04
RO RemoteTech v1.8 updates

PhineasFreak:
Updates to the RO RemoteTech compatibility patch.

Change log:

* Convert the RemoteTech settings file into a Module Manager patch to be compatible with the new RT loading system (as of RT version 1.8.2).
* Remove the "HideGroundStationsBehindBody" parameter (now part of RT itself).

### DIFF
--- a/GameData/RealismOverhaul/RemoteTech_Settings.cfg
+++ b/GameData/RealismOverhaul/RemoteTech_Settings.cfg
@@ -1,9 +1,14 @@
-RemoteTechSettings:NEEDS[RealismOverhaul]
+//  ==================================================
+//  Custom RemoteTech settings for Realism Overhaul.
+
+//  *   Make "Root" the default communication rule set.
+//  *   Allow combining multiple antennae for range extension.
+//  *   Zero the active engine throttle on connection loss.
+//  ==================================================
+
+@RemoteTechSettings:FOR[RealismOverhaul]
 {
-	RangeModelType = Root
-	MultipleAntennaMultiplier = 1
-	ThrottleZeroOnNoConnection = False
-  
-  // Should be removed after RemoteTech v1.7.0 has released 
-	HideGroundStationsBehindBody = True
+	%RangeModelType = Root
+	%MultipleAntennaMultiplier = 1.0
+	%ThrottleZeroOnNoConnection = False
 }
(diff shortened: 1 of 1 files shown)


stratochief66:
@SirKeplan or @NathanKell 

I know nothing about RT, either of you want to take & test this?

rsparkyc:
I might be able to look at this as well

PhineasFreak:
To give some background for the changes: https://remotetechnologiesgroup.github.io/RemoteTech/modders/settings

In short: RT changed the way it loads the configs. Before it would scan the game database and apply the patches in an alphabetical order. Some mods that patched other mods had problems because of that and so the loading system was changed to use MM instead. It works exactly like before but it now uses the MM conventions.

I did something similar for the RSS RT patch: https://github.com/KSP-RO/RealSolarSystem/commit/513cc937a41b475cc08fb10db4be04c468b85334

GameData/RealismOverhaul/RemoteTech_Settings.cfg
```diff
@@ -1,9 +1,14 @@
-RemoteTechSettings:NEEDS[RealismOverhaul]
+//  ==================================================
+//  Custom RemoteTech settings for Realism Overhaul.
+
+//  *   Make "Root" the default communication rule set.
+//  *   Allow combining multiple antennae for range extension.
+//  *   Zero the active engine throttle on connection loss.
```
rsparkyc: Looks like we have this set to `False`, but the comment implies we have this set to `True`.  I would think the more realistic option is leaving this to `False`, but we should probably update this to say "Don't zero the active engine throttle on connection loss."

> ThrottleZeroOnNoConnection (default = True)
If true, the flight computer cuts the thrust if you have no connection to mission control.

PhineasFreak: Theoretically once you lose control of a craft you cannot do anything. So, if your engines are firing and you lose control then the throttle should remain as-is. I done goofed and wrote a bad comment there...

rsparkyc: Yep, my thoughts exactly.

PhineasFreak: Took the chance to also change the MM patching modifiers since these fields are always going to be present.

PR #1450, opened 2016-12-07
SSTU MUS tanks

ctiberious:
I've removed the Hydrazine resource since it's not working in game anyway.
I've adjusted the tech requirement for the built in 300km passive antenna so you have it right from the start.


Theysen:
Care to elaborate? The RCS is set to use Hydrazine iirc and is thus drawn from the tank accordingly.

ctiberious:
Theysen, the RESOURCE I removed was separate from the Real Fuels tank.  It was originally intended as a separate, fixed capacity tank that would allow you to use the default Hydrazine RCS with a Balloon or Cryo-Balloon tank (simulating later versions of Centaur).  Unfortunately, even though the Hydrazine RESOURCE was part of the config (you could see it if you looked through ALT+F12), it never displayed anywhere and as far as I could tell, was never used.

PR #1454, opened 2016-12-15
Sensible config prices for the H-1 family

ec429:
Based on a part cost of 750, making the costs be:
H-1     750
H-1B   1050
RS-27   800
RS-27A  825
I have no source for these prices, but they're hopefully better balanced
than the old ones.


ec429:
Since the other half of this ([RP-0#603](https://github.com/KSP-RO/RP-0/pull/603)) has been merged, any reason why this can't be too?

PhineasFreak:
Pinging @NathanKell, @SirKeplan and, @stratochief66.

raidernick:
What is going on with this config? I don't see any issue why it can't be merged. If someone else doesn't chime in on it I'm going to assume it's ok to merge then.

PR #1461, opened 2016-12-22
Kestrel-1B Global Config

PhineasFreak:
* Add a global engine configuration for the hypothetical Kestrel-1B engine (a Methalox variant of the kerolox Kestrel engine).

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Kestrel1B_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Kestrel1B_Config.cfg
@@ -76,8 +76,8 @@
 
             atmosphereCurve
             {
-                key = 0 300
-                key = 1 120
+                key = 0 331
+                key = 1 160
             }
         }
     }
--- a/GameData/RealismOverhaul/Engine_Configs/Kestrel1B_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Kestrel1B_Config.cfg
@@ -1,0 +1,86 @@
+//  ==================================================
+//  Kestrel-1B global engine configuration.
+
+//  Inert Mass: 60 Kg
+//  Throttle Range: 75% to 100%
+//  Burn Time: 418 s
+//  O/F Ratio: 2.77
+
+//  Space Launch Report - Falcon LV: http://www.spacelaunchreport.com/falcon.html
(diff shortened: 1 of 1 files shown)


Ravenchant:
Assuming all other variables kept equal, why would a switch to methalox propellant significantly reduce performance?

SirKeplan:
Assuming all things being equal methalox should give 10-15s more ISP than kerolox. For a hypothetical pressure fed kestrel on methalox, probably about 10s more than the kero one i would think.

http://yarchive.net/space/rocket/fuels/fuel_table.html

PhineasFreak:
I will be completely honest: i have not done any theoretical modelling of the engine since i do not know how to do that. I was hoping that more experienced members would help to fix/improve that (and it seems like it!).

SirKeplan:
Fair enough. If i had some more stats on this engine(oxidiser fuel ratio, nozzle expansion ratio) etc, i could model a good approximation of the stats in [RPA](http://www.propulsion-analysis.com/index.htm). However as this methalox variant is completely hypothetical i believe, then about 330s ISP is fine.

PhineasFreak:
I have the O/F ratio defined in the opening engine notes but not the expansion ratio. I think that it can probably be calculated from the nozzle diameter? The diameter is approximately 0.5 m (50 cm).

That tool seems to be what i need to calculate the actual performance. Thank you for introducing it to me!

stratochief66:
@PhineasFreak that second set (in Edit 3) of numbers appears reasonable compared to SirKeplan's suggestion of 330s vac ISP and my general understanding of Methalox. It can always be adjusted when more information comes in.

Do you want to change this PR to reflect those numbers, then I can merge this?

PhineasFreak:
Yep, no problem. Give me a couple of minutes.

SirKeplan:
These stats in the config look fine to me, though i would note that an O/F Ratio of 2.77 is unusually low. Methalox tends to have a ratio of around 3 - 3.5.

PhineasFreak:
@SirKeplan maybe change it to an O/F ratio similar to the Raptor (3.8)? It is the closest that we have now in terms of methalox flight hardware.

SirKeplan:
Probably nearer 3.0, they are at opposite ends of the spectrum for engine performance.

PhineasFreak:
OK, i will push the changes with the next CMES patch updates.

PR #1464, opened 2016-12-22
CMES Propulsion updates (part I)

PhineasFreak:
Updates and fixes for the CMES "Propulsion" parts. First pass.

Changelog:

* Update all engiine parts to use the new global engine configuration system.
* Update all thermal animation modules to use the new FXModuleAnimateThrottle module.
* Fix the volume of the Delta IV CBC (was lower than it should).
* Fix the size of the 4 and 5 Segment RSRM.
* Add proper web sources.
* Remove invalid part modules from many parts.
* Explicitly set the part categories (and use the new KSP 1.2 ones).
* Balance inert and gross mass values against real sources and/or other parts.
* Adjust some insane max temperature values.
* Update the part titles and the descriptions.
* Fix formatting (tabs, false spacing).

**Note 1: some new parts are missing from these PRs to try and keep things cleaner. These will be included in a separate PR in the future.**

[Alternate diff ignoring whitespace changes](https://github.com/KSP-RO/RealismOverhaul/pull/1464/files?w=1)


stratochief66:
In the Diff version ignoring whitespace changes, it looks like you remove the thrust curve, but I don't see a matching tag to use the global engine config that I would expect. Is that done in a different PR, is the whitespace change-less Diff misleading me, or did I miss something else?

example:
https://github.com/KSP-RO/RealismOverhaul/pull/1464/files?w=1#diff-88d32b025fbde54d94dc7ed4b8c3a65eR633

PhineasFreak:
I have added the necessary code to reference the RSRM engine config in the lines 575-578:

https://github.com/KSP-RO/RealismOverhaul/pull/1464/files?w=1#diff-88d32b025fbde54d94dc7ed4b8c3a65eR575

stratochief66:
Excellent, I had just overlooked it then :)

PhineasFreak:
To be fair, the web git diff is difficult to traverse, especially if a lot of content has been added or removed. Hopefully, the "PR-per-part" will help this time!

PR #1465, opened 2016-12-22
CMES Propulsion updates (part II)

PhineasFreak:
Updates and fixes for the CMES "Propulsion" parts. Second pass.

Changelog:

* Update all engiine parts to use the new global engine configuration system.
* Remove invalid part modules from many parts.
* Remove the small separation motor and replace it with the (old) medium one.
* Separate the RL10 configs so that only the B/C and CECE versions are available (due to it's bell structure).
* Explicitly set the part categories (and use the new KSP 1.2 ones).
* Balance inert and gross mass values against real sources and/or other parts.
* Adjust some insane max temperature values.
* Update the part titles and the descriptions.
* Fix formatting (tabs, false spacing).

[Alternate diff ignoring whitespace changes](https://github.com/KSP-RO/RealismOverhaul/pull/1465/files?w=1)


PhineasFreak:
Closing since it did not work as expected.

PR #1379, opened 2016-09-16
Update to KSP 1.2.

ahmedcharles:

PR #1469, opened 2017-01-04
change to recompile against 1.2 dependencies

rsparkyc:
Hopefully I did this right.  Attempted to recompile and saw that this line was breaking against 1.2


PhineasFreak:
Sorry to disappoint you but the PR #1379 already covers most of these. The EGC recompile is a nice addition though.

rsparkyc:
Ahh, and from @NathanKell's comment, it looks like a lot of this code can be deleted.

PhineasFreak:
Keep this one open though because @ahmedcharles has not been active lately. Also, it does not include any of the EGC stuff.

rsparkyc:
Ok, I updated it based on @NathanKell's comments too.  Looks like there's no more KSPUtil.dll, and it compiles fine without the reference, so I killed that

PhineasFreak:
Correct, the KSPUtil is now part of the Assembly-CSharp.

PhineasFreak:
Something else that i noticed is that you did not remove the KSPUtil references from EGC. Not really necessary but it would be nice to do so.

rsparkyc:
Thanks @PhineasFreak, I've updated it, and also pulled from master, so this should be up to date.

stratochief66:
@NathanKell @SirKeplan if you can verify these changes are good, could you please merge?

SirKeplan:
There seems to be a problem with crew mass in 1.2, adding or removing kerbals from a part is no longer having any effect on mass. Parts just keep their base mass no matter how many crew are onboard.

Same happens without removing EditorCrewMass as in #1379 

@NathanKell

rsparkyc:
Help us @NathanKell, you're our only hope...

SirKeplan:
OK, seems the issue was a MM glitch on my end (sigh) 

@rsparkyc the issue you see should be because we need to remove the inclusion of EditorCrewMass in `RealismOverhaul_Global_Config.cfg`

rsparkyc:
@SirKeplan Yeah, that made sense the more I thought about it.  Glad that the kerbal mass issue seems resolved.  I'll go ahead and make those changes for EditorCrewMass

rsparkyc:
There's some ships with a module named EditorCrewMass, should that module be deleted too?

SirKeplan:
@rsparkyc ahh i see, some craft files have an EditorCrewMass module. I am not sure here, @raidernick these look like your crafts, can we simply remove the module form the .craft files?

raidernick:
@SirKeplan I don't have the time to read this huge thread, can someone quickly explain wtf that module is why it matters if it's removed or not

rsparkyc:
@raidernick give a short answer in your other PR.  For now just leave them in, I think the answer will become self evident once things start getting merged into master.
On that note, what's the process for getting stuff merged in? Are we waiting for @NathanKell ?

PhineasFreak:
@raidernick it does not matter as the KSP modules are added/removed dynamically on craft load/save. Opening the craft file using the new RO assembly (without the editor mass fix) and re-saving them is probably enought to take care of that.

Invalid modules spam the log with harmless errors but it is spamming it non the less.

PR #1471, opened 2017-01-08
FASA Tacls Update For 1.2

stratochief66:
Update FASA Configs for 1.2 to be compatible with TACLS Version 0.12.8 for KSP 1.2.2

The syntax requirements of TACLS changed, these changes to FASA RO configs are to adhere to that syntax.


stratochief66:
@JPLRepo If you have a chance, could you skim this for any errors that would cause something to break or not function as expected?

I have some formatting inconsistencies in there, but I don't see anything that would cause unexpected behaviour.

JPLRepo:
@stratochief66 Looks goot to me.

PR #1481, opened 2017-01-11
CMES final pass

PhineasFreak:
Generic edits to some parts that were missed during the main passes.

Change log;

* Improve the wording of some part descriptions.
* Improve the descriptions of some web source links.
* Add some trailing zeros to be consistent with other part configs.
* Remove some leftover custom sound definitions from the AJ10-190 engine.
* Adjust the O/F ratio of the Orion MPCV RCS blocks to be the same as the main engine. Increase the VAC Isp value to be closer to other hardware.
* Remove an unneeded AeroHider module from the procedural SLS fairing.

And with that PR, Chaka Monkey Exploration Systems can be considered (probably for the first time) 100% compatible with RO!


stratochief66:
Thanks for all the hard work on CMES (and everything else) @PhineasFreak . I remember patching a few parts in this a year or so ago to make an SLS, thinking "no way I could create RO configs for all these things.

PhineasFreak:
When i started with CMES a year ago i though "Oh, a nice mod to get me started with RO dev". I was new back then and it took me a while to realize that "Boy, this mod pack is huge! How am i ever going to get them ready for RO?"

Now, 14 months after that, i sat down and did a quick recap: 7 (almost) game-breaking KSP updates, at least 2000 hours of part R&D and a hundred or so of PRs to make them compatible with RO.

Damn, i hate my life ;)

PR #1483, opened 2017-01-11
Smoke screen shuriken

rsparkyc:
2 commits here, apparently a bunch of files were missing ending newlines, (which sed wanted to create), so the first commit adds those.  The second commit dose the actual replacement.  I'm _assuming_ that I can just a global search/replace, but I'm not in a position where I can really test this yet.


stratochief66:
Looks good! Will need some in game testing, naturally.

PR #1487, opened 2017-01-12
Fixes to RD-858/859

Ravenchant:
Both the main and reserve engine were pump-, not pressure-fed. Tidied up some Isp and thrust numbers and reduced number of ignitions on RD-858 from 20 to 12, as indicated on the manufacturer's page.

I let %ullage = False for now even though I see no reason the LK wouldn't be subject to the usual requirements. Also, the actual thrust settings would perhaps be better replicated by adding a ModuleHybridEngine with full thrust and throttled modes, but I need to look up how that works first.

Sources:
http://www.yuzhnoye.com/en/company/history/liquid-propellant-rocket-engines.html
http://www.astronautix.com/r/rd-858.html


SirKeplan:
Looks good, but wouldn't the actual minimum thrust be `5.469` as the minimum thrust is throttleable by +/- 35% if I understand it right.

Also, I tend to agree  that a ModuleHybridEngine would be the best way to set this up, if you get round to it, that would alos allow the throttled mode to be given the lower ISP.

Ravenchant:
That's correct. Whoops.

raidernick:
@Ravenchant The changes look good so far. The asl isp values are irrelevant as the engines were never intended to be used in any sort of atmospheric environment, they can literally be any value within reason.

Also for future changes to my partsthat anyone makes I ask that you please tag me in the PR so I can keep track of the changes as I usually have my own repo with a ton of my own changes not committed here yet that could conflict with this.

Also, if you ever change a value that will affect any of the RO craft files that those parts are used in, you MUST update the craft files for the changes also, don't leave that mess for me to discover at a much later time that I had no idea about. I'm not saying you did this, and none of the changes so far have affected a craft file, I'm just making a note for any future change to anyone who sees this.

Ravenchant:
Will do, @raidernick (testing, testing).

stratochief66:
@raidernick good to merge? Or do you want to test that the changes don't make one of your crafts unworkable?

raidernick:
@stratochief66 don't need to test it, just minor changes to thrust and ISP. I checked to make sure there aren't any syntax errors, it looks fine.

PR #1488, opened 2017-01-14
RS-68 gimbals in global config, fix for CMES

SirKeplan:
Put the main gimbal in the global config, seems to work ok. RS-68s with the gas gen modelled need to set their specific verniers.

@PhineasFreak separate issue, but i noticed the CMES RS-68 doesn't attach inside the engine shroud, is there a missing attach node?

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RS68_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RS68_Config.cfg
@@ -33,12 +33,12 @@
         %EngineType = LiquidFuel
     }
 
-    //@MODULE[ModuleGimbal]
-    //{
-    //    @gimbalRange = 6.0
-    //    %useGimbalResponseSpeed = True
-    //    %gimbalResponseSpeed = 16
-    //}
+    @MODULE[ModuleGimbal]
+    {
+        @gimbalRange = 6.0
+        %useGimbalResponseSpeed = True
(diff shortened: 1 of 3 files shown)


PhineasFreak:
@SirKeplan not exactly. Since the CMES RS-68 engine includes the complete chamber structure the node is placed at the top of the structure (the small thrust plate) to make it compatible with other parts (like the PF interstages and such).

For the CMES Delta IV you have to use the offset tool and slide the engine inside the engine mount (so that the engine mount can have it's nodes to attach other engines if you want).

I assume that i could add a third internal node to be "RS-68 exclusive" i guess.

SirKeplan:
@PhineasFreak tabs to spaces done, we really need to enforce some style guide on this..... :/

Requiring an offset should be OK, though maybe that should be mentioned in the part description for the mount.

PhineasFreak:
I will probably add an extra attachment node to the engine mount. Cleaner and easier to use.

For the formatting, i think we are doing good. As long as the file is consistent (tabs with tabs, spaces with spaces) i do not think that we are going to have any problems (and boy, i am guilty for a lot of them!)

stratochief66:
'Enforcing style' isn't really isn't something that can be done with a volunteer project, in my opinion. You could make sure we have our preferred style documented somewhere prominent, and ensure new contributors see it, and politely remind offenders.

PR #1490, opened 2017-01-16
BNTR global engine config updates

PhineasFreak:
Change log:

* Improve the part description.
* Re-evaluate the thrust and mass values for each operating mode.
* Replace the ModuleEngineConfigs with ModuleHybridEngine to make it
possible to switch between the bimodal and trimodal modes in flight.
* Increase the amount of ignitions (no reason to keep it low).
* Add spool-up time to the trimodal mode.
* Change the requested Uranium ratio to be 1/lambda (lambda = 703800000
years).
* Add a generator module (Brayton converter).
* Add the required resources for operation to the global engine config itself instead of relying on the standard RO part config. This also fixes the missing resources from the BahamutoD Constellation Essentials BNTR engine.
* Disable the above nuclear fuel resources from being tweakable.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/BNTR_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/BNTR_Config.cfg
@@ -1,10 +1,10 @@
 //  ==================================================
-//  Bimodal NTR global engine configuration.
+//  "Bimodal" NTR global engine configuration.
 
-//  Inert Mass: 2170 Kg (base configuration)
+//  Inert Mass: 2270 Kg
 //  Throttle Range: N/A
 //  Burn Time: 5400 s
-//  O/F Ratio: 0 (SNRE), 3.0 (TRITON)
+//  O/F Ratio: 0 (BNTR), 3.0 (TRITON)
 
 //  Sources:
 
@@ -16,6 +16,7 @@
 
 //      BahamutoD Constellation Essentials
 //      CMES
+//      SXT
 //  ==================================================
 
 @PART[*]:HAS[#engineType[BNTR]]:FOR[RealismOverhaulEngines]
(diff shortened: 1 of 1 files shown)


SirKeplan:
Just noticed we have 2 very similar BNTR engine configs, the TRITON variant in each is essentially the same. On the non TRITON the main difference is thrust level.

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/BNTR_Config.cfg

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/BiModal_NTR_Config.cfg

thinking we should probably merge these in some way.

PhineasFreak:
Oh, i just noticed it! Let me see the commonalities and differences between the two.

SirKeplan:
- 25t mass is ridiculous, must be paste error.

- the 111.2 kN (25 klbf) engine is only mentioned in this source. https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20140009579.pdf, so must be specific for the nuclear thermal electric Mars vehicle. the other ones assume roughly 15 klbf (66.72 kN)

- spool time is nice, can't find any info on throttle, though it's low enough thrust that throttling down probably isn't that useful. should spool time be lower as it's a small engine?

- yep that's a mass-volume mistake, BTNR_Config has it correct.

_____________
 The other thing, the point of a LANTR mode is it makes the engine throttleable in-flight by allowing you to "switch gears". however we have the LANTR mode as a RF engine configuration which can't be switched outside the VAB. this seems worth doing and it's a simple issue of switching `ModuleEngineConfigs` to a `ModuleHybridEngine`

PhineasFreak:
Alright, i am making the necessary changes, including the mode switching.

PhineasFreak:
OK, full update is live!

Issue #1494, opened 2017-01-18
Structural failure for large interstage fairings on craft load.


SirKeplan:
This on 1.2

Thinking this might be a KJR problem, when i load crafts on the launchpad with large interstage fairings, there is a sudden bump as it loads, then 1 or more fairings just fall of and generally explode as the hit something.

![ksp-bug](https://cloud.githubusercontent.com/assets/1227975/22087967/f82a7672-ddd8-11e6-8bc6-16e370e97f7a.png)

Test craft.
[fairtest.craft](https://github.com/KSP-RO/RealismOverhaul/files/715339/fairtest.craft.txt)

[relevant log section](https://github.com/KSP-RO/RealismOverhaul/files/715333/fairingtestLog.txt)

PhineasFreak:
I have a suspicion that it is caused (again) by the PF payload autostruts, but this time from having too much rigitity (similar to the "too-many-launch-clamps" syndrome). If you temporary disable the autostruts do the fairings still fail?

SirKeplan:
@PhineasFreak right, disabling autostruts fixes it. thanks.

ahah, browsing the [PF forum](http://forum.kerbalspaceprogram.com/index.php?/topic/36371-12-procedural-fairings-320-november-8/) i now see some similar reports.

PhineasFreak:
You are lucky! I was testing an Ares I & Ares V (with large interstages) and both of them kept loosing the interstage and/or payload fairings. After i saw your report i though "could it be what i imagine" and disabled the option. Presto! No more failures!

OK, time to re-introduce the PF patch that i removed some time ago...

PR #1496, opened 2017-01-19
Generic PF updates & fixes

PhineasFreak:
Various updates and fixes for the Procedural Fairings RO compatibility.

Change log:

* Merge all possible global PF modifiers to a single pass.
* Add a "generic" manufacturer.
* Add back a patch to remove the PF auto struts (see https://github.com/KSP-RO/RealismOverhaul/issues/1494 for more information).
* Add some white space for better readability.

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
@@ -1,110 +1,150 @@
+//  ==================================================
+//  Procedural Fairings fairing sides.
+//  ==================================================
+
 @PART[KzProcFairingSide1]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	%title = Payload Fairing - Egg-Shaped [Procedural]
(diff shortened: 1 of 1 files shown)


SirKeplan:
Hmmm , something odd going on. I can't find the `squad_interstage_adapter` adaptor part in-game at all, what am i missing?

I have PFFE and CLS installed.

PhineasFreak:
@SirKeplan i may have messed up something. The specific part is a copy of the "KzInterstageAdapter2" from PF, as defined by PFFE. So, if this part is not compiled correctly then neither that part will be.

Can you confirm that the part "Interstage - w/ Decoupler [Procedural]" is available in the part list?

SirKeplan:
yep, the `KzInterstageAdapter2` part is there in game. though all my proc bases have `IB - ` prefixed.

PhineasFreak:
The prefix is normal, it is added by PFFE.

Another test, is the KzInterstageAdapter2 part included in the MM cache file?

SirKeplan:
Yep, both `KzInterstageAdapter2` and `squad_interstage_adapter` actually are in the mm cache file. i'm thinking this is an unrelated issue to this PR.

PR #1500, opened 2017-01-21
RD-701 & RD-704 updates

PhineasFreak:
Updates for the BobCat Soviet Engines RD-701 and RD-704.

Change log:

* Implement ModuleHybridEngines instead of the stock MultiModeEngine for better compatibility with RF.
* Change the patching of the editor parameters (title, manufacturer, description) to use the "%" MM modifier instead of the "@" (covers parts that may be missing these fields entirely).
* Disable the TEATEB ignitor resource from being tweakable by the user (no reason to change the carried amount).

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD701.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD701.cfg
@@ -36,17 +36,19 @@
 
     !MODULE[ModuleEngineConfigs],*{}
 
+    !MODULE[ModuleHybridEngine],*{}
+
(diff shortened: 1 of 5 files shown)


SirKeplan:
Cool! i think you need to update the plume file to patch a `ModuleHybridEngine` now though.

PhineasFreak:
Oh right, give me a sec. I also have to do the same for the BNTR.

Thanks for noticing it!

PhineasFreak:
OK, plumes are good now!

PR #1502, opened 2017-01-21
SXT engines updates & fixes

PhineasFreak:
Various updates for the SXT engines.

Change log:

* Remove the TweakScale patches (already taken care by a global config).
* Remove two old part configs that were never converted for RO.
* Update the BNTR to use the new global engine config (#1490) and remove the old global engine config.
* Minor formatting fixes (convert spaces to tabs, remove trailing spaces).

[Alternate diff ignoring white space changes](https://github.com/KSP-RO/RealismOverhaul/pull/1502/files?w=1)


SirKeplan:
Unless i've missed it in another PR, i think the SXT BNTR plume needs fixing :)

PhineasFreak:
***...***

I really need someone to pat me in the head for these...

SirKeplan:
Do you mean "on" the head, not "in" the head :P, nevertheless your doing good work here :)

Would be nice to globalise the plumes between mods a little more, i.e. have a plume set in the global config maybe, and just set the plumes offset in the mods part .cfgs. I need to look into this.

PhineasFreak:
Today is my "mistake day" i guess :D

Making the plumes work like the engines would be nice but you would still need to do a "per-part" pass to make sure that the position of the plume is correct. Heck, sometimes the plumes are wrong even within a single part model, more so for different ones.

PR #1503, opened 2017-01-21
Fix FASA Explorer SRB Global config error

stratochief66:
The parts were not tagged with the global engine config, and so they were staying as stock.

Tested in game, KSP 1.2.2

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Explorer.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Explorer.cfg
@@ -110,6 +110,8 @@
 		basemass = -1
 		type = PSPC
 	}
+	
+	engineType = BabySergeant
 }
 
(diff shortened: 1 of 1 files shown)


raidernick:
@stratochief66 does this need to be fixed in this file also? https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_USRockets.cfg

I use a modified copy of the fasa sergeants for my juno ii.

stratochief66:
@raidernick 
Nope. You're still setting up the ModuleEngineConfigs within that file. The Global Engine Configs scheme applies a tag (like engineType = BabySergeant) to the part, then the relevant engine config code is added in from a central copy by MM at run time.

PhineasFreak:
really....REALLY? Damn, i am stupid...

Glad that you fixed it. For the record, the VSR Baby Sergeant includes the global engine definition as expected.

PR #1508, opened 2017-01-22
CMES BNTR & RS-68 mount

PhineasFreak:
Change log:

* Update the BNTR part config for the latest changes of the BNTR global engine config.
* Add an extra stack node to the RS-68 Engine Mount to make installation of the CMES RS-68 engine easier (#1488).

@SirKeplan the RS-68 gas exhaust will have to wait, since i have some problems making that work correctly.


SirKeplan:
Yeah the gas gen can wait, i'll do it myself if i get round to it, in other mods it could do with a bit of tweaking too.

PhineasFreak:
The weird thing is that even after copying the setup from LR91 it refused to start correctly: it claimed the one ignition but the engine stopped immediately after igniting, leaving the gas generator running.

PR #1513, opened 2017-01-22
Avoid testflight cfgs causing MM errors with SXT installed.

SirKeplan:
Add an entry cost to AJ10_Mid to avoid causing MM errors. without this `TestFlightRnd.cfg` tries to set values based on the entryCost field, which didn't exist on this part.

EDIT: [https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/TestFlightRnD.cfg#L28](This) is the line causing issue, should this MM code be more robust?


stratochief66:
@SirKeplan If you can think of a way to make it more robust, feel free to try. I don't think it is a big priority, though. We could have it check if entryCost is already set, and if not, set it to something high so that this works, but still provides some red flag for us that one should be set in the config?

PhineasFreak:
Fixing half of that problem would be to check if an entry cost parameter is defined:

```
@PART[*]:HAS[@MODULE[TestFlightCore],#entryCost[*]]:AFTER[zTestFlight]
{
    .
    .
    .
}
```

And you can probably check the part and flag it by:

```
@PART[*]:HAS[~entryCost[]]:FOR[RealismOverhaul]
{
    // Optional: add an empty entryCost field.
    //%entryCost = 0

    @desription ^= :$: (Missing entryCost field)
}
```

The MM PASS can be changed, i just used one that will be present and valid and will not create false check (i.e. FOR[TestFlight]).

PR #1516, opened 2017-01-24
RSB: Fix Common Centaur Propellant Ratio

andrewcchen:
The Real Scale Booster Atlas V Centaur has way too much LOX for RL10-A or RL10-C burning at a 5.5:1 mixture ratio. This patch changes the tank ratios to 5.5:1.


PhineasFreak:
Hmm, i was using an average O/F ratio of 5.85 but since this is a Centaur V then it is OK i suppose.

<nitpick>BTW, why you merged the two fuel and mass comments? I had separated them for easier identification.</nitpick>

andrewcchen:
The Atlas V RL10 (RL10A-2/RL10C) burns at a 5.5:1 O/F ratio instead of the 5.88 of the Delta IV RL10 (RL10B-2).

PR #1518, opened 2017-01-24
SSTU: Orion: Fix missing '}' causing the european service module config to not be applied

andrewcchen:
A config node was not terminated, causing things below it to not work

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Orion.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Orion.cfg
@@ -624,6 +624,7 @@
 			ResourceName = ElectricCharge
 			Ratio = 0.01
 		}
+	}
 
 	@MODULE[ModuleRCS]
 	{


stratochief66:
@andrewcchen thanks!  A recent script we ran to update a ton of things ate a few of those end brackets. Much appreciated.

rsparkyc:
Yep, I'll provide a mass fix.

stratochief66:
@rsparkyc Will your mass fix resolve this example as well? If so, I'll close this and wait for your fix.

rsparkyc:
Yep, it resolved this one: https://github.com/KSP-RO/RealismOverhaul/pull/1520

stratochief66:
Strange, I would expect this PR to now show a conflict with the master, since I merged #1520 . I'm not sure if either merging or closing it would be the right move at this point.

rsparkyc:
I think doesn't conflict because the change is the same, but it's weird that it doesn't reflect that there's now no changes in the diff. Let me look into it.

rsparkyc:
I pulled @andrewcchen's branch locally, and merged it into master.  The resulting commit had 0 changes, so merging this PR will not have any negative impact, and gives @andrewcchen credit for looking into what was wrong.  My suggestion is to merge this.

stratochief66:
Thanks for the research @rsparkyc , and the PR @andrewcchen. The more we know about Git, the better.

Merging!

rsparkyc:
and as a sanity check, you can see that merge commit didn't break anything: https://github.com/KSP-RO/RealismOverhaul/commit/966270396134cb86402620de5697345d2ab49019

PR #1520, opened 2017-01-24
Fix missing bracket

rsparkyc:
Only found four instances where we missed a closing bracket.


rsparkyc:
Just so my process on this is documented:

1. I checked out the version of master before I ran my modification script
1. I reapplied my script, modifying it to account for no output resources
1. I merged in the current version of master, and resolved any conflicts
1. that left the following diff with only 4 spots that it was an issue.

stratochief66:
Thanks again. Documentation is always nice.

I think raidernick noticed the majority of them and had already resolved
them leaving just the 4 you just addressed. Any idea what caused the script
to eat closing brackets some of the time?

rsparkyc:
yep, I was expecting the line after `inputResources` to be `outputResources`, so i would eat it do do my transformation on it.  When it turned out to be a closing `}`, my script didn't break, but it didn't re-emit that closing `}`.  I updated the script to specifically handle a scenario where there were only `inputResources`

PR #1523, opened 2017-01-26
STAR 17 global engine config updates

PhineasFreak:
**Change log:**

* Add some web sources and basic info.
* Set a default part category.
* Remove any fuel tank modules (should be set at the part config level).
* Remove any previous engine configs.
* Remove any leftover RESOURCE nodes.
* Formatting (white spaces).

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Star17_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Star17_Config.cfg
@@ -22,7 +22,7 @@
 
 	@MODULE[ModuleEngines*]
 	{
-		%EngineTYpe = SOlidBooster
+		%EngineTYpe = SolidBooster
 
 		!IGNITOR_RESOURCE,*{}
 	}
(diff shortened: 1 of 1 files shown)


SirKeplan:
Why do you think fuel tank modules should be set at the part level on solids? (not saying that's wrong, just seems more efficient the other way to me)

PhineasFreak:
For that one you can blame my subconscious: we have been setting the engine configs to only have the basic information and the ignition resources but i guess we could add the solid propellant mixtures "in the mix".

In the end: no particular reason, just consistency with the other engine configs. It would be nice if we could force the propellants to be automatically set though.

stratochief66:
I think that solids global configs should contain the solid fuel tanks, IMO. That makes sense according to the DRY principle, and ever solid with X config will have the same amount of fuel, since it represents the same thing entirely.

PhineasFreak:
We will have to fix that then sometime in the future for all solids.

PR #1524, opened 2017-01-26
SXT engines updates (part II)

PhineasFreak:
**Change log:**

* Increase the length of the STAR 17 motor from 0.475 m to 0.69 m.
* Remove leftover engine response parameters (taken care by the thrust curve).
* Add a modular fuel tank for the HTPB propellant.
* Increase the size of the BNTR ~~(was 1/3 of the correct one)~~.
* Remove category definitions (taken care by the global engine configs).
* Remove other modules (taken care by the global engine configs).
* Add default engine parameters for the nuclear engines.
* Rework the plume setups for the STAR 17 and BNTR due to the size changes.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/RO_SXT_Engines.cfg
@@ -712,7 +712,7 @@
 
 //	“BIMODAL” NUCLEAR THERMAL ROCKET (BNTR) PROPULSION: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf
 
-//	Dimensions: 1.87 m x 8.3 m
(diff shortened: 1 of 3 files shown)


SirKeplan:
What is your source for the correct BNTR sizing, it looks to me like they might actually *all* be too large.

According to https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20030014643.pdf the BNTR is 4.3m long with 1 metre nozzle diameter. (though [this](https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf) has 4.5m x 1.56 m), still a lot smaller than the 7-8m long we have them at now.

PhineasFreak:
Hope that i did good. I set the dimensions of the engines to be 1.55 m x 4.5 m (W x L) since these were the values that i came across more often (DRA papers and general BNTR studies).

Still, i have no idea why i set these dimensions in the first place. Probably something to do the the Constellation Essentials BNTR. :P

PR #1525, opened 2017-01-26
Fix TestFlight RnD entryCost checks

PhineasFreak:
As reported and fixed by @SirKeplan in #1513.

Change log:

* Add an entryCost check in the TF after pass (just to make sure that an entryCost field is present).
* Add a patch to add a template entryCost field to the part config if one is missing and tag them for further fixups.

**Note:** the entryCost check is probably reduntant since we already add the missing entry cost in a previous pass.

### DIFF
--- a/GameData/RealismOverhaul/TestFlightRnD.cfg
+++ b/GameData/RealismOverhaul/TestFlightRnD.cfg
@@ -1,17 +1,23 @@
-// Set the global TF RnD settings
+//  ==================================================
+//  Set the global TF RnD settings.
+//  ==================================================
+
 TFRNDSETTINGS
 {
     updateFrequency = 86400
+
     TEAM
     {
         points = 10.0
         costFactor = 1
     }
+
     TEAM
     {
         points = 12.5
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/TestFlightRnD.cfg
```diff
@@ -1,26 +1,48 @@
...
 
-// After pass: fix up RnD
-@PART[*]:HAS[@MODULE[TestFlightCore]]:AFTER[zTestFlight]
+
+//  ==================================================
+//  Before pass: check if any parts are missing an
+//  entry cost field. If so, add one and tag them.
+//  ==================================================
+
+@PART[*]:HAS[~entryCost[]]:AFTER[RealismOverhaul]
```
SirKeplan: Needs to be `:AFTER[RealismOverhaulEngines]` or similar.

otherwise the description gets overwritten later.

PhineasFreak:
Hmm, should we use BEFORE[zTestFlight] then? We already have an AFTER pass for it.

PR #1528, opened 2017-01-26
Constellation Essentials updates & fixes

PhineasFreak:
**Change log:**

* Validate the web sources used on the configs.
* Explicitly set the categories of the parts.
* Update the BNTR and the RS-68 engines to use the new global engine configuration.
* Fix the size of the BNTR engine (was too big).
* Increase a bit the diameter of the trusses in order to fit tanks with larger volume (though still less that it should due to the structure of the models).
* Allow parts to be surface-attached to the End Plate Adapter.
* Adjust the mass and temperature parameters of all parts.

**Notes:**

* This is compatible only with the latest unofficial release by forum user **MeCripp**.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
@@ -47,7 +47,7 @@
     %skinMaxTemp = 673.15
     @CoMOffset = 0.0, 0.0, 0.0
     %emissiveConstant = 0.9
-    %heatConductivity = 0.75	
+    %heatConductivity = 0.75
     %thermalMassModifier = 5
     %radiatorHeadroom = 0.5234
     %bulkheadProfiles = size6, size10
--- a/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
@@ -1,13 +1,13 @@
 //  ==================================================
 //  Sources:
 
-//  1: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20040182399.pdf
-//  2: http://alternatewars.com/BBOW/Space_Engines/AIAA-2004-3863_TRITON.pdf
-//  3: http://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20140009579.pdf
-//  4: http://www.alternatewars.com/BBOW/Space_Engines/RS-68.pdf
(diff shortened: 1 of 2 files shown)


stratochief66:
@PhineasFreak could you please provide a link to the new release by user MeCripp? I suck at digging in the forum, and we should probably start keeping track of the mods we gain compatibility with again and linking them in the OP :)

GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
```diff
@@ -452,88 +385,55 @@
...
         {
-            key = 0 414
-            key = 1 366
+            @key,0 = 0 409
+            @key,1 = 1 357
         }
     }
 
-    MODULE
+    @MODULE[SmarterGimbal]
```
stratochief66: Huh? Not sure I've seen this one before. Is this from a pre-req for this mod?

PhineasFreak:
@stratochief66 yep, here you go: https://dl.dropboxusercontent.com/u/72893034/Test%20Parts/bahaConstellationv1.31RSS.zip

I think that the original release also works but i did not test it (we have an updated one!)

GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
PhineasFreak: Someone, someone did not a perfect job at updating it. This module comes from the "Exsurgent Engineering" mod that has not been updated for at least 2 years. I consider it deprecated since the stock gimbal module includes all the mod fixes and enhancements that were available over the years.

stratochief66: Alright. Can you change this to use an existing gimbal module?

stratochief66:
@PhineasFreak I mean, where is it on the forum? So we can point users to the OP/wherever this comes from, since it isn't something we made or provide support for?

PhineasFreak:
Oh, sorry, i thought that you wanted it for testing the parts. The original post by the user MeCreep is here: http://forum.kerbalspaceprogram.com/index.php?/topic/69581-rssro-constellation-essentials-v131-trusses-engines-adapters-apr21/&do=findComment&comment=2227917

GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg
PhineasFreak: But i did: https://github.com/PhineasFreak/RealismOverhaul/blob/447c8a91f7c6d872a28e13207d9c704884565188/GameData/RealismOverhaul/RO_SuggestedMods/BahamutoD/RO_BahamutoD_Constellation.cfg#L430-Lundefined

PR #1533, opened 2017-01-29
S-II Ullage adjust to GlobalEngineConfig problem

stratochief66:
The FASA S-II ullage motor was taking the name and properties of the BabyS engine, due to inheritance. Removing the engineType flag from this part, and re-enforcing a few properties (ie. fuel type to use) returned to expected behaviour.

Also, LEM base node for decoupling from CSM set to "stagingEnabled = True", as it was no longer showing as stage-able by default for some reason.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Saturn.cfg
@@ -71,6 +71,7 @@
 	{
 		@ejectionForce = 1
 		@explosiveNodeID = top
+		stagingEnabled = True
 	}
 	MODULE
 	{
@@ -1065,6 +1066,10 @@
 	{
 		%minThrust = 101
 		%maxThrust = 101
+		%PROPELLANT[SolidFuel]
+		{
+			%name = PSPC
+		}
 		@atmosphereCurve
 		{
 			@key,0 = 0 251
@@ -1075,6 +1080,8 @@
 	{
 		@volume = 85.4
 	}
+	!RESOURCE[SolidFuel] {}
+	!engineType = DEL
 	@MODULE[ModuleEngineConfigs]
 	{
 		@CONFIG[BabySergeant]


PhineasFreak:
Two questions:

* Why not create the part instead of copying already existing stats from another one? That way there are less things to go wrong if something is changed in the original part (and less hacks to make it work the way it should).
* We have at least 3 mods using the Saturn ullage & separation motors (4 if you count BDB in the future). I think it would make sense to create engine configs for them.

stratochief66:
Feel free to do so.

I don't argue that I've made the perfect fix here. I just found the source of some odd behaviour, then made changes that resulted in the original expected behaviour.

PhineasFreak:
Oh, it's all right then. Sorry if i gave a hostile impression, it was not my intention. Some better wording from my part would be better i guess.

stratochief66:
Yeah, no hostility meant on my part either. I believe that is why people generally use emoticons online. I would've tried to implement a cleaner and better fix, but just tracking down and confirming that bug (and a few others) and their fixes ate most of my spare time, so not much time was left to slap the fix together.

I'd be entirely fine if a superior solution is implemented at some point, this is just "good 'nuf" to regain the desired behaviour for now. You also shouldn't feel any pressure to improve my patchy fix, whatever you're working on is probably more impactful and interesting. :)

PhineasFreak:
The Saturn V Flight Manual has some values for the ullage and retro rockets. Per the document, the stats of the S-II ullage motor are:

* Maximum Thrust: 23000 lbf (102.3 kN)
* Propellant Mass: 336 lb (152.4 Kg - 86.1 units of HTPB)
* Inert Mass: 100 to 104 Kg
* Burn Time: 3.5 s to 3.75 s
* Isp (ASL): ? s (maybe -20/30 s from the VAC Isp?)
* Isp (VAC): 236 s to 256 s (depending on the burn time)

Posting these here for feedback.

stratochief66:
That is a surprising variation in the vacuum ISP! Feel free to configure a global config for it as you choose. For this patch fix, I was just trying to carry the previous behaviour forwards.

I'll just merge this PR now, since it brings the function back, with values close to the correct ones that you've researched. Feel free to open a new PR (if you have a chance, and the interest) creating a new global config and implementing it on this part, and on any others. :)

PhineasFreak:
It is impressive how a small difference in the burn time (0.25 s) translates to the large difference of Isp (20 s)! Anyway, since you approve the info i will make the necessary changes to add the global config and the engine config tags to the parts.

PR #1535, opened 2017-02-01
Saturn V S-II ullage motor global engine config

PhineasFreak:
Preliminary global engine configuration for the Saturn V S-II ullage motor.

Points of intererest/discussion that i'd like to apply not only for that engine config but for every one available in RO:

* [Is it a good idea to patch the original ModuleEngines module like that?](https://github.com/PhineasFreak/RealismOverhaul/blob/abb637c24af1922ccd5e1a8eabbfd8e831e0d4d3/GameData/RealismOverhaul/Engine_Configs/SIIUllageMotor.cfg#L24-L35) I have seen fields and modules being patched for a specific part but not for others and i feel it would be nice to have some consistency over these.
* [Should the minimum thrust be the same as the maximum thrust?](https://github.com/PhineasFreak/RealismOverhaul/blob/abb637c24af1922ccd5e1a8eabbfd8e831e0d4d3/GameData/RealismOverhaul/Engine_Configs/SIIUllageMotor.cfg#L52) I remember @NathanKell saying that the minimum thrust for solids should be zero in order to allow thrust customization but i do not find it realistic (the propellant grain and overall structure of the motor must be changed in reality to reflect this).
* Should we have the ModuleFuelTanks that defines the propellant type and amount within the global engine config (for solid rocket motors only)? This was also a point that @SirKeplan raised for PR #1523.

**Note:** i might be taking this too far...

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/SIIUllageMotor.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/SIIUllageMotor.cfg
@@ -59,7 +59,7 @@
 
             PROPELLANT
             {
-                name = HPTB
+                name = HTPB
                 ratio = 1.0
                 DrawGauge = True
             }
--- a/GameData/RealismOverhaul/Engine_Configs/SIIUllageMotor.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/SIIUllageMotor.cfg
@@ -1,0 +1,78 @@
+//  ==================================================
+//  Saturn V S-II ullage motor global engine configuration.
+
+//  Throttle Range: N/A
+//  Burn Time: 3.75 s
+//  O/F Ratio: 2.77
+
+//  Sources:
+//      NASA - Saturn V Flight Manual: http://history.nasa.gov/ap12fj/pdf/a12_sa507-flightmanual.pdf
+
+//  Used by:
+//      FASA
+//      Real Scale Boosters
+//      OLDD Saturn V
+//  ==================================================
+
+@PART[*]:HAS[#engineType[SIIUllageMotor]]:FOR[RealismOverhaulEngines]
+{
+    %category = Engine
+    %title = S-II Ullage Motor
+    %manufacturer = Thiokol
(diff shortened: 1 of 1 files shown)


stratochief66:
A: I don't think we will ever get consistency, sadly. I agree that just nuking the config, then creating a new one is probably the safest way to go, but I focus on function and not form. I appreciate clean reading and consistent code, but I'm not going to reject functional improvements just because it isn't formatted like my or your changes. Plus, habits are hard to break.

B: I find engines that tapper down to 20% thrust, then just stop instantly unrealistic, and I prefer whatever config setup results in it gradually going down to 0% instead, which makes it easier to decide when to jettison boosters. Here, I don't think it matters for these parts. I'm unaware of what differences of function might occur with configs with max=min, vs min=0; and given how infrequently I configure solids, I'm unlikely to remember once I'm told :p 

C: Yep, seems to be the same question as in #1523. I think that solids global configs should contain the solid fuel tanks, IMO. That makes sense according to the DRY principle, and ever solid with X config will have the same amount of fuel, since it represents the same thing entirely.

stratochief66:
I don't think you're taking things too far in asking questions, how else can we all do our best to stay on the same page or at least know where our opinions and standards differ?

PhineasFreak:
@stratochief66 when i spoke about "consistency" i did not mean formatting but patching available modules the same way as others that have the same functionality. For example, the SXT RD-107 and RD-108 engines had defined the "!IGNITOR_RESOURCE,*{}" in their part configs but not on the global config. Why not let all RD-107/108 engines have the same level of patching as the SXT ones?

Also, by doing some more work now we will have to do less work in the future.

For historic SRM parts we mostly have the thrust curves to take care of the thrust values. For generic ones (like the separation motors) it is really useful to be able to set the thrust but for the others it will seriously mess up the thrust curve. This one might need some more feedback from other RSS/RO/RP-0 players to see how (if) they use the solids and act accordingly.

We will have to take care of the solid propellant tanks to be included in the respective global engine config. A slow and painfull process i have to say but i am sure it will be done! One thing that will definitely need fixing will be the STAR 37.

PR #1536, opened 2017-02-01
RW8 Reaction Wheel Settings

rsparkyc:
This reaction wheel was showing 0 torque in the UI, not it shows up as having 0.1 Nm max torque


stratochief66:
Newton-Metre (N*m) vs Kilonewton-meter (kN*m) ? My recollection is that the torque was listed in kN *m, just as the thrust is shown in kN. I could of course be mistaken.

rsparkyc:
Ahh, no you are totally right. Now that i look the Reaction Wheel window has it listed in `kN` and not `Nm`.  Man, this thing is useless.

rsparkyc:
I updated the PR to only update the data sheet, since the last one had a 404

stratochief66:
It shows how preposterously overpowered reaction wheels are in stock. They are really just used on small-ish satellites. I wish it weren't so, but it is. Dealing with no reaction wheels seems to be one of the greater challenges of going from stock to RSS/RO.

rsparkyc:
yeah, I would consider this a small satellite though, just a ranger III block core, some antennas, and solar panels

PR #1538, opened 2017-02-06
Increased burn time of AJ10-118E to 400 seconds

rsparkyc:
http://www.astronautix.com/a/aj10-118e.html


rsparkyc:
Also, from what I'm seeing online, the AJ10-104 had a rated time of 296 (as opposed to our 300) second burn time.  However, I didn't find anything specific to the "D" model, so i didn't update that.

jwvanderbeck:
> Also, from what I'm seeing online, the AJ10-104 had a rated time of 296 (as opposed to our 300) second burn time. However, I didn't find anything specific to the "D" model, so i didn't update that.

Burn times were typically rounded up a bit on purpose to account for the fact that many burntimes are more driven by the tankage than the engine, and also to try and avoid recreation overhaul.

PR #1541, opened 2017-02-09
Add RO support for the RealEngines pack

PhineasFreak:
Change log:

* Add RO support for the RealEngines pack. Includes support for the following engines:

    * NK-33
    * NK-43
    * RD-58
    * RD-0110
    * RD-0120
    * RD-0124
    * RD-0146
    * RD-170
    * RD-180
    * RD-191
    * RD-0110 Vernier

PR #1542, opened 2017-02-09
Soviet & Russian engines updates & fixes

PhineasFreak:
**Change log:**

* Make sure that the gimbal modules are patched only if they exist. In the case that the engine uses different gimbal ranges for the X and Y axis, adding the "gimbalRange" parameter will break them (it is either gimbalRange or gimbalRangeXX, never both).
* Do not allow the ignitor resources be tweakable by the end user.
* Fix some typos.
* Formatting updates.

**Notes:**

* The gimbal fixes are required by the RealEngines pack: #1541

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/NK33_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/NK33_Config.cfg
@@ -17,7 +17,7 @@
 {
 	%title = NK-15/33
 	%manufacturer = SNTK Kuznetsov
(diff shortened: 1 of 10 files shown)


SirKeplan:
Yep, using `@` MM qualifier seems to make sense for gimbal ranges, we can probably set a precedent for this.

The RD-0124 has 7 degrees gimbal here, and 3.5 [in the other PR](https://github.com/KSP-RO/RealismOverhaul/pull/1541/files#diff-a2bd1fb4335e77bd63e4574f41ecc4cbR688), which is correct?

PhineasFreak:
Nobody knows as far as i can tell. None of the sources (both supplied with the config file and outside ones) even mention anything about the gimbal range value.

I have to manually fix the gimbal range on the RealEngines because it is not patched correctly by the global engine config. I will recycle the global gimbal range until (if?) a proper source for it is found.

SirKeplan:
ahah, that looks like it is copied from spaceflight101

http://spaceflight101.com/spacerockets/angara-1-2pp/

PhineasFreak:
And that is why i left it untouched. You cannot really know who copied what and i do not know if both of these sources can be trusted. On the other hand, for an upper stage engine like this, the 7 degrees of gimbal seem overkill (in fact it may not even be mechanically possible on such a small package).

Should we change it to the new value or keep the old one?

PR #1545, opened 2017-02-11
Took a stab at setting up testflight configs for the Astris engine

rsparkyc:


jwvanderbeck:
Travis appears to be checking the wrong source branch on the PR build.  Odd.

rsparkyc:
My guess is because I first accidentally pushed by branch to the main RO fork, and so it thinks that commit should live there.

GameData/RealismOverhaul/Engine_Configs/Astris_Config.cfg
```diff
@@ -51,7 +51,7 @@
 		}
 		CONFIG
 		{
-			// source: http://www.b14643.de/Spacerockets/Diverse/European_Rocket_engines/engines.htm
```
stratochief66: Unless you found this previous source to be bad, I would suggest keeping this source in addition to the new one.

rsparkyc: It 404'd on me, which is why I removed it

stratochief66: Naw, if it doesn't work, removing it makes sense. I tried to search for a page including info on the engine on b14643, no Astris 1 info, just for 2 so far:
http://www.b14643.de/Spacerockets_1/West_Europe/Europa/Description/Frame.htm

jwvanderbeck: wayback link: 
http://web.archive.org/web/20160731091204/http://www.b14643.de/Spacerockets/Diverse/European_Rocket_engines/engines.htm

jwvanderbeck: Actually better, here is the new link:
http://www.b14643.de/Spacerockets/Specials/European_Rocket_engines/engines.htm

Guess bnumbers rearranged the site a bit

PhineasFreak: For the record, any link from b14643 that was originally under "Diverse" can be found under "Specials", meaning that the broken links just need a word replacement.

Issue #1547, opened 2017-02-13
RealEngine emissive missing for 1st nozzle per engine


Temeter:
The RealEngine emissive are incomplete in RO, while they work fine in stock KSP. Just so I don't confuse terminology, with emissive I mean the red/white heat shimmer of the engine bell and structure.

This is the emissive in a clean stock install, which works fine: http://i.imgur.com/uboIqND.png

And these are the same engines in a clean RO install: http://i.imgur.com/VEoLQwP.jpg

Basically, in RO all (RealEngine) multi-nozzle engine miss one bells emissive, the single nozzle engines don't have any.

PhineasFreak:
So, RO uses the ModuleEnginesRF module, subclassed from the stock ModuleEngines(FX). This is done via SolverEngines. If the emissives are broken then a suspect would be the [ModuleAnimateEmissive module](https://github.com/KSP-RO/SolverEngines/blob/master/SolverEngines/ModuleAnimateEmissive/ModuleAnimateEmissive.cs).

I may be wildly wrong here though so i will tag the coding pros @blowfishpro, @NathanKell and @Starwaster.

Temeter:
I've only observed the issue with RealEngine's. IE SSTU's RD107 works fine.

PhineasFreak:
These engines are also the only ones that i have ever seen something like that so i am more inclined to say that it is caused by the way these engines are set up **in addition** to the new engine module.

Temeter:
Oh, I don't doubt that, never seen this before.^^

The modder, as skilled as he was, seems to be relatively new to the game. He might have something unusual, which conflicted with RO.

Starwaster:
The source for ModuleAnimateEmissive that was linked to can't compile under KSP 1.2.2 as  it references a member that isn't defined. It also hides inherited members (which isn't necessarily a show stopper but makes me think that I'm not looking at updated code)

So are you using an outdated MAE? Or is it even running properly to begin with? What do your logs say when that assembly is loaded?

Or is there an updated assembly and the code in the repository isn't updated yet?

blowfishpro:
`ModuleAnimateEmissive` does not exist in SolverEngines anymore.  It hasn't for quite some time.  I should probably delete it from the repository.

What animation module do those engines use, and do the RO configs change it?  There are issues with certain animation modules not recognizing more than one `Animation` component.

Starwaster:
[...]

PR #1550, opened 2017-02-14
RealEngines heat production fix

PhineasFreak:
Change log:

* Increase the heat production to make the heat animations show up.
* Fix a bad heat production value on the NK-43 engine.

**Note:** the rest of the heat animation fixes will be part of the RealEngines pack itself.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
@@ -79,7 +79,7 @@
         @name = ModuleEnginesRF
         @minThrust = 840.96
         @maxThrust = 1681.93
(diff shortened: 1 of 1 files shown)


Temeter:
Correct me if I'm wrong, but isn't heat production determined by the RO's general Engine_Configs? They certainly have a field for heat production.

Also seems to be a general problem; seeing the same emissives not activating on all engines of RO, stock and FASA. See https://github.com/KSP-RO/RealismOverhaul/issues/1547#issuecomment-279713499

PhineasFreak:
No, you are correct, these should be fixed at the global config. Still, i wanted to update them to fix the typo that i found but it is not a big deal.

Temeter:
I see, it was zero. Can't hurt to fix that!

PhineasFreak:
Note that this PR will be deprecated soon, as i am preparing the RO configs for the new RealEngines release(s).

PR #1553, opened 2017-02-16
RD-0110R vernier global engine config

PhineasFreak:
**Change log:**

* Add a global engine configuration for the RD-0110R vernier engine.


SirKeplan:
Why have this one as a single chamber? inconsistent with the RD-8 and the actual engine is a single thing.

Also i saw 42 degrees for gimbal in one source, not 45.

PR #1555, opened 2017-02-16
RD-120 global engine config

PhineasFreak:
**Change log:**

* Add a global engine configuration for the RD-120 engine.


SirKeplan:
Nice, a few things i noticed.

- 310s sea level Isp for the vac engines is too high(and higher than the RD-120K), a figure around 200s is probably fair here.

- It seems like the RD-120K should have 6 degrees of gimbal. from the sources available

- you say both vac engines have 70%-110% thrust(i understand the RD-120F was the one rated for 110% or 912kN), however shouldn't they both have the same baseline minimum throttle?

PhineasFreak:
* I originally intended to set to to VAC Isp / 2 (175 s) but i confused the values with the ASL - rated RD-120K and set it as 310 s instead. Will fix.

* Forgot to add the gimbal patch as i was testing with the baseline RD-120. Will fix.

* I failed at math again. I went with the 70% as the minimum throttle multiplier for each of the variants (833.56 kN * 0.7 and also 912.02 kN * 0.7). Will also fix.

SirKeplan:
gimbal range and isp still need fixing?

PhineasFreak:
What's wrong with the gimbal range?

GameData/RealismOverhaul/Engine_Configs/RD120_Config.cfg
```diff
@@ -0,0 +1,205 @@
...
+                key = 1 175
+            }
+        }
+
+        CONFIG
+        {
+            name = RD-120K
+            minThrust = 416.78
+            maxThrust = 853.18
+            gimbalRange = 0
```
SirKeplan: range needs to be set here too, or field removed.

PR #1557, opened 2017-02-18
Separate the S5.92 and S5.98M engine configs

PhineasFreak:
**Change log:**

* Separate the S5.92 and the S5.98M global engine configurations.
* Create a new global engine configuration for the S5.98M engine.
* Increase the ignition count of the S5.92 engine.
* Add the low thrust config of the S5.92. It can be toggled in-flight (ModuleHybridEngine) between the nominal and the low thrust options.
* Other general patching updates and fixes.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/S5_92_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/S5_92_Config.cfg
@@ -7,17 +7,21 @@
 //  Burn Time: 2000 s
 
 //  Sources:
-//      KhIMMASH Research Institute - KVD-1 & S5.92 Brochure: http://www.k204.ru/books/vrd/wiki2/PDF/Himmash.pdf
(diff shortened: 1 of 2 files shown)


stratochief66:
Does it really use the same propellant mixing ratio when it is throttled down? That is my concern about using this feature to represent different throttle/mixing ratio/ISP settings for engines, starts to get a bit overcomplicated, filling tanks for a variable mixing ratio and trying to hit zero propellant at stage burnout.

This is an interesting test case, since the propellants don't boil off, which further complicates the picture.

PhineasFreak:
I did not find any sources about the potential O/F ratio changes, apart from the normal O/F range for every given engine (in this case it is between 1.95 and 2.05).

Since they are using pumps to operate (gas generator or staged combustion cycles) they most probably throttle just like every other throttleable engine (kerolox or hydrolox - excluding engines like the J-2 and J-2X), without changing the propellant ratio but in just two discrete throttle settings (instead of a continuous range).

PR #1558, opened 2017-02-19
RO R-40/R-42 global engine configs

PhineasFreak:
**Change log:**

* Add global engine configs for the R-40 and the R-42 series of bipropellant thrusters.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/R40_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/R40_Config.cfg
@@ -34,7 +34,6 @@
         %EngineType = LiquidFuel
         %ullage = False
         %pressureFed = True
-        %ignitions = 50000
 
         IGNITOR_RESOURCE,*{}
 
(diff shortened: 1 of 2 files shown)


stratochief66:
I'm having trouble finding it through the sources, so apologies, but where do you find data for the ignition count? 
@SirKeplan is there a point at which we should just say a high enough demonstrated ignition count can just be 'unlimitted'? Would 142 or 50000 be on the ~arbitrarily restart-able side of that divide?

PhineasFreak:
From the AJR thruster datasheet and Alternate Wars.

SirKeplan:
@stratochief66 that would make sense to me, it's simpler to set engines that have effectively near unlimited ignitions to infinite(as all our RCS already is). 50,000 would absolutely be on that side of the divide in my mind. 134 ignitions on the R-42 seems oddly specific though, and is probably not on the infinite side of the divide.

stratochief66:
Just to confirm, having no ignition count makes it infinite, right? In the back of my memory, I remember the 'right way' being either 0 or -1, but that probably changed at some time in the past few years.

PhineasFreak:
@stratochief66 Yep, and an ignition count of 0 or -1 as you said are also valid for having infinite ignitions.

PR #1562, opened 2017-02-22
RD-805 global engine config

PhineasFreak:
**Change log:**

* Add a global engine configuration for the RD-805 engine.

**Notes:**

* The end values of the TestFlight reliability patch might be a bit high.
* Travis is having problems again...


PhineasFreak:
Same as with #1563. Any new configs from me will use spaces.

Issue #1556, opened 2017-02-17
STAR 37 variants


PhineasFreak:
Currently we have a global config for the STAR 37 but with a catch: the filename referrs to the baseline STAR 37 but the actual config stats are set for the STAR 37**FM**, two completely different motors.

I have compiled some data for the original STAR 37 (required by the Coatl Aerospace GroundOps) and that brings me to the question: should we add all the possible STAR 37 variants in a single config or keep them separate from each other?

* The first option is logical and easy (engine series, both liquid and solid, have their stats defined in the respective config and easy to find/compare/change them between different variants) but it conficts with the option of adding a **ModuleFuelTanks** with the SRM propellant (would have to add an :AFTER[RealismOverhaul] patch for every variant to correct the propellant mass).

* The second option is more messy (more files, more engine tags, more checks, more everything) but it eliminates the propellant mass patching problem.

Thoughts?

raidernick:
if it were configs for my parts I would keep them all separate

PhineasFreak:
Probably. I had the same problem with the S5.92/S5.98M and R-40/R-42 engine configs and i ended up separating them.

PhineasFreak:
I will assume that with the merge of the PR #1563 this is a moot point now. I will keep it open though in the case something else comes up.

PR #1563, opened 2017-02-24
STAR 37 global engine configs updates

PhineasFreak:
**Change log:**

* Rebuild the STAR 37 config to use the baseline STAR 37 operational parameters.
* Create a new STAR 37FM config for the Squad STAR 37 motor.
* Add by default a modular fuel tank with the correct propellant mass.
* Set some default engine parameters.

**Notes:**

* Open discussion for the RO STAR 37 variants here: https://github.com/KSP-RO/RealismOverhaul/issues/1556

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Star37FM_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Star37FM_Config.cfg
@@ -19,299 +19,299 @@
 
 @PART[*]:HAS[#engineType[Star-37FM]]:FOR[RealismOverhaulEngines]
 {
-	%category = Engine
-	%title = STAR 37FM
-	%manufacturer = ATK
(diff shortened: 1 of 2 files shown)


PhineasFreak:
I will have to refuse. I am using spaces for all of my configs (since i hate tabs and they usually break my workflow) so any new file will use them.

rsparkyc:
@PhineasFreak Fair enough.  I know people have mixed feelings between tabs and spaces.  As long as it's consistent per file, I think that's reasonable.

PhineasFreak:
Until there are some formatting rules that one is required to follow before accepting the config, keeping the formatting consistent for a given file is a goo compromise. Remember that this is a community effort, so changes like these are unlike to occur.

Also, if we were to argue about the tabs vs spaces, we should also argue about the tab depth, the way the MODULE{} and RESOURCE{} nodes should be structured, the format of the basic part config (first the RSSROConfig tag, then the MODEL nodes, then the attachment nodes e.t.c.) and other stuff like this. The "tabs vs spaces" wars will never be solved, as i see from other projects.

BTW, i had a conversation with @NathanKell about it (some months ago) and in the end he had the same opinion: enforcing rules now would not work, not without a major restructure of the whole KSP-RO project.

PR #1566, opened 2017-02-25
RealEngines updates

PhineasFreak:
**Change log:**

* Add RO support for all the new engines (RD-0110R, RD-0210, RD-0212, RD-253, S5.92, S5.98M, RD-0110 Vernier, RD-8, RD-805, RD-120).
* Fix all the instances of the of the RD-0110 engine part name.
* Remove the config of the RD-0120 engine red nozzle variant (was removed from the pack).

**Notes:**

* Some part config names were changed. The RealPlume configs have been edited for these changes.
* These engines require the following global engine configs to operate:
    - RD-8 (#1552)
    - RD-0110R (#1553)
    - RD-0110 Vernier (#1554)
    - RD-120 (#1555)
    - S5.92 and S5.98M (#1557)
    - RD-805 (#1562)
    - RD-0210 and RD-0212 (#1565)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
@@ -589,7 +589,7 @@
         minThrust = 6.0
         maxThrust = 6.0
         heatProduction = 100
-        EngineType = LiquidEEngine
+        EngineType = LiquidFuel
         ullage = True
         pressureFed = False
         ignitions = 1
@@ -1373,7 +1373,7 @@
         minThrust = 30.89
         maxThrust = 30.89
         heatProduction = 100
-        EngineType = LiquidEEngine
+        EngineType = LiquidFuel
         ullage = True
         pressureFed = False
         ignitions = 1
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
@@ -1467,8 +1467,8 @@
     @MODULE[ModuleEngines*]
     {
         !engineID = NULL
-        @minThrust = 1627.90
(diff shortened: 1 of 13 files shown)


SirKeplan:
Reviewing this now, will update this comment for things i note.
________

- NK-15/33 has gimbal axes mixed up, engine won't pitch, but will swivel for a roll.
- RD-0110 verniers don't seem to have any plume.

PhineasFreak:
* The NK33 has the same gimbal structure as the NK43. I have no idea why this happens (the MM cache is correct on all aspects).

* I fixed it but it broke again. Everything is the same as before so, again, no idea why it does not have a plume.

* The RD-0212 main engine and the verniers have the exact same name for the gimbal transform ("gimbal") so KSP applies the gimbal range on both of them. This is an issue with how the model itself is structured and currently i cannot do something about it. I have reported it to Alcentar to fix it for the next version.

SirKeplan:
The NK-33 problem is probably the transforms on the model having the wrong axes, So that's probably another thing that only Alcentar can fix.

will take a look at the plume thing if i have time.

rsparkyc:
Side question for @PhineasFreak: you do so much work on this, why don't you join the IRC channel?

PhineasFreak:
@rsparkyc what's the correlation between the IRC and the config work?

rsparkyc:
@PhineasFreak IRC just seems to be a good forum to flush out ideas and interact with each other in a more real-time manner to discuss things related to RO.  Since you're such an active contributor here, I'm sure the community would love to have you in the IRC channel as well.

PhineasFreak:
Sorry for the stupid question but joining the IRC is as simple as finding the #RO IRC via an IRC client? Also, time zones?

PhineasFreak:
With the latest RealEngines release the RD-0212 gimbal should be fixed. Not sure about the NK-33 though.

stratochief66:
@PhineasFreak yep, it is that simple. also, we tend to use the IRC for asynchronous communication as well. We would definitely love to see you there :)

PhineasFreak:
OK, i hope that in the following days i will have a bit more spare time to do some PRs and check out the RO IRC!

PR #1571, opened 2017-03-02
Fixed IVA model for landerCabinSmall

rsparkyc:


stratochief66:
For posterity, the 'fix' here is that the new attached IVA has 2 seats for the 2 person part, rather than 1 seat.

rsparkyc:
I've updated this to have it's own dedicated IVA model, which is a clone of the previous one, with just an extra seat added.  However, in naming the seats, I think the model no longer knows where people should be inside the model, and I'm not sure how to go about fixing that properly.

stratochief66:
Ideal is to have an IVA that matches the inside of the craft, but nobody expects you to create an IVA. Generally, we just attach one that already exists with the proper number of seats. I don't know offhand what RO necessary parts (stock, SXT, Stock Revamp) have 2 seat parts with IVA's.

stratochief66:
@rsparkyc are you fine with us merging this, or do you want to change anything about it beforehand?

PR #1573, opened 2017-03-04
As of KSP1.2, removes interstage nodes

stratochief66:
As of KSP1.2, removes interstage nodes. These appear by default normally, this stops them. The nodes will appear before you drop the part, but disappear/won't show again after you've dropped it (unless manually re-enabled?)

I tested & use this patch in my game. @rsparkyc , does this more or less line up with what you used?


PhineasFreak:
@stratochief66 I have been using this patch as a part of some personal configs for some time now and i can confirm both your observations and the correct operation of it. 👍

stratochief66:
Thanks for the feedback, merging.

PR #1575, opened 2017-03-05
NearFuture Reactors Update

gavinIRL:
**Please read issue 3 detailed near the bottom as it involves a potential issue with a lot of engines**

Fix to update Near Future reactor configs to work with update 1.2 

**Note 0**: Apologies for the amount of writing, however a lot of explanation and reasoning is required for this PR. 

**Note 1**: I am aware of work done such as by Schnobs, etc. however I have not seen any progress on the reactor configs in the past few months (years?) so I decided to have a stab at updating the configs. I have kept to as realistic values as possible, however certain estimations and changes have been made taking realism and balance into consideration. Feel free to critique and ask for changes, I aim to help RO first and foremost. Sources provided for all data, and reasoning for estimations where data is missing is also provided.

**Note 2**: I removed the 455 ton reactor and replaced it with a BES-5 Reactor (FYI flown 31 times, compared to 2 times for next most common reactor). I also added the TOPAZ-I in the 3.75m reactor slot (the aforementioned twice-flown reactor) which I believe gives a more realistic and useful lineup.

**Note 3**: Due to the strange way NF heat is handled, coupled with the fact that the mass quoted for most of the reactors includes the cooling systems, the required cooling has been kept to a value that is low enough to be believably realistic (and not look stupid with 50 large extendable radiators required), but high enough to require enough radiators to look and behave realistically. The exception is the SAFE 400, which based on its power-to-mass ratio should require a lot of added cooling.

**Changelog:**
- The modules have changed for NF Reactors, as a result the current patches don't patch much and stock stats are therefore used in-game. Applied fix to update to new modules (FissionReactor and FissionGenerator).

- Resource [DepletedUranium] is now referred to as [DepletedFuel] in NF. Applied fix for this change.

- Module CoreHeat has been added for all reactors with correct stats.

- All descriptions were changed/updated, one contained a factually incorrect statement (the BES-5 was used on 31 missions, the Topaz-1 was used on 2 missions) and the others were brought more in line with other RO descriptions (include decade, etc.).

- The uranium barrels have had their masses and capacity changed to reflect more realistic and useful values(the 1.25m and 2.5m barrels were very close in stats). See issue 1 below.

- In real life, all of these reactors take 6-24hrs to heat up. A value of 8hrs heat up has been applied across the board (without the addition of the GeneratorSpinupRate field which NF supports, the values are typically ~30mins so it is not too far fetched).

- Replaced 455-ton NuScale with BES-5 reactor.

- Added TOPAZ-I in place of the unused 3.75m reactor.

- Added a commented-out fix to resource definition to change mass of 1 unit Uranium from 11kg to 1kg. See issue 3 below.

- Reduced electric charge storage to more realistic levels, scaled to mass. Although storage should probably be removed for full realism as that is what proc batteries are for. Reduced to 1MJ/ton for now.

- Full source and basic part information provided as comments for all 5 reactors in config file.

- The FissionReactor modules are first deleted before being re-added due to issues with altering the input and output resource ratios, not due to inefficiency.

- The order that the reactors appear in the tech tree has been changed to a more logical order with the 2 new additions.

- Nuclear Reprocessor has been changed to WIP as I haven't looked at updating it to work with 1.2 yet, although it is clear that it currently does not work.

- B9PartSwitch modules fixed for BES-5 and SAFE 400 as they removed nodes by default (bottom and top respectively).

**Issues**

1) All of these reactors are designed in real life for either single use, or in the case of the RAPID-L (and maybe the SAFE 400) two uses at most. In other words, you shouldn't be able to just turn on and off the reactor as you please. One possible route is to use ModuleGenerator with the ability to shut down once started removed, similar to an SRB. Ideally I'd remove the barrels also, or else make them specific to the Rapid-L and SAFE 400 designs. However I have left them for now, with the disincentive of taking 8 hours to heat up to try to negate this behaviour.

2) In order to tweak the heat generated to levels low enough for an appropriate number of radiators, it shows up a relatively arbitrary number for the cooling required/heat generated as a result. This can't be fixed unless ChrisAdderley changes NFE to accommodate this somehow (which is quite an ask for a superficial issue), but for now arbitrary numbers in the display will have to stay.

3) The mass of 1 unit of uranium is currently set to 11kg and I'm not sure why? I have examined the other places that the resource EnrichedUranium is used, and it only appears to be for the engines such as NERVA I and II, Atomic Age, Bimodal, Laztek, Novapunch, and CMES nuclear engines. In each case, changing the mass from 11kg to 1kg does not break anything immediately (consumption of kg/s goes down, but the actual consumption in units/s will stay the same i.e. fuel will last just as long as before) although the ratios for those parts might have to be revisited if they are written for a density of 11kg. I've left the fix in for now and will edit based on what the consensus is about changing the density, although if changing this value causes an issue please point it out. Just making sure that people are aware the NERVA II carries 165kg of Uranium and not 15kg for example. The fix to density should probably happen in RO_Resources.cfg if it is approved (my bad, just wanted to get it out there first). It makes more sense to have 1 unit = 1kg. 

4) Using the stock NFE models isn't quite aesthetically appropriate but I do not have the time to model them myself plus very little data exists on the appearance of the SAFE 400, etc. The 2 new reactors have been sized appropriately from NFE models however.

5) The TOPAZ-I fuel consumption is below the minimum threshold coded by ChrisAdderley, and therefore it will always show that the reactor has a "very long" time left. Again, no fix for this from my side. However it should still be clear from the part description and the current fuel level what the situation is at any time so this issue can be ignored.

**Future Work**

1) Adding more reactors including the SNAP-10A, KiloPower 1kWe to 10kWe series, SAFE 30, SAFE 300, etc.

2) Update Nuclear Reprocessor to the same standard.

3) Look at the NFE radiators.

### DIFF
--- a/GameData/RealismOverhaul/RO_Resources.cfg
+++ b/GameData/RealismOverhaul/RO_Resources.cfg
@@ -285,10 +285,17 @@ RESOURCE_DEFINITION
 @RESOURCE_DEFINITION[DepletedUranium]:FOR[RealismOverhaul]
 {
   @unitCost = -1.097
+  @density = 0.001
+}
+@RESOURCE_DEFINITION[DepletedFuel]:FOR[RealismOverhaul]
+{
+  @unitCost = -1.097
+  @density = 0.001
 }
 @RESOURCE_DEFINITION[EnrichedUranium]:FOR[RealismOverhaul]
 {
   @unitCost = 54.85
+  @density = 0.001
 }
 @RESOURCE_DEFINITION[LeadBallast]:FOR[RealismOverhaul]
 {
--- a/GameData/RealismOverhaul/RO_SuggestedMods/NearFutureTechnologies/RO_NearFuture_Electrical_Nuclear.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/NearFutureTechnologies/RO_NearFuture_Electrical_Nuclear.cfg
@@ -29,14 +29,15 @@
 	{
 		@type = RealismOverhaulStackHollow
 	}
-	@RESOURCE[DepletedUranium]
+	@mass = 0.266
+	@RESOURCE[DepletedFuel]
 	{
-		@maxAmount = 400
+		@maxAmount = 800
 	}
 	@RESOURCE[EnrichedUranium]
 	{
-		@amount = 400
-		@maxAmount = 400
+		@amount = 800
+		@maxAmount = 800
 	}
 }
 @PART[nuclearfuel-125]:FOR[RealismOverhaul]
@@ -46,7 +47,8 @@
 	{
 		@type = RealismOverhaulStackHollow
 	}
-	@RESOURCE[DepletedUranium]
+	@mass = 0.133
+	@RESOURCE[DepletedFuel]
 	{
 		@maxAmount = 400
 	}
@@ -63,7 +65,8 @@
 	{
 		@type = RealismOverhaulStackHollow
 	}
-	@RESOURCE[DepletedUranium]
+	@mass = 0.033
+	@RESOURCE[DepletedFuel]
 	{
 		@maxAmount = 100
 	}
@@ -75,7 +78,7 @@
 }
 @PART[nuclear-recycler-25]:FOR[RealismOverhaul]
 {
-	%RSSROConfig = True
+	%RSSROConfig = False
 	!MODULE[TweakScale]
 	{
 	}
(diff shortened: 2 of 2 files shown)


gavinIRL:
Apologies for the 2 to-dates, I'm tired :sob:

PhineasFreak:
Adding the patch to change the nuclear fuel parameters is bad, really bad:

* The EnrichedUranium is curated by CRP so patching basic characteristics of a resource is out of the question.
* 1 "unit" of HEU has a density of 19100 kg/m3. Since RO requires and uses RF, this 1 "unit" is 1 liter and so we get 19.1 kg/L (19.1 Kg/"unit").
* RO never patches any basic resource characteristics of CRP (and will never do so). If you take a look inside the "RO_Resources.cfg" you will see that it adds some custom resources and changes the **unit costs** (something that should probably be part of RP-0 and not RO). This can be **only** changed via CRP.
* Changing the density may not break the "consumption" of the nuclear-powered engines but it will break the inert/gross mass characteristics of all engines. 1 kg/L to ~20 Kg/L is a huge difference (x20 to be more specific).

gavinIRL:
- Ah ok....I wasn't aware CRP is a no-go. 

- I'm aware of the true density of Uranium in CRP (has it at 10.7kg/l) and I understand what you are saying, but ablator is treated on a mass basis if I am not mistaken, which is how I feel Uranium should be dealt with too. I do realise that goes against why Uranium is mentioned in CRP.

- But can someone confirm that those engines I mentioned are supposed to contain e.g. 15kg or 286.5kg? It just seems strange that almost all (if not all) of the engines have perfectly round numbers for the amount of Uranium they hold if they are designed to hold masses of Uranium to decimal places. I'm not trying to be smart, just trying to see if there is a problem.

- The point I'm trying to make is that it seems as though this patch wouldn't break the gross mass characteristics of the other parts.........as the parts don't appear to have the right mass characteristics in the first place i.e. should all the nuclear engines have the amount of uranium they hold reduced by a factor of 10.7?

- I'm just trying to help, and I'll edit all the numbers for the NF config to reflect the density given by CRP if you feel that is the right approach.

PhineasFreak:
* Who said that the Ablator is treated on a mass basis? It is a resource like all the other resources and, incidentally, it's density is 1 Kg/L. Also, it is one of the worst resources to compare with, as it is one of the least realistic resources for the sake of game-play. Anything under the CRP/RF umbrella is using liters instead of "units".

* We have data for the NERVA and the BNTR engines but the mass of the fissionable Uranium is not well defined (as reactors are usually measured in MWt). We are using about 96.5 Kg of HEU, at least for the NERVA and the BNTR configs. The 289.5 Kg of HEU for the much bigger NERVA II is abstracted, as the mass of HEU depends not only from the engine variant but also on the engine class (within this variant). It can be anything between 35 Kg and 230 Kg.

* This is not the case of "if **i** feel that is the right approach" but what are the rule sets you are abide to work with. The CRP densities are defined as per the real life ones and you cannot change that. Water has a density of 1 Kg/L, period, Hydrogen has a density of 0.0000899 Kg/L, period, **You cannot change that.**

*  The "DepletedFuel" and the "DepletedUranium" resources have exactly the same density (10.97 Kg/L), The only difference between the two is the name. So, it does not matter.

BTW, you might have been looking at the depleted U-235 density (10.97 Kg/L) and not the enriched one (19.3 Kg/L). There is a big difference between the two.

gavinIRL:
- I mean that it is assumed to be handled on a per kg basis because it is a solid, in the same way that uranium is spoken of in kg not liters usually.

- Ah ok, opened up google for a few mins on the NERVA and have 60kg Uranium load for the Alpha for a 2.2ton rocket so that checks out. To me, only knowing reactors and not much about the engines those numbers didn't seem right (60kg Uranium producing 10kWe for 3 to 5 years is by far the poorest ratio I've seen).

- I know it is not just about what I think, hence why I made reference to the fact that what I say goes against the point of uranium being included in CRP. I'm not trying to change the density of uranium in real life here, I was putting forward the case to measure it in kg like it is in reality, as that is what setting the density to 0.001 pretty much does. It is clear now that it isn't the right approach due to factors I wasn't aware of when I put forward the proposal (and which I asked to be enlightened about so thank you @PhineasFreak ).

- I drew attention to the depleted fuel difference to see if there are any issues with it? For example the scenario where someone would want to remove the depleted uranium from any of the engines, I'm not aware of any storage for it? Probably a non-issue as I doubt the engines can be refuelled anyway? Also the nuclear reprocessor uses depleted fuel so you wouldn't be able to reprocess anything from the engines either. Is that going to be an issue?

- I'm going to edit the configs to conform to liters, but can you confirm where you are getting 19.3kg/l from? I'm seeing 10.97kg/l in CRP? I'm just checking so that I put the right numbers in for the edit.



- Any visible issues aside from that? For example the radiator assumptions, etc.

PhineasFreak:
* It has a volume though. And in KSP the resource amount is measured in liters and not Kg. It is weird but it is how it works. Example: in the stock fuel tanks you will see some "units" of propellant. To get the propellant mass you will have to multiply these with the density of the propellant(s), as the "units" in the UI represent volume. In the end, it is just a matter of consistency with the rest of the resources. That problem also existed before CRP and that is  another thing that it "fixes".

* Theoretically you can reuse the engines but the support equipment might not be rated for longer periods of time. BNTRs theoretically can be used for multiple firings and missions over the years.

* Any part module that uses either the DepletedFuel or the DepletedUranium resources will not be able to add/remove/change the amount of it, as the RESOURCE_DEFINITION sets the flow as NO_FLOW and the transfer as NONE. So, we have two identical resources with a different name, Maybe a leftover from other mods that used it? I will have to check with **RoverDude** for that one.

* That's strange...! The nominal density of Uranium is around 19 g/cm3 but CRP has it at 10.97 g/cm3...i was going by the nominal density for all of my calculations so my last point is wrong as it seems. Anyway,

gavinIRL:
- Yeah that's fine, I accept the consistency argument. In reality everything works with mass but that's not an option here and that's fine by me!

- Ok, I could create a duplicate set of barrels that have DepletedUranium instead in a PR later if you think there is a need/demand for it? I'm considering rescaling the current barrels to be the size of one "refuel" for the refuellable reactors i.e. 12kg, 30kg, etc.? And then have a set for the engines that hold 5 and 15 liters? Might as well kill a few birds with one future PR. What I can also do is add a second (or fourth it seems!) module to the reprocessor to handle DepletedUranium.

- For the reactors there is the RadioactiveStorageContainer module which allows the transfer once there is an engineer. Would adding that module to refuellable engines in the future be a bad thing? I know without NF that MM will throw a wobbler over it.

- From what I understand, NF used DepletedUranium in the past but now uses DepletedFuel although the file log doesn't go back far enough to see what was actually used. The main indication was that the previous NF config had DepletedUranium with the really old modules and I remember them working although I have lost most of my old ksp installs so can't confirm :disappointed: 

- So is that a case of annoying Bob Palmer at CRP to change the value or is it best left alone?

PhineasFreak:
To be honest, i have never played with any of the NF mods so i have no idea of how the reactors work and are refueled/managed or how the waste resources are handled. In the real world i would imagine that it is a closed system, not accessed by anyone or anything, not even for the basic maintenance (similar to the NERVA engines).

I also do not know if there were any plans for on-orbit replacement of the fuel rods of the NTRs. Sounds too risky and expensive but, again, i may be wrong. If you want to add the module then a check will take care of the log spam:

```
MODULE:NEEDS[NearFutureElectrical]
{
    .
    .
    .
}
```

gavinIRL:
I only use the NF reactors for end-game craft with ions, although replacing the 455ton reactor with the 385kg reactor will probably make using them earlier a possibility. It is a bit of a tricky mod, although it certainly replicates real-world reactors better than mods that just use a generic resource converter. 

Not all reactors are closed actually! The RAPID-L is specifically designed to be refuellable at least once (the R stands for refuellable if I remember correctly) and there is at least some mention of the SAFE and KiloPower reactors being refuellable. The source file for RAPID-L is in Japanese and I vaguely remember the description of the waste removal and refuelling procedure, but it is intended to be carried out on the moon or mars using something like a hazmat spacesuit so probably doesn't transfer very well to engines in orbit, so I won't touch the engines. I'll actually remove the refuellability of the early reactors that aren't designed for it now.

On another note I did just spent 30 mins of a drive kicking myself for forgetting about NEEDS :laughing:

gavinIRL:
I'll close this and make a new PR later as I might not have the time to make changes in the next week or two

stratochief66:
@PhineasFreak two forms of pure uranium, whether depleted or enriched should have nearly identical densities. It is odd that CRP would have such different densities there, perhaps one represents a uranium metal, the other represents uranium dioxide?

gavinIRL:
EnrichedUranium, DepletedUranium, and DepletedFuel all have the same density in CRP. There is also UF4, at 6.7kg/l, and Uranium Nitride at 14.3kg/l although The first is involved in the manufacture while the second is only theoretical. After a bit of research the densities of the enriched and depleted in CRP correspond to Uranium dioxide, while pure uranium metal has a density of around 19.1kg/l. So I guess CRP has it right?

stratochief66:
@gavinIRL 

This was a really ambitious PR, and it is already very good! Perhaps I'd suggest configuring just one part at a time, so it is easier for us to discuss and resolve issues within a week or so, making the PR fit your free time?

PF made a few good points worth considering though. It would definitely be better to adjust the amount of fuel in an engine to fit any source you find for the amount of fuel in an engine, rather than adjusting the global density of fuel.

My guess would be that last time someone worked on NearFuture configs was long ago enough that DepletedUranium was standard or good enough? There is also a flag that can be added so that the consumed uranium won't be used in the game's ISP calculation, I'll look for that and share it. Whatever CRP has it configured as is what I'd suggest you go by.

A module is only removed with !MODULE{NameOfModule] {}, IIRC, so you'd want to use that here instead. Note the ! instead of -
https://github.com/KSP-RO/RealismOverhaul/pull/1575/files#diff-db2f79e4576f52c9e66a2135128489daR172

Your take on the way of dealing with reactor heat seems sufficiently sound for a start. Really, the thermal system and the way roverdude's and NF parts interact with it are a bit screwy, so we make the best of it we can. Anything functional is >> our broken old configs for these parts, so we're definitely willing to look at anything that ends up functional.

gavinIRL:
@stratochief66 Thanks! For some reason the ! would not remove the module completely, it would create a duplicate if I was to add the code below it in the file which adds the new module i.e. you'd end up with 2 FissionReactor modules in ksp. In a similar way, using the @ for the Input_Resource and Output_Resource to just change those values would not work. I believe it is because of something in the NF code but I couldn't find the cause. As a last resort, I had to use a - instead of a ! (I read documentation somewhere for MM that - is worth trying if there are issues with !) and it is the only method I can find that works unfortunately! Ditto for the * as without it there were issues.

Thanks for the optimism though! I spent a long time lurking here seeing if any progress was being made in areas such as NearFuture, and just decided to finally jump in and help. I suppose it was a bit much throwing in 5 reactors in one go, but I felt like it's more of a case of "if one works they all work" here. My free time is generally infrequent but when I do have time I have a lot of it, meaning a large change such as this PR.

I have looked at many ways of getting NF to use DepletedUranium instead of DepletedFuel but unfortunately it doesn't appear to be a simple case of sticking in an @ statement, so until then I'll make do with DepletedFuel.

If the NF parts could have an in-built radiator module I could experiment with that, but I'm not sure how to do that (unless copy-pasting directly from a proper radiator part would be acceptable?). But yes, my main goal is to fix the old broken configs first, anything extra is just a suggestion.

PR #1580, opened 2017-03-15
Fixed engine plume for lunar ascent engine

rsparkyc:


rsparkyc:
Actually going to delete this, as this may have been fixed in some work @PhineasFreak did (i didn't have the latest master locally)

PR #1586, opened 2017-03-26
Launchers pack Falcon 1 part names update

AlphaMike741:
Like the ULA parts , i updated the names so configs should be functional again .
I have a concern however , should the names of derived parts (+PART[KK_SPX_Something]) change as well?


PhineasFreak:
No, there is no need to change the name of the cloned parts.

AlphaMike741:
@PhineasFreak  Thanks! updated and removed comments regarding it (in RO_KK_Falcon_Tanks.cfg)

raidernick:
this looks good to me

PR #1587, opened 2017-03-31
RealEngines v1.5

PhineasFreak:
**Change log:**

* Add RO support for the new engines: RD-100, RD-107, RD-108 and RD-0124A.
* Update the RD-0110 main and vernier plume patchers to be more robust.

**Notes:**

* ~~The RD-0110 is missing the vernier plumes. This is caused by the naming of the main and vernier engine thrust transforms (they are both called "thrustTransform2"). I have contacted **Alcentar** about that and it will hopefully be fixed in the next update.~~ (should be fixed with the 1.5 update)
* The RD-100 engine requires the new RD-100 global engine config: https://github.com/KSP-RO/RealismOverhaul/pull/1611

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_Russian.cfg
@@ -9,6 +9,12 @@
 //  Encyclopedia Astronautica - RD-58 engine:              http://astronautix.com/r/rd-58.html
 //  Norbert Brügge - Block D upper stage:                  http://www.b14643.de/Spacerockets_1/East_Europe_2/Proton/Gallery/Block-D.htm
 
+//  Encyclopedia Astronautica - RD-103 engine:             http://astronautix.com/r/rd-103.html
+//  Encyclopedia Astronautica - RD-103M engine:            http://astronautix.com/r/rd-103m.html
+
+//  NPO Energomash - RD-107/RD-108 engines:                http://www.npoenergomash.ru/eng/dejatelnost/engines/rd107/
+//  Norbert Brügge - RD-107/RD-108 engine modifications:   http://www.b14643.de/Spacerockets_1/East_Europe_1/Semyorka/Propulsion/history.htm
+
 //  Russian Space Web - RD-0110 engine:                    http://www.russianspaceweb.com/rd0110.html
 //  Encyclopedia Astronautica - RD-0110 engine:            http://www.astronautix.com/r/rd-0110.html
 
@@ -450,9 +456,9 @@
 }
 
 //  ==================================================
-//  RD-103 engine.
+//  RD-103 series engine.
 
-//  Dimensions: - m x - m
+//  Dimensions: 1.3 m x 3.05 m
 //  Gross Mass: 870 Kg
 //  ==================================================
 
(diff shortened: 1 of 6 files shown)


SirKeplan:
Another thing. you say the RD-0110 has missing vernier plumes? looks fine to me.

PhineasFreak:
@SirKeplan For me, the best way to handle the RD-100 series would be to have them all under a common engine config and have the rest be upgrades (as it happened in reality, the engine followed a simlilar form factor as with the A-4 engine and re/used the same hardware but with incremental improvements to it's performance). What do you think about that?

The MM fue/oxidizer/HTP patching are just oversights that have been corrected. I have also fixed the wrong thrust value for the vernier engines (i forgot to change them from the template ones...)

The RD-0110 verniers did not have a plume previously (and it was reported in PR https://github.com/KSP-RO/RealismOverhaul/pull/1566#issuecomment-282496939). I am very sure that it is impossible for the verniers to have the correct plumes due to the thrust transform names. The main engine plume may be wide enough to cover the verniers and then look like they work correctly.

SirKeplan:
@PhineasFreak RD-0110 plumes: http://i.imgur.com/XcsVTih.jpg

SirKeplan:
Having the RD-100 series as one config makes sense, though i'm unsure how to handle it, the RD-100 is almost the same as the A-4,  then the later ones up to RD-103M i think are successive upgrades(and on some the engine is more compact). problem is, should this be part of the A-4 config, or separate?

PhineasFreak:
@SirKeplan It is possible to define a global engine config inside another one (and use both of them with the same engine part config). Example: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/AJ10_Adv_Config.cfg#L149

Using this system we can reuse the standard A-4 engine config for the RD-100 and create a new "RD-100_Config" that will include the RD-102 and RD-103/M engine configs. We will need to just patch the baseline A-4 config for the part title, manufacturer and description in an AFTER pass.

SirKeplan:
@PhineasFreak yeah, that sounds like a good idea.

PhineasFreak:
Now, a simple question: how can we name the new global engine config? RD100_Config?

PhineasFreak:
Updated for the RealEngines v1.5 update. This should hopefully fix the RD-0110 plume inconsistency.

PhineasFreak:
@SirKeplan I found the cause of the RD-0110 plume inconsistency:

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RealPlume_Configs/BobCat/RD0110.cfg#L1

Now, how to handle the name confict? Contact Alcentar and tell him to change the part module name or change the name of the BobCat Extras config?

SirKeplan:
@PhineasFreak Hmmmm, really there should never be parts named exactly after the engine designation, it just gets too confusing. 
In this case I think it's probably just easiest to add an `RO-` prefix to our part like we do here etc. https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg#L7

PhineasFreak:
OK, i'll handle the changes.

SirKeplan:
Something is up with the RD-108 vernier plumes, when the vernier swivels the plume for the verniers at 90 degrees to it swivels instead, misplaced vernier plume.

PhineasFreak:
@SirKeplan Yea, the thrust transforms of the verniers have not been grouped correctly. I'll bug Alcentar about that as it is a model-level fix.

SirKeplan:
Then this may as well be merged in the meantime i guess.

PR #1588, opened 2017-03-31
Coatl Aerospace GroundOps RO compatibility

PhineasFreak:
**Change log:**

* Add RO support for the Coatl Aerospace GroundOps pack. Currently includes the Surveyor lander.

**Notes:**

* The Surveyor lander is currently lacking the Nitrogen ACS thrusters. Use your own for controlling the attitude while in free flight.
* ~~The antenna ranges for the stock antenna modules have been calculated with the maximum effective range of the Tracking Station (level 3 - 50 Tm). From the perspective of spaceflight history, the correct Tracking Station level would be 2 (500 Gm) so the antenna ranges will be vastly lower. Need some feedback on that end, as it will also directly help the ProbesPlus pack.~~

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_GroundOps_Command.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_GroundOps_Command.cfg
@@ -5,7 +5,7 @@
 //  JPL - Surveyor 1-7 missions:            http://www2.jpl.nasa.gov/missions/past/surveyor.html
 //  Gunter's Space Page - Surveyor program: http://space.skyrocket.de/doc_sdat/surveyor.htm
 //  NASA Press Kit - Surveyor A project:    https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19660022877.pdf
-//  Surveyor III Mission Report:            https://www.hq.nasa.gov/alsj/a12/Surveyor-III-MIssionRpt1967028267.pdf
+//  NASA - Surveyor III Mission Report:     https://www.hq.nasa.gov/alsj/a12/Surveyor-III-MIssionRpt1967028267.pdf
 
 //  ==================================================
 //  Surveyor probe core.
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_GroundOps_Command.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_GroundOps_Command.cfg
@@ -1,0 +1,131 @@
+//  ==================================================
+//  Sources:
+
+//  NSSDCA - Surveyor 1 spacecraft:         http://nssdc.gsfc.nasa.gov/nmc/spacecraftDisplay.do?id=1966-045A
+//  JPL - Surveyor 1-7 missions:            http://www2.jpl.nasa.gov/missions/past/surveyor.html
+//  Gunter's Space Page - Surveyor program: http://space.skyrocket.de/doc_sdat/surveyor.htm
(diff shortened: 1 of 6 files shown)


raidernick:
leaving this open per your suggestion in the other PR. this will need to be looked at.

stratochief66:
@PhineasFreak for antenna range, RO should probably have those calculated against the right level of tracking station (level 2)

PhineasFreak:
@stratochief66 As i imagined. I will make the necessary corrections soon then.

stratochief66:
@PhineasFreak 
The Star-37 (for landing the Surveyor) needs the solid fuel tank removed, and an appropriate amount of HTPB added. You could probably just use the central config. I'm not sure if this is the right varient or not, but:
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/Star37FM_Config.cfg

PhineasFreak:
@stratochief66 I have created a baseline STAR 37 global engine config for this motor: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/Star37_Config.cfg

The SRMs, especially the STAR series, are difficult to keep together since every newer variant differs significantly over the previous one. I had opened an issue about it: https://github.com/KSP-RO/RealismOverhaul/issues/1556

stratochief66:
Right, I remember reading that Issue. So, as it stands, we went with the second (more messy) option. (?)

So yeah, I've been going over the patches with a magnifying glass for a bit now, and see that you did indeed apply a global engine config as I would expect. Furthermore, looking at the global config patches themselves, they appear fine, and another engine (RO-STAR-37) appears to have taken the global config fine, but this part didn't (for me) as it still seems to have the RESOURCE as well as no RealFuelTanks.

Any insight? If you've tested this in game, I'll take your word that it seems to work fine, and assume something in my test install here is borked.

PhineasFreak:
I wanted to implement the first option (multiple configs) but in the end it had the potential to be much more of a pain to implement and, most notably, to maintain in the long term: you have to remove any other engine config, force the configuration to the required one, patch carefully the tank modules, the propellant mass, the inert mass...not nice even for my own (high) level of OCD...

They are also different engines so there is that.

If you search for the "ca_landv_srm" part in the MM cache what is the config structure?

stratochief66:
@PhineasFreak 
https://pastebin.com/BhQcMU0R

PhineasFreak:
There is something weird here: while the STAR 37 global config is correctly applied, it is the wrong one (Star-37FM).

Is there any chance that the global config tags are the same for your installation? Or the part that is loaded via a saved craft file? I am asking because:

* The STAR 37 FM config originally had the "Star-37" tag.
* The stats were then replaced by the ones for the baseline STAR 37 (so the tag was 100% correct).
* I created a new STAR 37FM engine config named "Star-37FM" to replace the previous one.

stratochief66:
I did bring up the craft file from the stock version of the craft, but I then replaced the SRM using a fresh one from the VAB. I also added those craft files after the game was loaded and I was already in the VAB, so I'd be very surprised if that altered the MM cache for that specific part at all, let alone in this odd way.

I'll try a fresh boot, avoid any saved craft files, and see if that has an impact.

To test this PR, I downloaded the mod, and also deleted my RO folder in Gamedata and replaced it with the one from the branch for this PR.

stratochief66:
@PhineasFreak yeah, it was just that the version of the Star config on the branch with this PR is out of date:
https://github.com/PhineasFreak/RealismOverhaul/blob/RO-CoatlAerospace-GroundOps-Updates/GameData/RealismOverhaul/Engine_Configs/Star37_Config.cfg

I tested using the proper version from the current main branch, and the part worked fine.

PhineasFreak:
Yea...this time i forgot to add a note saying that the STAR 37 global config is required...

Anyway, i am working on updating the antenna ranges. It will soon be ready.

PR #1595, opened 2017-04-10
Fixed the mass on the AJ10-137

pap1723:
Per the www.alternatewaers.com source in the file, the amount that was in for the mass of .650 was incorrect. The weight listed on the source is in pounds of 650 pounds. This converts properly to KSP weight as 0.2948.


stratochief66:
This seems sound. @SirKeplan or other (@PhineasFreak ?), second opinion logic check?

PhineasFreak:
Yep, the mass was incorrectly set as pounds instead of kilograms. Another source from NTRS also defines the inert mass value as 295 Kg: https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19700026405.pdf

PR #1596, opened 2017-04-13
Kestrel 1B global engine config updates

PhineasFreak:
**Change log:**

* Add a default engine part category.
* Add some default parameters for the engine module.
* Add the diameter to the part description.
* Fix the wrong EngineType field.


raidernick:
merging this per discussion in #1597

PR #1597, opened 2017-04-13
RD-0242M2 global engine config updates

PhineasFreak:
**Change log:**

* Add a default engine part category.
* Add some default parameters for the engine module.
* Add the diameter to the part description.
* Fix the wrong EngineType field.
* Remove any gimbal modules.


raidernick:
@PhineasFreak is this PR and your other 2 open pr's done being added to? I checked these configs and they all look good. I would like to see them merged.
@SirKeplan objections?

PhineasFreak:
The engine config stuff are just bug fixes so they are OK to be merged. I did not get any feedback on the GroundOps PR though so i do not know if something is misconfigured.

Nothing that a later PR cannot fix, since i will be pushing an update for the ProbesPlus parts. Anything that pops up will also be fixed.

raidernick:
I'm merging this then

PR #1602, opened 2017-04-21
J58 plume update by JeeF (potentially trimmable config curves untouched)

gazpachian:
Suggested changes to the RealPlume config for the J58, adding shock cone visuals (and file size to boot). I suspect many config curve values could be pruned to keep the patch size down/in line with other RP configs, but the particle count is reasonable at the very least.

### DIFF
--- a/GameData/RealismOverhaul/RealPlume_Configs/Squad/turboFanEngine.cfg
+++ b/GameData/RealismOverhaul/RealPlume_Configs/Squad/turboFanEngine.cfg
@@ -3,18 +3,337 @@
 	@MODULE[ModuleEngines*]
 	{
 		!runningEffectName = DELETE
-		%powerEffectName = Turbojet
+		%powerEffectName = Hydrolox-Lower
 		%spoolEffectName = Turbojet-Spool
 	}
-    PLUME
+  PLUME
+  {
+      name = Hydrolox-Lower
+      transformName = thrustTransform
+      localRotation = 0,0,0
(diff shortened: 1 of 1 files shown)


PhineasFreak:
Would be possible to split this into two parts, the part patch and a plume config? I imagine that more users would like to have the plume config available to them for other parts and not only for the specific one.

stratochief66:
@PhineasFreak 
This looks like it only changes the plume, to me at least. 
It does poke " @MODULE[ModuleEngines*] ", but only in a way that affects the plume, and I think that occurs in other plume configs as well.

@gazpachian I just haven't had a chance to test this yet, it looks good otherwise.

PhineasFreak:
@stratochief66 I saw the structure of the config and it was the same as the "generic" ones shipped with RealPlume.

It just seems "off" for me to have a very custom plume config for a very specific part. What if someone likes this plume and wants to use it for another jet engine part or another J58 is ROfied by a new mod?

stratochief66:
@PhineasFreak if you can get the same (or a very similar) effect with a simpler config, all the power to you.

I look at a PR and ask "Does this work like the author intended?" and "Is this an improvement over what existed before?"

Once I've tested this, it will hopefully check off Yes and Yes. If I could think of a clever way to get the same effect with less code, I'd test and suggest it.

stratochief66:
Looks good to me. I used two as a jet powered for stage for a Vanguard, just 'cause.

PR #1606, opened 2017-04-27
SSTU-Remove the Upgrades from the adjustable Sizes

pap1723:
This makes the SSTU parts behave much more like procedural parts. When playing a RP-0 career, the tanks, decouplers, fairings and heat shields should all be able to be adjusted.


pap1723:
I am going to re-work this plan.

PR #1610, opened 2017-04-27
Added Test Flight Data for RD-180, F-1 and F-1A

pap1723:
* Added rated burn times and reliability data for all three engines.
* Removed a duplicate entry on the F-1A config file for techRequired and cost.
* Allowed F-1A to throttle to 75% as is stated in the ETS Wiki

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_137_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_137_Config.cfg
@@ -16,7 +16,7 @@
 		type = ModuleEngines
 		modded = false
 		configuration = AJ10-137
-		origMass = 0.650
+		origMass = 0.2948  // per the above source, this was in pounds, former value is 0.650
 		CONFIG
 		{
 			name = AJ10-137


GameData/RealismOverhaul/Engine_Configs/F1_Config.cfg
```diff
@@ -126,3 +129,30 @@
...
+		cycleReliabilityEnd = 0.985
+	}
+}
+
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[F-1A]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
+	{
+		name = F-1A
+		ratedBurnTime = 180					// No data, but if throttable, let's say it would last longer
```
jwvanderbeck: Maybe instead of arbitrarily making the burn time longer because it *can* throttle, instead use the feature of TestFlight that lets TF scale the burn time *based on* current throttle?

pap1723:
Does that mean to just remove the burn time on the config? I didn't know
about that option.

jwvanderbeck:
Not sure if the RO config wrapper exposes it or not, but it can be defined directly in the TestFlight config for `TestFlightReliability_EngineCycle` using the `thrustModifier` property which is a float curve.

Might check in IRC how to do that in the RO config wrapper.  The configs you are modifying here are RO wrappers around the underlying actual TF configs, and I'm not up on how they work.

PhineasFreak:
@pap1723 I just now noticed that you changed the minimum throttle value to 75% **but** we have a separate global engine config for the ETS F-1A variant: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/F-1A_ETS_Config.cfg

pap1723:
@PhineasFreak why is there a separate F-1A for ETS? Shouldn't there only be one? The information that is out there about the F-1A doesn't preclude it from being able to thrust? Or is it unlikely that a real world version would have been able to?

PhineasFreak:
@pap1723 Because the ETS variant is a completely hypothetical engine that also uses the "F-1" name. I guess that we could have the ETS variant being an option of the RL F-1 engine config but i think that it is better to keep them separate.

There is no information that the F-1 (and it's variants) could throttle as it was not designed with that capability in mind. Given that it is the largest **single** chamber engine ever created i do not think that the combustion would be stable in lower throttle values.

Pinging @stratochief66 (as he created the ETS config) to drop by and leave a comment.

pap1723:
@PhineasFreak that makes complete sense to me and I assumed that your answer about throttleing was going to be something very similar.

Obviously it is up to you guys, but I would think that it would be best to have the F-1A ETS as a completely separate engine like it is currently. If it did actually develop the throttling ability, the previous TF data would not be very useful as it is.

I will remove the Pull Request and take a fresh stab at it again.

PR #1623, opened 2017-05-01
X-405 engine updates & fixes

PhineasFreak:
**Change log:**

* Add some web sources to the X-405 global engine config.
* Add a default engine category.
* Add some default ModuleEngines patching (were part of the SXT X-405 part config).
* Add a template gimbal module (patches the main engine gimbal module).
* Add the missing HTP turbopump requirement and engine ignitor.
* ~~The engine can now only be ignited on the ground.~~
* Tweak the positioning of the SXT X-405 attachment nodes.
* Separate the SXT X-405 engine patching into pre and after RO engine passes (as required by the global engine config).
* Remove some deprecated field and module patches from the SXT X-405 engine.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/X405_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/X405_Config.cfg
@@ -1,36 +1,106 @@
-//X-405 engine
-//SXT
+//  ==================================================
+//  X-405 global engine configuration.
+
+//  Gross Mass: 192 Kg
+//  Throttle Range: N/A
+//  O/F Ratio: 2.20
+//  Burn Time: 145 s
+
+//  Sources:
+
+//  NASA - Vanguard, A History (chapter 5):          https://history.nasa.gov/SP-4202/chapter5.html
+//  Norbert Brügge - Vanguard launch vehicle:        http://www.b14643.de/Spacerockets_2/United_States_7/Vanguard/Description/Frame.htm
+//  Alternate Wars - General Electric space engines: http://www.alternatewars.com/BBOW/Space_Engines/GE_Engines.htm
+//  NSSDCA - Vanguard TV3:                           https://nssdc.gsfc.nasa.gov/nmc/spacecraftDisplay.do?id=VAGT3
+//  Encyclopedia Astronautica - X-405 engine:        http://www.astronautix.com/x/x-405.html
+//  Designation Systems - Vanguard launch vehicle:   http://www.designation-systems.net/dusrm/app4/vanguard.html
+
+//  Used by:
+
+//  * SXT pack
+//  ==================================================
+
 @PART[*]:HAS[#engineType[X405]]:FOR[RealismOverhaulEngines]
 {
-	@title = XLR50 (Vanguard X-405) Booster
-	@manufacturer = General Electric
-	@description = Booster engine used on the Vanguard launch vehicle. Very early kerolox gas-generator engine. Diameter: [1.0 m].
-	
-	MODULE // configs for the engine
+	%category = Engine
+	%title = X-405 (XLR50-GE-2)
+	%manufacturer = General Electric (GE)
+	%description = A very early kerolox gas generator booster engine used on the Vanguard launch vehicle. Diameter: 1.0 m.
+
+	@MODULE[ModuleEngines*]
+	{
+		%minThrust = 134.3
+		%maxThrust = 134.3
+		%EngineType = LiquidFuel
+		@useEngineResponseTime = False
+		!engineAccelerationSpeed = NULL
+		@useThrustCurve = False
+		%ullage = True
+		%pressureFed = False
+		%ignitions = 0
+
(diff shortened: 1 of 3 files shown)


stratochief66:
Is it "literalZeroIgnitions = True" that makes it only ignitable on the ground? This feels like a factor that we should put more discussion into before starting to make some engines only ground light-able, as it is more of a career realism/RP-0 style restriction.

What other engines would we be moving over to being ground light-able only? For example, the LR105 Atlas sustainer was a ground lit engine, but I know people in RO/RP-0 like to use it as an upper stage engine. Should this usage be always restricted, or only in RP-0 ?

PhineasFreak:
Well, that is up to the individual and how far you can get before realism steps over the gameplay fan. The truth is that i should have opened an issue and let the members decide if they want this change to happen and, if yes, on what engines.

We can always restrict engines that would not make any sense to have as air/vaccum ignitable:

* BE-4
* F-1/A
* RD-107/108
* RD-170/180/191
* RD-253/273
* RS-68/A

These are mainly booster engines and are lacking the ability to ignite by their own (GSE is required). The RS-25 is a bit of an outlier, as it was considered for the Ares I upper stage (but the difficulty of adapting it for this role was another nail in the coffin, before it was replaced by the J-2X)

Sustainers can be considered as a separate family, since they are optimized for Isp and to operate mostly at vacuum conditions. Example: for the Atlas SLV-3D with the MA-5A propulsion system the sustainer contributed only ~13.8% of the overall thrust. The rest was left upon the LR-89 power plants.

We can keep the sustainers as-is and only touch the engines that are clearly designed as/for boosters. Frankly, it is just silly to use a 3000+ kN engine for an upper stage. We have stock KSP for those crafts! :stuck_out_tongue:

In the case of the X-405 (and the RD-100 series) there would be no way to air lit them or use them for upper stages. Granted, there has been a craft that used the X-405 as a second stage engine (forget where i saw that) but i have only seen it once (and it did not really work).

One could also expand the idea and make under expansion a limitation of using vacuum engines at SL conditions (followed by a big boom and shrapnel flying at all directions).

NathanKell:
A competitor/backup to Atlas for Mercury Launch Vehicle was a three stage
Thor derivative. First stage was Thor, second was a vacuum optimized X405,
and third was flourine based. So airlighting the X405 was deemed not
impossible...

NathanKell:
Also it's worth noting we already, via TF, penalize air lighting with a
higher ignition fail chance

PhineasFreak:
@NathanKell Hmm, is it the "ignitionDynPresFailMultiplier" parameter? I saw it as part of the A-4 engine config but no other engines use it.

Should i just remove the ground ignition requirement from the X-405 and RD-100 engines then?

NathanKell:
If they don't specify they use the default generic-TF-configger curve for
that. Vac engines also often play with the curve, see the LMDE. That one
number is just a multiplier to the curve IIRC.

I am not sure what to say re ignitions because as usual it's a barrier
between reality and plans. The A-9/A-10 system certainly envisioned air
lighting for example. But maybe the better solution there is alternate
configs rather than allowing air lighting the primary config? Dunno

On May 1, 2017 2:19 PM, "Phineas Freak" <notifications@github.com> wrote:

@NathanKell <https://github.com/NathanKell> Hmm, is it the
"ignitionDynPresFailMultiplier"
parameter? I saw it as part of the A-4 engine config but no other engines
use it.

Should i just remove the ground ignition requirement from the X-405 and
RD-100 engines then?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/KSP-RO/RealismOverhaul/pull/1623#issuecomment-298435781>,
or mute the thread

stratochief66:
@PhineasFreak yeah, I think allowing every engine to be air-lit for now would be best, so people can build the rockets of their choosing.

PhineasFreak:
I am implementing that feature on my own "mod" now. If sometime in the future this feature is reconsidered for implementation then it will be ready to go.

Thanks to @ec429 i am now a happy camper!

ec429:
@stratochief66 
> For example, the LR105 Atlas sustainer was a ground lit engine, but I know people in RO/RP-0 like to use it as an upper stage engine.

So, my intention (as per my [lr105-s4 branch](https://github.com/ec429/RealismOverhaul/tree/lr105-s4)) was to have a separate config for the air-lightable version.  So if you want to use it as an upper stage you can, but you have to pay to unlock it.
I think realism demands that you can't air-light an LR87, F-1, or (especially) SSME...
If the consensus is against it then I, like @PhineasFreak, will just have to maintain some MM patches locally, which is fine by me.

stratochief66:
@PhineasFreak works fine, but I don't get a plume from the engine at all.

PhineasFreak:
@stratochief66 Hold on, i think that i forgot to push something...

stratochief66:
Works great, and looks much nicer with the plume :)

PR #1629, opened 2017-05-04
Coatl Aerospace ProbesPlus v0.15RC

PhineasFreak:
### **Change log:**

**General:**

* Add RO support for all the new ProbesPlus parts.
* Lower the maximum temperatures of some parts.
* Adjust the fuel cross feeding rules of some parts.
* Adjust the crash tolerances and breaking forces of some parts.

**Antenna parts:**

* Add CommNet system support to all antenna parts (not dependent on Remote Tech anymore).
* Increase the power consumption of the Pioneer omnidirectional antenna.
* Repurpose the small folding dish antenna to be the Mariner Mars HGA (as it should be in the first place).
* Small tweaks to all generic antenna parts to better fit with each other.

**Command parts:**

* Remove the upgrades from the SAS modules.
* Patch the Voyager and Pioneer command modules to not have hibernation enabled.
* Fix the RT patches (were missing some NEEDS checks).
* Tweak the inert mass of the Voyager probe core to account for the auxiliary scientific and structural parts.

**Control parts:**

* Remove the upgrades from the reaction wheel modules.

**Electrical parts:**

* Revamp the battery stats. They had too small of a capacity (compared to both other batteries and real life battery packs).

**Propulsion parts:**

* Add back the gimbal module to the generic 0.5 kN thruster (**Note 1**).
* Remove the engine module configs from the R-40 engine (now has it's own global engine config).
* Fix the typos on the gimbal module of the Fregat-M upper stage.
* Lower the utilization of the generic propellant tanks (might need further tweaking).

**Science parts:**

* Remove the upgrades from the scientific instruments.
* Fix the scaling of the Voyager Science Boom (was smaller than normal).

**Utility parts:**

* Repurpose the "Quetzal" service module to be an extra spacecraft bus (**LM900**).

### **Notes:**

**Note 1:** There is a missing closing bracket on the base part module that causes MM to fail to patch the part correctly. I have informed **raveloda** about it and it will be fixed in the next version of ProbesPlus (https://github.com/raveloda/Coatl-Aerospace/issues/56).
**Note 2:** This PR requires the following global engine configs from the main RO branch:

* **R-40** and **R-42**
* **S5.92**
* **STAR-37E** (with the https://github.com/KSP-RO/RealismOverhaul/pull/1631 bugfixes)

**Note 3:** This PR requires https://github.com/KSP-RO/RealismOverhaul/pull/1627 in order for the parachute parts to appear without RealChute installed.
**Note 4:** Due to the way RealPlume removes the stock plume nodes, the Voyager Payload Module will be missing it's RCS plumes: https://github.com/KSP-RO/RealPlume/issues/24. A fix for that is provided by https://github.com/KSP-RO/RealPlume/pull/29.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_ProbesPlus_Antenna.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_ProbesPlus_Antenna.cfg
@@ -44,14 +44,14 @@
     @bulkheadProfiles = srf
     @tags ^= :$: communications
 
-	@MODULE[ModuleDataTransmitter]
-	{
-		@antennaPower = 750000
-		@packetInterval = 1.0
-		@packetSize = 3.84
-		@packetResourceCost = 0.01
-		@antennaCombinable = True
-	}
+    @MODULE[ModuleDataTransmitter]
+    {
+        @antennaPower = 750000
+        @packetInterval = 1.0
+        @packetSize = 3.84
+        @packetResourceCost = 0.01
+        @antennaCombinable = True
+    }
 }
 
 //  ==================================================
@@ -130,13 +130,13 @@
 
     !MODULE[DMModuleScienceAnimateGeneric],*{}
 
-	@MODULE[ModuleDataTransmitter]
-	{
-		@antennaPower = 500000
-		@packetInterval = 1.0
-		@packetSize = 6.144
-		@packetResourceCost = 0.012
-	}
+    @MODULE[ModuleDataTransmitter]
+    {
+        @antennaPower = 500000
+        @packetInterval = 1.0
+        @packetSize = 6.144
+        @packetResourceCost = 0.012
+    }
 }
 
 //  ==================================================
@@ -214,14 +214,14 @@
     %fuelCrossFeed = False
     @tags ^= :$: communications
 
-	@MODULE[ModuleDataTransmitter]
-	{
-		@antennaPower = 1500000
-		@packetInterval = 1.0
-		@packetSize = 1.024
-		@packetResourceCost = 0.018
-		@antennaCombinable = True
-	}
+    @MODULE[ModuleDataTransmitter]
+    {
+        @antennaPower = 1500000
+        @packetInterval = 1.0
+        @packetSize = 1.024
+        @packetResourceCost = 0.018
+        @antennaCombinable = True
+    }
 }
 
 //  ==================================================
@@ -296,15 +296,15 @@
     %fuelCrossFeed = False
     @tags ^= :$: communications parabolic
 
-	@MODULE[ModuleDataTransmitter]
-	{
-		@antennaPower = 3500000
-		@packetInterval = 1.0
-		@packetSize = 2.56
-		@packetResourceCost = 52.5
+    @MODULE[ModuleDataTransmitter]
+    {
+        @antennaPower = 3500000
+        @packetInterval = 1.0
+        @packetSize = 2.56
+        @packetResourceCost = 52.5
 
-		!UPGRADES,*{}
-	}
+        !UPGRADES,*{}
+    }
 }
 
 //  ==================================================
@@ -383,15 +383,15 @@
     @bulkheadProfiles = srf, size0
     @tags ^= :$: communications parabolic
 
-	@MODULE[ModuleDataTransmitter]
-	{
-		@antennaPower = 1000000
-		@packetInterval = 1.0
-		@packetSize = 0.512
-		@packetResourceCost = 0.03
+    @MODULE[ModuleDataTransmitter]
+    {
+        @antennaPower = 1000000
+        @packetInterval = 1.0
+        @packetSize = 0.512
+        @packetResourceCost = 0.03
 
-		!UPGRADES,*{}
-	}
+        !UPGRADES,*{}
+    }
 }
 
 //  ==================================================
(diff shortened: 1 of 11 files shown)


pap1723:
I have been playing with these parts, and have had no issues with any of them. I do not have detailed testing reports, but will continue to report back as I continue.

PhineasFreak:
@pap1723 Thank you for the review! I do want some feedback since not many RO players use these parts (so balancing may be off).

pap1723:
@PhineasFreak I will continue to play with them for sure. Do you have IRC or use Discord to chat easier? Anyway, I have started messing around with NSS-OctoSat to generate RO configs and wanted to make sure it wasn't something that you were working on?

PhineasFreak:
I wanted to start hanging out on the RO IRC but most of the time i am at work and it is difficult to do so (time is available but broken into a ton of small "windows").

No, if you want you are free to go ahead and make configs for it (i am working on the Alcentar's R-7 pack - proves to be somewhat on the difficult side).

stratochief66:
@PhineasFreak IRC is pretty good for the 'small bits of time' state of life. It can be used asynchronously fairly well, where you can ask a question or say something, then look in again later when you have a chance to see if anybody responded, and others can do the same with you. We'd love to see you there, even if you don't talk much (or at all), and just skim the chat log when you have free time, at work, home or otherwise. :)

stratochief66:
PR Looks good to me.

@pap1723 or @PhineasFreak , did either of you try landing Venera on Venus? Last time someone tried to land something on Venus, it had trouble surviving the G's, so that part/those parts may need very high G tolerance. The heat shield may also be a bit finicky, since IIRC we've never configured a specifically Venus rated shield before.

If either of you have a Venera craft assembled in RO with this config, I'd appreciate it :)

pap1723:
I haven't tried the venera lander yet, I didn't think it was configured? I
only had one with Surveyor and the satellite parts configured?

PhineasFreak:
@pap1723 all Coatl Aerospace parts are configured for RO (but not RP-0). Venera is configured properly but some notes for testing:

* The parachute parameters are probably misconfigured for "stock". I really recommend testing with RealChute installed.
* Make sure that the entry velocity is smaller than 12 km/s or the heat load will be too much regardless of the entry periapsis.
* The entry periapsis must be lower than 35 km (subterranean better) or the heat shield will not be able to handle the heat load.
* In some cases the decoupler forces the lander to rotate upon decoupling it from the bus. This is probably a collider + attaching node position issue and i will fix it due time.
* Make sure that the lander speed is sufficiently low after parachute deployment so that the heat shield can be safely discarded (there were some cases where the higher drag of the heat shield did not allow a clean separation or it even destroyed the lander after flying away and returning).
* The lander parts can handle 400+ G but they also probably need some DeadlyReentry compatibility patches (does DRE touch the part strengths anymore or is it left up to the stock KSP?).

stratochief66:
@Starwaster ^^ 
last bullet point there

stratochief66:
Yep, confirmed that the lander and bus end up with a rotation after seperation. Thankfully, without persistent rotation I could 'fix' that by timewarping the rotation away.

The chute still ends up heating up really high, and exploding for me. My guess would be that the 'shroud' for the chutes catches heat during entry, passing it to the shield. Not sure how best to handle this. Perhaps having some ablator on the chute to absorb that heat and dissipate it?

PhineasFreak:
* There is an occlusion problem with the parachute "heat shield" (since it is just a fairing) and it allows the heat to traverse freely from inside out. The only way to fix this would be to increase by a ton the heat emissivity of the part and the maxTemp to 5773.15. I'd like to avoid and add extra resources to it, since the main heat shield also works as a ballast, keeping the capsule stable enough for the reentry (the CoM is high enough with just the lander inside).

* From my tests if the capsule reentered heat shield first then it was OK. But if it rotated, even by a small amount, then...bad luck. And you cannot shield it since most of the capsule is comprised of dummy fairings (much like the engine fairings), offering absolutely no protection. A similar situation would be to use a lunar-rated heat shield on the bottom of a generic probe core: if you don't reenter heat shield first then it will explode.

* I might have forgotten to add the KDTU plume config. let me see if i missed something (again...).

* The shrouds might be problematic from the start (as i do not touch them in the RO configs). Will double check using a stock + CA installation (no RO).

* Not sure about the decoupler but this may be the problem with the decoupling action (and not the node position). Will check.

stratochief66:
Alright! Some of these challenges may be worth addressing later, but I'm trying to merge up anything big like this so I can make a release.

Merging!

Starwaster:
Re: DRE g-forces - yes it still does its own implementation of g-force damage and it's based on the parts crashTolerance.  It's something that's been on my mind to revamp for quite awhile. One thing I have in mind to do is adjust the final tolerance by how large/massive the part is. REALLY large parts would be less tolerant for a given value then more compact parts (such as small probes meant for planetary reentry)

the code involved is: 
`                    if (is_engine && damageCube.averageDamage < 1)
                        gTolerance = (float)Math.Pow(UnityEngine.Random.Range(11.9f, 12.1f) * part.crashTolerance, 0.5);
                    else
                        gTolerance = (float)Math.Pow(UnityEngine.Random.Range(5.9f, 6.1f) * part.crashTolerance, 0.5);
`
Which is then used here:
`                if (gTolerance >= 0 && displayGForce > gTolerance)
                { // G tolerance is based roughly on crashTolerance
                    AddDamage(TimeWarp.fixedDeltaTime * (float)(displayGForce / gTolerance - 1), Vector3.zero);
`

PhineasFreak:
@Starwaster so i guess that we still have to implement custom DRE patches for these parts, right? If so, Tantares has some nice patches for it's Venera. I'll take a look and implement them for the next ProbesPlus update.

PR #1634, opened 2017-05-09
Squad antennae CommNet support

PhineasFreak:
**Change log:**

* Fix the CommNet support
* Decrease the maximum temperatures (electronic parts).
* Improve the RT descriptions.

**Notes:**

* The CommNet ranges have been calculated with a mix of the available CommNet station ranges, meaning that they need to be placed in the tech tree accordingly.

| Antenna Name | Station Level |
| :---------------------: | :------: |
| Communotron 16-S | Level 1 |
| Communotron 16 | Level 2 |
| DTS-M1 | Level 2 |
| Communotron HG-55 | Level 2 |
| HG-5 | Level 2 |
| RA-2 | Level 2 |
| Communotron 88-88 | Level 3 |
| RA-15 | Level 3 |
| RA-100 | Level 3 |

* There is a bug regarding the CustomBarnKit DSN ranges that requires a reload of the MM patches before the correct antenna ranges are set. Reload them before entering the SPH/VAB.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Communication.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Communication.cfg
@@ -31,7 +31,7 @@
         @antennaType = DIRECT
         @antennaCombinable = True
         %antennaCombinableExponent = 1
-        @antennaPower = 800
+        @antennaPower = 8000000
         @packetInterval = 1.0
         @packetSize = 1.28
         @packetResourceCost = 0.004
@@ -93,7 +93,7 @@
         @antennaType = DIRECT
         @antennaCombinable = True
         %antennaCombinableExponent = 1
-        @antennaPower = 0.32
+        @antennaPower = 32
         @packetInterval = 1.0
         @packetSize = 0.768
         @packetResourceCost = 0.0005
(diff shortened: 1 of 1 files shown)


stratochief66:
Looks good to me. I don't know the most about the Comms system, so this should get a second set of eyes, and also some in-game testing.

PhineasFreak:
The only thing "missing" is a constant EC consumer even when not transmitting (much like the RT antennae). That can be fixed in a later patch with the ModuleGenerator hack.

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Communication.cfg
```diff
@@ -69,9 +82,22 @@
...
-    %skinMaxTemp = 673.15
+    @maxTemp = 473.15
+    %skinMaxTemp = 473.15
+
+    @MODULE[ModuleDataTransmitter]
+    {
+        @antennaType = DIRECT
+        @antennaCombinable = True
+        %antennaCombinableExponent = 1
+        @antennaPower = 32
```
Bornholio: This is what think needs a value ~4-8Mm to give lunar rating without RT installed.  More if it is supposed to be lunar rated with L1 DSN.

PR #1636, opened 2017-05-10
Aestus/Aestus II engine updates & fixes

PhineasFreak:
**Change log:**

* Port the Aestus II engine config to the main Aestus global engine config.
* Add some web sources.
* Patch some basic engine module parameters.
* Fix the inert mass value of the baseline Aestus.
* Fix (port) the gimballing range.
* Use an O/F ratio of 1.9 for both configs.
* Remove the requirement of propellants for the engine ignition and replace it with an electric one.
* Add some basic TF compatibility patches.
* Increase the size of the Aestus part module.

**Notes:**

* The Aestus and Aestus II have the same overall mechanical structure, with the addition of the turbo pump on the Aestus II (nozzle diameter, gimbal range e.t.c.)
* The diameter of the part module (1.32 m) now refers to the nozzle exhaust instead of the "tank butt" base. **This may make it more difficult to use so some feedback is required.**

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
@@ -4,10 +4,13 @@
 //  Inert Mass: 111 Kg (Aestus), 138 Kg (Aestus II)
 //  Throttle Range: N/A
 //  O/F Ratio: 1.9
-//  Burn Time: 1100 s (Aestus), 600 s (Aestus II)
+//  Burn Time: 1100 s (Aestus), 2500 s (Aestus II)
 
 //  Sources:
 
+//  Astrium - Aestus rocket engine:           https://web.archive.org/web/20120224102231/http://cs.astrium.eads.net:80/sp/launcher-propulsion/rocket-engines/aestus-rocket-engine.html
+//  Astrium - Aestus engine data sheet:       http://projecte-hermes.upc.edu/Enginyeria%20Aeroespacial/2B/Sistemes%20Propulsius/Treball/Altres%20cursos/LL1%20AESTUS/Aestus.pdf
+//  Astrium - Aestus II rocket engine:        https://web.archive.org/web/20120824081115/http://cs.astrium.eads.net:80/sp/launcher-propulsion/rocket-engines/aestus-rs72-rocket-engine.html
 //  Norbert Brügge - Ariane 5 launch vehicle: http://www.b14643.de/Spacerockets_1/West_Europe/Ariane-5/Description/Frame.htm
 //  Norbert Brügge - European rocket engines: http://www.b14643.de/Spacerockets/Specials/European_Rocket_engines/engines.htm
 //  Encyclopedia Astronautica - Aestus:       http://www.astronautix.com/a/aestus.html
@@ -41,7 +44,7 @@
 		%EngineType = LiquidFuel
 		%ullage = True
 		%pressureFed = True
-		%ignitions = 5
+		%ignitions = 20
 
 		IGNITOR_RESOURCE,*{}
 
@@ -73,7 +76,7 @@
 			massMult = 1.0
 			ullage = True
 			pressureFed = True
-			ignitions = 5
+			ignitions = 20
 
 			IGNITOR_RESOURCE
 			{
@@ -110,8 +113,8 @@
 			heatProduction = 56
 			massMult = 1.2432
 			ullage = True
-			pressureFed = True
-			ignitions = 3
+			pressureFed = False
+			ignitions = 20
 
 			IGNITOR_RESOURCE
 			{
@@ -168,7 +171,7 @@
 	TESTFLIGHT
 	{
 		name = Aestus-II
-		ratedBurnTime = 600
+		ratedBurnTime = 2500
 		ignitionReliabilityStart = 0.968
 		ignitionReliabilityEnd = 0.98
 		cycleReliabilityStart = 0.97
--- a/GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
@@ -1,65 +1,178 @@
-//Aestus
-// AIES
+//  ==================================================
+//  Aestus & Aestus II global engine configuration.
+
+//  Inert Mass: 111 Kg (Aestus), 138 Kg (Aestus II)
+//  Throttle Range: N/A
+//  O/F Ratio: 1.9
+//  Burn Time: 1100 s (Aestus), 600 s (Aestus II)
+
+//  Sources:
+
+//  Norbert Brügge - Ariane 5 launch vehicle: http://www.b14643.de/Spacerockets_1/West_Europe/Ariane-5/Description/Frame.htm
+//  Norbert Brügge - European rocket engines: http://www.b14643.de/Spacerockets/Specials/European_Rocket_engines/engines.htm
+//  Encyclopedia Astronautica - Aestus:       http://www.astronautix.com/a/aestus.html
+//  Encyclopedia Astronautica - Aestus II:    http://www.astronautix.com/a/aestus-2.html
+
+//  Used by:
+
+//  * AIES
+//  * RealFuels
+
+//  Notes:
+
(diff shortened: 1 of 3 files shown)


GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
```diff
@@ -1,65 +1,178 @@
...
+	{
+		%minThrust = 30.0
+		%maxThrust = 30.0
+		%heatProduction = 36
+		@useEngineResponseTime = False
+		@useThrustCurve = False
+		%EngineType = LiquidFuel
+		%ullage = True
+		%pressureFed = True
+		%ignitions = 5
```
SirKeplan: What meaningfully limits the number of ignitions on this engine?(pressure fed hypergolic, should be easy to restart usually) my sources simply say "multiple ignitions" no suggestion of a limit at 5.

GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
```diff
@@ -1,65 +1,178 @@
...
+
+		CONFIG
+		{
+			name = Aestus-II
+			minThrust = 55.4
+			maxThrust = 55.4
+			heatProduction = 56
+			massMult = 1.2432
 			ullage = True
 			pressureFed = True
```
SirKeplan: Aestus II is pump fed

GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
```diff
@@ -1,65 +1,178 @@
...
-			ignitions = 0 //unlimited
+			ignitions = 3
```
SirKeplan: My source says mission duty for 5 ignitions, engine designed for 20.

GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
```diff
@@ -1,65 +1,178 @@
...
+
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[Aestus-II]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
 	{
-		@gimbalRange = 5
-		%useGimbalResponseSpeed = true
-		%gimbalResponseSpeed = 16
+		name = Aestus-II
+		ratedBurnTime = 600
```
SirKeplan: If TF works how i remember, then the engine is likely to fail after 600s with this. my source says design life of 2500s

SirKeplan:
The plume on this engine is seriously misplaced, appears metres above the engine.

also something is odd, both variants require pressurised tanks ingame for some reason, some strange bug happening?

PhineasFreak:
@SirKeplan I did not update yet the plume config since i do not know if the new physical size of the engine is acceptable (as i also noted in the PR notes).

I also see the same thing with the Aestus II variant. Maybe RF is caching the ullage info from the first variant and not replacing them correctly? Pinging the RF master @Starwaster.

Starwaster:
They want pressurized tanks because pressureFed = true

PhineasFreak:
@Starwaster Well yes but:

* Aestus I RF parameters: https://github.com/PhineasFreak/RealismOverhaul/blob/927cd8a02200d273c90efdc81c8c423aad851b5d/GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg#L77-L79

* Aestus I RF parameters: https://github.com/PhineasFreak/RealismOverhaul/blob/927cd8a02200d273c90efdc81c8c423aad851b5d/GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg#L115-L117

When you place the engine in the editor and change the variant RF will correctly identify that it will work with non pressure-fed tanks. But upon launching the test craft and trying to ignite the engine you will be greeted by a "Feed pressure too low" warning message.

Starwaster:
Ah , I missed that they were different configs on the same engine... Ok I'll look over the code and see if some weirdness is happening there

Starwaster:
I'm starting to wonder if the pressureFed field is even being read from the MEC configs, except when it's assembling the engine info text.

Just as a point of interest, there is extensive backwards compatibility code for ModuleEngineIgnitor configs when _reading_ the MEC to pull out any possible pressureFed settings (and make a proper config node for them) but I can't actually see that it does much with the field after that.

In YOUR config I think it's just pulling pressureFed out of the main engine config and it never gets overridden by the MEC data.

Starwaster:
It's weird though because MEC doesn't even hard code fields like that; it should even be able to set fields on modules we don't actually support as part of RF.

One thing I notice though is that the type is just set to ModuleEngines...  try setting that to ModuleEnginesRF (or you could even delete that field because that's the default anyway) since that's what you're using for the engine module

Starwaster:
So I just tried adding pressureFed = true to some MEC on a KWR Wildcat. Not defined on the engine so defaults to false, added pressureFed = true to the MMH config and made sure that type is set to ModuleEnginesRF on the MEC main body, so everything should work there.

Engine info correctly says that it's pfed but it's just not loading it at all in flight, so I think something is missing in MEC somewhere

PhineasFreak:
@Starwaster I did a quick test by changing the "pressureFed" field of the ModuleEngines patcher into "False" (to see if the RF information were read from there) but it did not make any difference.

IIRC MEC overrides the fields of the ModuleEngines, right?

Starwaster:
Here, try copying this RealFuels.dll over your existing one and then see if you can still repro the problem. Make sure to test both configurations https://www.dropbox.com/s/jw27k5004rxpih3/RealFuels.dll?dl=1

PhineasFreak:
No luck, the pump-fed engine variant still refuses to operate with non pressurized tanks. I tested both variants, with and without ServiceModule tanks and the results were consistent with the previous RF assembly.

Starwaster:
Uhm. Load a craft with that engine (each variant) and look at the console. One line in there right after loading should say 'pressureFed = '  true/false depending on the config

PhineasFreak:
Oh...sorry...

Give me some time to rerun the tests.

Starwaster:
btw I didn't mean 'load in the editor', load in the world. If it acts like pressure fed but says pressurefed = false then something's wrong somewhere (well something's wrong anyway but that should have fixed it)

PhineasFreak:
Yes, all tests are always done in the flight view scene.

* **Test No. 1 (Aestus with "Default" tank):** "liquidEngineMiniRescale.pressureFed = True"
* **Test No. 2 (Aestus with "ServiceModule" tank):** "liquidEngineMiniRescale.pressureFed = True"
* **Test No. 3 (Aestus II with "Default" tank):** "liquidEngineMiniRescale.pressureFed = False"
* **Test No. 4 (Aestus II with "ServiceModule" tank):** "liquidEngineMiniRescale.pressureFed = False"

So it seems that the field correctly reflects the status of the engine but it does not upon igniting it.

Starwaster:
Wow, I don't get it then, unless something in the code is changing it back to its default value? And I just don't see WHERE that could happen

PhineasFreak:
Log file incoming, just to make sure that i did not do anything stupid...

https://raw.githubusercontent.com/PhineasFreak/DocBin/master/DocBin/log_files/RF_24052017_100827_KSP_log.txt

I followed the same steps while testing as with before.

Starwaster:
Here, try with this one: https://www.dropbox.com/s/jw27k5004rxpih3/RealFuels.dll?dl=1

PhineasFreak:
Great! The new RF assembly did the job! It now correctly respects the engine pressurization requirements.

Starwaster:
Well, the last thing I did was to set isPersistant on the pressureFed field. Which is ok I guess except that's not done on other fields in the class that don't need persistence. Ignitions are handled that way because you have to keep track of how many they have. The ullage field doesn't have it set and it's getting set the same way so why does this one have to be handled like that? Not seeing it.

SirKeplan:
@PhineasFreak i think the physical size currently set in this PR is about right.

stratochief66:
Alright, so with the plume fix, is this PR good to merge?

SirKeplan:
@stratochief66 Yes, unless we want to wait for the new RF

PR #1637, opened 2017-05-13
Move J-2-200klbf to Top of J-2 List

pap1723:
It was confusing to me as a new player (and I am sure others) since the oldest, and worst version of the engine was listed on the bottom. This moves things in line with the other Engine Configs.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/J2_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/J2_Config.cfg
@@ -31,6 +31,40 @@
 		modded = false
 		CONFIG
 		{
+			name = J-2-200klbf
+			minThrust = 676.66
+			maxThrust = 889.325
+			heatProduction = 100
+			massMult = 1.02
+			PROPELLANT
+			{
+				name = LqdHydrogen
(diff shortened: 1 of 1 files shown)


stratochief66:
IIRC, there is a patch in RP-0 that does just that.

RO specifically has the most commonly used J-2 as the default, to avoid confusion that was occurring before we did that.

stratochief66:
To clarify, I don't know anything about the physical ordering of the list, just about which one is the default config, and why.

pap1723:
No, i agree, the J-2 that we all know and love is the default Config. This just moves the worst engine to the top of the list, but it will not be active from default.

NathanKell:
IIRC the reason I put it last is because I generally kept the 'default' first, because people take the first as the default. And you need the real J-2 to do a real mission, so that's what you use in Sandbox. But since we can just set the default config anyway for non-RP0 as the J-2, reordering should be fine. It'll also be save-safe since everything is by ID.

pap1723:
Yeah, I thought moving it would help since the first time I played, I selected that as it was on the bottom of the list. I thought it was the best option! Reading is hard sometimes!

PR #1638, opened 2017-05-18
Delete SSTU_SRB.cfg

pap1723:
The parts in this file do not exist in SSTU. This must be from an older version of SSTU, but does not use correct part names that will work. There are other SRB files in the folder that work.


stratochief66:
These were the old stand-alone parts? Do the modular srbs work, or no?

stratochief66:
@pap1723 poke

pap1723:
I have not had a chance to test the Modular SRB's that you created yet. I will step out of my career tonight to give them a try.

PR #1640, opened 2017-05-19
Update SSTU_XLR81.cfg

pap1723:
Part does not exist in SSTU anymore


stratochief66:
Ahh, indeed. As of 0.5.33.126 from Oct 16, 2016.

PR #1643, opened 2017-05-25
Remove 5x SSTU engines

blowfishpro:
This layout can be selected through the GUI


blowfishpro:
Addresses a few parts of #1625

stratochief66:
@blowfishpro, @pap1723 asked me about this on IRC a few weeks ago. IIRC, the reason I could recall for creating the part was so the price and mass could be set (to include the mount thrust structure it includes) while the GUI derived part couldn't be set to a fixed cost & mass.

PR #1646, opened 2017-05-26
BDB Global Patches

AnticlockwisePropeller:
Create a global config to remove Bluedog Design Bureau part modules that would conflict with RO mods - specifically engine reliability and cryogenic fuel boiloff.


stratochief66:
This seems reasonable.

@PhineasFreak would it be more efficient to have this patch only run if BDB is installed? Or would we not expect a performance difference, searching every part for these modules?

pap1723:
This should definitely be set as :NEEDS[BluedogDB] as it only needs to run
if that is installed.

PhineasFreak:
@pap1723 @stratochief66 yes, an additional check to see if BDB is both easy to implement and needed (better to skip patching and apply it only when BDB is installed than checking all available parts).

AnticlockwisePropeller:
Added the NEEDS: criterion, as suggested!

PR #1654, opened 2017-05-28
Add part upgrade for engine configs proof of concept

NathanKell:
Hope it works.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/LMDE_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/LMDE_Config.cfg
@@ -229,6 +229,7 @@
 		techTransfer = LMDE-H:50
 		reliabilityDataRateMultiplier = 2
 	}
+	*@PARTUPGRADE[RFUpgrade_LMDE-J]/deleteme -= 1
 }
 @PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[TR-201]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
 {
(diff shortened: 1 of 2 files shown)


NathanKell:
Note that it is currently targeted for a TestFlight config but you can move the *@PARTUPGRADE line to the appropriate CONFIG block in the engine configger rather than in the testflight patches section (some engines may not have testflight configs yet). I.e.
CONFIG
{
name = LMDE-J
*@PARTUPGRADE....
....
}

pap1723:
@NathanKell 
Alright, I have done a TON of testing on this over the last few hours. I have tried a lot of different approaches and I cannot get this to work the right way.

First, MM did not like the syntax for this code: `*@PARTUPGRADE[RFUpgrade_LMDE-J]/deleteme -= 1` Not a big deal, I changed it to this and it works correctly:

    @PARTUPGRADE[RFUpgrade_RD-103]
    {
        %deleteme -= 1
    }

The place that we keep running into problems is with this area:

    -PARTUPGRADE:HAS[#deleteme[>0]]
    {
    }

I know that this is the whole point of this attempt is to only show if the different configs are loaded. From what I can tell from the log, what is happening is that the PARTUPGRADE is correctly running, and then immediately being deleted by this set of code. I attempted to modify if to this to get it to run later:

    -PARTUPGRADE:HAS[#deleteme[>0]]:FOR[zzzzROEngines]
    {
    }

However, that seems to not accomplish what we are going for. What we need to have happen is to have that run after all of the configs have been analyzed and determined to exist, and then the deleteme will change over to 0 and should work.

Any ideas?

NathanKell:
This should be OK now.

I dunno why MM didn't like that syntax; it's crucial to making this work, and on my end it did like it. So I'm confused. Anyway I tested this in a standalone KSP install with just MM using [this cfg](https://hastebin.com/avuyumihip.scala) and it created the correct confignodes in configcache. The syntax is crucial because it needs to only decrement deleteme **if an engine config is applied**. Well, technically we're doing it when the TF config is applied, for clarity, but eh, you could put that line in the CONFIG above if you like.

The idea is every time the engine configs are _applied_, then and only then is deleteme decremented (to show that the config is in use), and at the end of the day any partupgrades with deleteme >0 (i.e. never touched) are deleted.

NathanKell:
@pap1723 This stuff is all on the new branch, right? I should close this?

pap1723:
Correct

pap1723:
This is completed

PR #1655, opened 2017-05-28
Increase merlin burn time

pianojosh:
The current Merlin 1 burn times are too short, as far as I can tell.

Watching the CRS-10 technical webcast (all T times in seconds): https://www.youtube.com/watch?v=rUDLxFUMC9c

Ignition occurs at T-3. MECO appears to be at T+146. Total: 149 seconds (Guessing times based on callouts, the timeline at the bottom of the video appears to be incorrect)
Boostback burn begins at ~T+183s. Boostback burn ends at ~T+213. Total: 30 seconds
Entry burn begins at ~T+393. Entry burn ends at ~T+410. Total: 17 seconds
Landing burn begins at ~T+462. Landing burn ends at ~T+481. Total: 19 seconds

The total burn time is 215 seconds.

I would suggest raising the rated burn time of all the Merlin engines to at least this number plus some reasonable padding. 240 seconds seems like a reasonable compromise. I don't have any source for "rated" burn time, but SpaceX static fires each stage for at least a few seconds at McGregor and at the pad before each mission, plus given that they claim the engine is designed to be used for multiple missions with minimal refurbishment, this seems like a reasonable amount of padding. I could even see raising it higher, but I'm hesitant to do so without some source beyond YouTube.

It's certainly likely that the engine, even if it can run multiple mission durations without being overhauled, could not run them in one straight burn, so setting this to a massively high number would also not seem reasonable. However, the current RO rated burn time is clearly too low to emulate existing SpaceX missions, so I feel pretty comfortable at least raising it this much without additional sources.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Merlin1_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Merlin1_Config.cfg
@@ -617,7 +617,7 @@
 	TESTFLIGHT
 	{
 		name = Merlin1A
-		ratedBurnTime = 170
+		ratedBurnTime = 300
 		ignitionReliabilityStart = 0.98
 		ignitionReliabilityEnd = 0.995
 		cycleReliabilityStart = 0.98
@@ -630,7 +630,7 @@
 	TESTFLIGHT
 	{
 		name = Merlin1B
-		ratedBurnTime = 170
+		ratedBurnTime = 300
 		ignitionReliabilityStart = 0.98
 		ignitionReliabilityEnd = 0.995
 		cycleReliabilityStart = 0.98
@@ -658,7 +658,7 @@
 	TESTFLIGHT
 	{
 		name = Merlin1C
-		ratedBurnTime = 255
+		ratedBurnTime = 300
 		ignitionReliabilityStart = 0.98
 		ignitionReliabilityEnd = 0.995
 		cycleReliabilityStart = 0.98
@@ -686,7 +686,7 @@
 	TESTFLIGHT
 	{
 		name = Merlin1D
-		ratedBurnTime = 270
+		ratedBurnTime = 350
 		ignitionReliabilityStart = 0.98
 		ignitionReliabilityEnd = 0.995
 		cycleReliabilityStart = 0.98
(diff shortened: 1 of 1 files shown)


SirKeplan:
This would seem to make sense, apart from now there's the odd situation of the M1C having a longer burn time than the M1D+/++ , this just seams backwards to me. maybe bump up all the M1Ds to same burn time?

pianojosh:
@SirKeplan I suspect that's related to the fact that someone set the initial rated burn times based on an expendable flight profile. As SpaceX has uprated the thrust on Merlin, they have not continued to stretch the stage, probably for aerostructural and road transport reasons. So, as they've uprated the thrust, and thus fuel flow rate, primary mission burn duration has actually shortened.

That being said, I strongly suspect you're right that burn time margins wouldn't really have changed. I'd be fine setting them all to an even 4 minutes, or even 5, but like I said, without real sources to back that up, I tried to preserve the existing pattern, just adding more margin.

I defer to the wisdom of the crowds what the Right Thing To Do is.

PhineasFreak:
@pianojosh you can safely blame me as i was the one that added the (initial) TF configs. Believe it or not, i completely forgot the fact that Falcon 9 also lands after boosting the upper stage...

Increasing the rated burn times can be safely done. Theoretically, there is no reason to limit them, since they can be reused for many launches (IIRC 10 for now - approximately >2150 s from the math that you did).

pianojosh:
Gotcha. That all makes sense then. So the only open question is, do we have any sense of what a reasonable single-mission burn time would be. @SirKeplan, @PhineasFreak 5 minutes seem like a reasonable guess? We haven't seen a non-major-refurbishment reuse yet, and only one reflown mission anyway, so we don't really know if the 10 missions thing is aspirational or realistic yet, anyway. Elon Musk is not exactly known for understating is goals.

PhineasFreak:
We can always update the stats as we gather more information so i guess we can set the burn time of the ASL variants to ~350 s. Since the Merlin VAC version, with 400 s of rated burn time, is a vaccum - rated version of the baseline Merlin engine then i would assume that the ASL variants can also burn for the same amount of time.

pianojosh:
Bumped up the burn times some more. Kept the early variants a bit lower so they stayed lower than their vacuum equivalents.

rsparkyc:
@PhineasFreak @SirKeplan can we merged this?

PhineasFreak:
Yes, the changes are good to merge.

PR #1659, opened 2017-05-29
Bluedog Design Bureau: update for Diamant A & B first stages + clamps

Jerre122:
Improvements to Vesta/Vexin/Valois engine.
Rescaling of Diamant A & B first stage.
Rescaling of launch clamps and launch towers to match the version of FASA.

Further improvements welcome.


Jerre122:
@PhineasFreak Thanks for the review.
In response to your feedback I have made some changes.
I did however set isTweakable to true since only the Vesta engine config requires this resource for its ignition. The other configurations (Vexin or Valois) can leave this resource out. But that's up to the user (as I haven't found a way to do this yet).

It has dawned on me that the massMult doesn't seem to do anything: the valois has a mass multiplier of 0,8 but the mass of the engine remains the same when switching configs. Is there anything I'm doing wrong?

Ravenchant:
@Jerre122 As only the earlier versions need extra ignition liquids and AFAIK addding or removing resources based on the config selected couldn't be done, the lower massMult was meant to negate carrying the unused starter trough the whole burn on the assumption that most people won't change the resource amounts, or at least I failed to find any indication that the engine mass changed between versions (it probably did though)...oops.

Sander815:
Well, according to astronautix the mass did change. However, information on the diamant rockets seems conflicting. With the astronautix page (http://www.astronautix.com/d/diamanta.html) stating that the second stage of both diamant A and B were liquid fueled stages (N2O4/UDMH) when in reality ( according to original sources) it was a solid stage.

Jerre122:
Seems like the massMult does work. Don't know what I did wrong.

raidernick:
@Jerre122 this needs it's conflicts resolved in order to merge

@PhineasFreak what's up with this pr

PhineasFreak:
For a start, the global engine configs are now outdated since costing and tech tree placement is now only handled by RP-1 (so it will conflict with the new EntryCosts and tech tree). The rest of the patches look good enough for a merge (not sure if the latest BDB releases have any effect on them though).

Should we have mod-specific patches inside global engine configs?

https://github.com/Jerre122/RealismOverhaul/blob/7192787a952d353dcdd17a5d5bd0eae2fc4809c7/GameData/RealismOverhaul/Engine_Configs/Vexin_Config.cfg#L205-L225

It would be better if this patch was part of the actual RO part patch.

raidernick:
@PhineasFreak I'm just going to close this then. It can be resubmitted with the changes required for the updates to RO if the op comes back.

PR #1660, opened 2017-05-31
SuperDraco global engine config updates

PhineasFreak:
**Change Log:**

* Add an extra web source.
* Add a default part config category.
* Add some default ModuleEngines patching.
* Make sure that no gimbal or alternator modules exist in the part config.
* Make sure that no resources exist in the part config.
* Add a basic TF config.

**Notes:**

* The burn time has been set as the cumulative burn time of all tests conducted with the same thrust chamber: http://www.spacex.com/news/2014/07/31/spacex-launches-3d-printed-part-space-creates-printed-engine-chamber-crewed

PR #1661, opened 2017-05-31
RealEngines v1.6 RO support

PhineasFreak:
**Change Log:**

* Add RO support for the new SuperDraco engine.

**Notes:**

* Requires the updates SuperDraco global engine config: https://github.com/KSP-RO/RealismOverhaul/pull/1660

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_American.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_American.cfg
@@ -7,8 +7,13 @@
 //  ==================================================
 //  SuperDraco engine.
 
-//  Dimensions: 0.6 m x 0.5 m
-//  Inert Mass: 130 Kg
+//  Dimensions: 0.8 m x 0.67 m
+//  Inert Mass: 150 Kg
+
+//  Notes:
+
+//  * The inert mass value includes the mass of the
+//    SuperDraco mounting hardware (approximately 20 Kg).
 //  ==================================================
 
 @PART[SuperDrago]:FOR[RealismOverhaul]
@@ -19,7 +24,7 @@
 
     @MODEL
     {
-        @scale = 1.0, 1.0, 1.0
+        @scale = 1.333, 1.333, 1.333
     }
 
     @scale = 1.0
@@ -29,7 +34,7 @@
 
     @attachRules = 0,1,0,1,0
 
-    @mass = 0.13
+    @mass = 0.15
     @crashTolerance = 10
     %breakingForce = 250
     %breakingTorque = 250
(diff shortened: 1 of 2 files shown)


pap1723:
@PhineasFreak I believe that you did the configs for RealEngines (I could be wrong) and I havne't used them yet, but @leudaimon has reported that the RD-108 does not have roll support with it's vernier engines.

The CFG for RealEngines looks different from the one for SXT and the information for gimbal is put directly into the RealEngines file. Now, I do not know enough about CFG files to know if this is affecting the results of @leudaimon but I wanted to check with you.

This is in the config for RealEngines RD108:

    @PART[RD108]:AFTER[RealismOverhaulEngines]    
    {
        @MODULE[ModuleGimbal]:HAS[#gimbalTransformName[gimbal2]]
        {
            @gimbalRangeYP = 0
            @gimbalRangeYN = 0
            @gimbalRangeXP = 45.0
            @gimbalRangeXN = 45.0
            @gimbalResponseSpeed = 16
        }
    }

While the config for the SXT version is just using the Engine_Config to get the desired gimbal.
Here is the Engine_Config gimbal information:

    @MODULE[ModuleGimbal]
    {
        %gimbalRange = 4.5
        %useGimbalResponseSpeed = True
        %gimbalResponseSpeed = 16
    }

PhineasFreak:
@pap1723 For the first part of the question: yes, i did all the configs for the RealEngines. The problem with the RD-108 is in it's model and not the RO config: the vernier transforms are incorrectly set, meaning that they are actually moving in different directions (instead of rolling). Do a quick test: load the engine mounted on a filled tank and launch clamps and fire it without decoupling it. Then try to do a roll maneuver. You will see that the vernier plumes are all moving inwards or outwards instead of rolling. That is a job for **Alcentar** to fix and IIRC i have reported it to him.

For the second part: the global engine config for the RD-107 and RD-108 engines assumes that the engine does not have integrated verniers. So, we set the gimbal module to affect the main engine module (instead of the non-existent verniers) otherwise you would not have any thrust vectoring control. The gimbal range is scaled down, since to match the same force from the thrust vectoring of the main engine (972 / 918 kN @ 4.5 degrees) to the force of the verniers (4 x 38 kN @ 45 degrees) you need less of it.

The SXT RD-107/108 engines are also not "real" RD-107/108 engines. The RealEngines pack, that includes separate verniers, require special handling (patches) to achieve the same results, as presented in the RO configs.

pap1723:
I knew you would have a detailed response and valid reasons for all that you did! So the roll function will stay broken until it is correctly fixed by RealEngines on their end?

PhineasFreak:
Unfortunately yes. I am hoping that it will be fixed for the next RealEngines release as it does not seem as a difficult thing (compared to making new engine models).

PR #1663, opened 2017-06-03
Kerbal Atomics Patch

Temeter:
Patch for Kerbal Atomics, adding Nerva I/Nerva II/BNTR, as well as the plumes.

Only two minor issues remaining:

1. Engines start off glowing red, before they visually cool off. Might be a bug; seems a bit inconsistent. (emissives)

2. non-ro engine don't get removed.

Help for those issues would be appreciated.


SirKeplan:
Great, nice to add engines from this pack.
couple of things.
* NERVA I: engine is way to small, nozzel dia should be about 2.95m (it's a big engine)http://www.alternatewars.com/BBOW/Space_Engines/ROVER_NERVA_Engines.htm
* NERVA II should be larger too (same link)

i'm not sure about the emissives, my guess would be to try `@heatProduction` in the ModuleEngines*. Though i see a different bug, they appear cool on the launch pad, but then warm up visually(but only a little)

to remove non RO parts you can use `!PART[Part_Name]:FOR[RealismOverhaul] {}`

Temeter:
Thanks. I couldn't find any information about the Nerva 1+2, only about Nerva Alpha/Gamma, including their pictures - seems like the latter two being small scale versions for the space shuttle payload bay.

>to remove non RO parts you can use !PART[Part_Name]:FOR[RealismOverhaul] {}

Tried that, also removed the RO version of the engines. 

I'll do the rescaling next. Gonna check the @heatproduction again, but otherswise I don't think it's much of an issue.

PhineasFreak:
@Temeter If it removed the parts then something is off (not with the current patches, the MM passes for them are OK). Did you double check the part names? As i see from the part names, it does not seem very difficult to mess up some of that "encoding".

Temeter:
@PhineasFreak Still removes RO engines. Basically, the top of my file looks like this for me:

>!PART[ntr-sc-25-1]:FOR[RealismOverhaul] {} 
!PART[ntr-gc-25-1]:FOR[RealismOverhaul] {} 
!PART[ntr-gc-25-2]:FOR[RealismOverhaul] {} 
!PART[ntr-sc-125-1]:FOR[RealismOverhaul] {} 
!PART[ntr-sc-125-2]:FOR[RealismOverhaul] {} 
>
>@PART[ntr-sc-25-1]:FOR[RealismOverhaul] etcetcetc

I've copied it almost exactly from another RO mod config, probably I'm doing something wrong?

PhineasFreak:
Hmm, maybe the base part modules are copies of the other parts (that get removed by the RO pass)? I cannot see any other reason for these parts to fail.

Temeter:
On a more positive note, the correctly scaled Nerva 1 still looks, despite being over 10 meters high. Nertea makes awesome models!

Temeter:
Oh dang it. What did I do just now? Github desktops interface drives me insane...

Well, at least the KA patches are at done.

Temeter:
Ok, fixed all that nonsense :D 

Done now, I've updated the dimensions to mirror the data points, and the plumes look good too. I might switch around models with the new KA 1.3 engines, but that's for the future.

SirKeplan:
Cool, looks good except the plumes look rather undersized to me, on the NERVAs atleast. 
This is most visible when testing them at sea level. consider using a similar[ plume scale](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RealPlume_Configs/Squad/nuclearEngine.cfg#L13) to the stock NERVA.

Temeter:
Done, increased it to 2.25 and 2.5. I'd almost like to make them bigger, but then the FX starts becoming pixelated. :/

PR #1664, opened 2017-06-03
Fix for the NEXT ion thruster

Max-Q:
Fixing the NEXT ion single and double thruster. Updated stats and working configuration.


raidernick:
Needs conflicts resolved but is good to merge

PR #1666, opened 2017-06-04
Baby Sergeant global engine config tweaks

PhineasFreak:
**Change log:**

* Add some default part config patching (category and engine module).
* Add tech transfer between the two engine configs (as noted by @NathanKell).


NathanKell:
Merged with pleasure :)

PR #1671, opened 2017-06-11
Added S1.5400, LR43-NA, S-3

pap1723:
Please review, especially the TF configs. As they are all very early engines, I gave them low reliability ratings.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/LR105_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/LR105_Config.cfg
@@ -53,6 +53,42 @@
 
 		CONFIG
 		{
+			name = LR43-NA-5
+			minThrust = 240.2
+			maxThrust = 240.2
+			heatProduction = 100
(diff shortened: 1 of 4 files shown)


NathanKell:
1. How about we rename the branch to just TreeRework? Since we should hold off merging this branch down until (a) RP-0's released and (b) an RO release to match it (without this and the other for-the-new-RP0 stuff).
2. LR43 reliability looks about right, maybe a smidge high? Then again Atlas had 3 of them so it comes down fast. S-3 seems about right from what I could find. The S1.5400, very hard to say, but I think it might be a bit high--as the first of these ever, I don't think it'd be that reliabile. Sadly I can't tell from the Molniya data on Gunter's _what_ exactly failed, but there were a lot of failures. I'd do something more like 0.72, 0.92?

pap1723:
We will leave this as a separate branch and make a lot of the changes that will be implemented in the RP-0 major update.

PR #1673, opened 2017-06-12
Add Ludicrous Propulsion Systems to have RP support

BenjaminCronin:


pap1723:
My issue with this, is that the technology does not exist because of how solid fuel works. This would have to be a Future technology, correct?

NathanKell:
I'm confused. These are plumes, so where's the actual configs?

NathanKell:
( @pap1723 if they're just hybrid solids that throttle that's what
SpaceShip One/Two use IIRC. Solid fuel and N2O as oxidizer, with a
variable-opening valve between its tank and the solid motor's chamber)

On Jun 15, 2017 3:35 PM, "NathanKell" <nathankell@gmail.com> wrote:

I'm confused. These are plumes, so where's the actual configs?

BenjaminCronin:
What else do i need to do? I have the real plume configs, do i need some to make the fuels different also?

NathanKell:
Do we are already support Ludicrous Propulsion System but just lack plumes?
Or do we support it not at all?

If you just want to add plumes, you're probably best in the
RealPlume-StockConfigs repo. This repo is rather different :)

BenjaminCronin:
Oh, ok. If you tell me what i need, i would love to add support for RO too!

Bornholio:
The main portion would be a file to set engine types and main real world values for for each engine. 
This would take one the your existing engines and hopefully set its stats to the same as a baseline real world version. Including using real fuels for the liquid portion and possible adding a real fuels tank volume to it.
Take a look at Altair I&II or the Castor series solids, and compare to how LR79 is set up, There is also a LR105 version with integral fuel tank for Atlas configured for FASA.

Then engine, add configuration file if there are multiple configurations for a version.
If there are multiple versions of it in the real world or good fundamental data for them make a file that allows the user to pick the one they want from the Engine Gui button.  For instance if data is available for LOX vs Nitrous Oxide liquid oxidizer you could make the engine switchable with and engine config.
Take a look at how the baby sergeant is configured for this.

Also if you have any run/max burn time information and reliabilty data it is used to create a Test Flight (TF) reliability config that will be located inside of either the RO part type config or inside the engine config.

Great start is to put together a spreadsheet with information first.

Lastly a RP-0  real world cost and date first available information should be gathered if possible.
Include them as // notes in the configs and include any useful links showing the source of the information.

pap1723:
This should not be here as it was intended for Real Plumes. If/When there is some configs created for RO, @BenjaminCronin can open up a new PR.

PR #1677, opened 2017-06-13
BDB Apollo Configs

pap1723:
Apollo configs for all major parts included with Bluedog.
Requires testing.


AnticlockwisePropeller:
Quick question: Why do you delete the pitch motor on the LES? Is it to do with the issue in FAR where it's prone to shredding aborting capsules when it shouldn't?

pap1723:
To be completely honest with you, I do not know. I followed the same thing that was done with the FASA and SSTU LES.

PR #1678, opened 2017-06-14
RO ProceduralParts Updates

PhineasFreak:
**Change log:**

* Adjust the part maximum temperatures to saner values (might still survive reentry though).
* Add a default part category (already taken care of by the PP side but...why not).
* Add a default part manufacturer.

Tagging @rsparkyc (https://github.com/KSP-RO/RP-0/issues/691)

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
@@ -47,7 +47,7 @@
 @PART[proceduralNoseCone]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	%category = Thermal
+
 	@title = Nose Cone [Procedural]
 	@manufacturer = Generic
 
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
```diff
@@ -1,7 +1,14 @@
 @PART[proceduralBattery]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	%title = Battery [Procedural]
+
+	%category = Electrical
+	@title = Battery [Procedural]
```
rsparkyc: Is there any benefit of using @ over %? % makes it if it doesn't exist, otherwise does an update, correct?  Is % any less efficient?  It would seem like anytime we want to make _sure_ we have something set, we should use %, where @ is "i only want to change this if it has a value"

NathanKell:
Saves a branch.

rsparkyc:
Ok, so it sounds like, if we're pretty sure a value is already going to be there, we should be using `@`?

NathanKell:
It's also good for readability, % is ambiguous between replace and add
(i.e. if you don't know the original file).

If for a global engine config where we don't necessarily know what we're
patching, % is perfect.

Issue #1686, opened 2017-06-19
Communotron 16 has extremely low non-RT value


Bornholio:
Communotron 16 has range value of 32 on comnet.

\GameData\RealismOverhaul\RO_SuggestedMods\Squad\RO_Squad_Communication.cfg
line: 96        @antennaPower = 32

Suggest 8Mm  giving 632Mm vs L2DSN
line: 96        @antennaPower = 8000000

PhineasFreak:
Did you use CBK while testing?

https://github.com/KSP-RO/RealismOverhaul/pull/1634

The ranges of the antennae under CommNet are **exactly** the same as the ranges under RT **for the respective tracking station range**.

Bornholio:
CBK installed while testing, the condition I noticed in was while at moon, released impactor that had two sputnik antenna, main craft had 4xCommu16's. Main craft then lost communications as the impactor left range.

PhineasFreak:
Something that i just remembered: did you enforce a reload of the MM cache? If not, can you do so and report back with the same craft file?

Bornholio:
> Edit: foget it, it is probably the antenna exponent(s) that have been set incorrectly...

thoughts on correcting it? the only comnet exponent I saw was the multiple antenna multiplier
```
    @MODULE[ModuleDataTransmitter]
    {
        @antennaType = DIRECT
        @antennaCombinable = True
        %antennaCombinableExponent = 1
        @antennaPower = 32
        @packetInterval = 1.0
        @packetSize = 0.768
        @packetResourceCost = 0.0005
    }
```

versus for the Surface Antenna

```
    @MODULE[ModuleDataTransmitter]
    {
        @antennaType = DIRECT
        @antennaCombinable = True
        %antennaCombinableExponent = 1
        @antennaPower = 8000000
        @packetInterval = 1.0
        @packetSize = 1.28
        @packetResourceCost = 0.004
    }
```

PhineasFreak:
Set the antenna power from 32 to 3200 (or even 320000 if you want ease of access even with the LV 1 tracking station).

Bornholio:
[img]http://i.imgur.com/QsBqM9j.png[/img]
[img]http://i.imgur.com/JBFPonP.png[/img]
neither setting (3,200 or 320,000) are sufficient to reach Earth from the Moon

PhineasFreak:
[...]

PR #1688, opened 2017-06-21
[Bug Fix] Communotron 16 CommNet range

PhineasFreak:
**Change log:**

* Increase the range of the Communotron 16 omnidirectional antenna to mach the RT stats of 400 Mm (https://github.com/KSP-RO/RealismOverhaul/issues/1686).

**Notes:**

* This is needed because CommNet does not follow the same antenna rules as RT regarding the multiple antenna multipliers.


Bornholio:
Thanks PhineasFreak

PR #1692, opened 2017-06-26
[Bug Fix] RO Aerobee engine plume

PhineasFreak:
**Change log:**

* Fix the flare position.
* Remove the fxOffset parameter (now patched via the global RealPlume engine patcher).

**Notes:**

* Same issue as with the 1 kN thruster. Any parts that define AFTER RealPlume passes might also be broken.

### DIFF
--- a/GameData/RealismOverhaul/RealPlume_Configs/RO/ROAerobeeSustainer.cfg
+++ b/GameData/RealismOverhaul/RealPlume_Configs/RO/ROAerobeeSustainer.cfg
@@ -9,8 +9,11 @@
         name = Hypergolic-Lower
         transformName = thrustTransform
         localRotation = 0.0, 0.0, 0.0
-        localPosition = 0.0, 0.0, 0.1
-        fixedScale = 0.18
+        flarePosition = 0.0, 0.0, 0.85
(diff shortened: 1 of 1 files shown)


Theysen:
I submitted a 1 kN plume fix lately which is already merged. Is there still issues after fixing the position and size manually?

PhineasFreak:
@Theysen No because you replaced the localPosition Z value with the one from the AFTER patch (so it will work the same):

https://github.com/KSP-RO/RealismOverhaul/blob/9b594553cbd20b01926530cf72243e67148ae149/GameData/RealismOverhaul/RealPlume_Configs/Squad/microEngine.cfg#L23

While previously this was done via the AFTER pass:

https://github.com/KSP-RO/RealismOverhaul/blob/9b594553cbd20b01926530cf72243e67148ae149/GameData/RealismOverhaul/RealPlume_Configs/Squad/microEngine.cfg#L66-L83 (note the line 78)

(BTW the above piece of code is also useless now)

**Offtopic:** The best way to create/configure/update the plumes now is via the flare/plumeScale and flare/Plume position (no idea why these are not used by default, the localPosition and fixedScale have many drawbacks).

PR #1693, opened 2017-07-01
Enable ignitions for the RD-100

Theysen:
- reports stated the engine can't be ignited anymore
- added 1 ignition to every derivative

- please check if I didn't break anything since there should have been a reason why this was 0 before and it worked (based on my memories)

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD100_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD100_Config.cfg
@@ -51,7 +51,7 @@
         %useThrustCurve = False
         %ullage = True
         %pressureFed = False
-        %ignitions = 0
+        %ignitions = 1
 
         !IGNITOR_RESOURCE,*{}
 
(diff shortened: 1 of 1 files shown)


PhineasFreak:
It is because this engine was supposed to only be able to be ignited on the ground:

https://github.com/Theysen/RealismOverhaul/blob/bfa381c75780af993c5ab5814a686ef3ad750409/GameData/RealismOverhaul/Engine_Configs/RD100_Config.cfg#L76

Also remove this line in order to make sure that ignitions will work at all times.

Theysen:
@NathanKell told me to leave this line at True, that's why I am asking to make sure this change does work as intended.

PhineasFreak:
In that case you can keep it.

PhineasFreak:
@Theysen : Forgot to point you to a small discussion about this feature a while back: https://github.com/KSP-RO/RealismOverhaul/pull/1623

NathanKell:
So...what do we do with this? Is it working as designed (close pull) or does something need fixing?

PhineasFreak:
@NathanKell Yes, the PR is perfect, no problems here. @Theysen could also safely remove the **literalZeroIgnitions = True** line if he wanted, as it is not used anymore.

NathanKell:
ok!

PR #1707, opened 2017-07-15
[Maintenance] NAA-75-110 global engine config

PhineasFreak:
**Change log:**

* Add some web sources and general config notes.
* Add some default part and engine module patching.
* Fix the inert mass value (the previous one assumed the FASA configuration that includes the thrust frame, engine fuselage and stabilization winglets).
* Add a gimbal module (simulates the jet vanes).
* Use an O/F ratio of 1.3 for both engine variants.
* Use the average value for the ASL Isp value of the A-6 variant (values are reported to be between 214 s and 218 s).
* Add the requirement of HTP for the turbopump.
* Add ignition and ullage parameters.
* Make sure that no other engine configs, alternator modules and/or resources exist in the part module.
* Decrease the rated burn time of the A-7 variant from 165 s to 155 s.

**Notes:**

* From the **Historic American Engineering Record AL-129-A** (page 54) we find that the operational O/F ratio for the A-6 was 1.3. The propellant tanks were **lengthened** for the Jupiter-C/Juno-I variants so we can assume that the O/F ratio for the A-7 engine remained similar to the A-6 one (the 1.73 reported previously was the **ideal** ratio between the propellants and not an operational one). It should probably be tweaked even further.
* The thrust increase does not match the reported percentage increase. Using the reported ASL thrust values (78000 lbf to 83000 lbf) we get an increase of ~6.4%.
* The HTP usage values were calculated from the overall propellant load of a Redstone SRBM, as reported by the **History of the Redstone Missile System** (page 72).
* The FASA NAA-75-110 engine part config will need to be updated to add the missing mass of the engine fuselage and the rest of the structural parts.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/NAA75_110_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/NAA75_110_Config.cfg
@@ -1,69 +1,187 @@
-//NAA 75-110 engine
-//FASA
+//  ==================================================
+//  NAA 75-110 engine series global engine configuration.
+
+//  Inert Mass: 670 Kg
+//  Throttle Range: N/A
+//  O/F Ratio: 1.3
+//  Burn Time: 145 s (A-6), 155 s (A-7)
+
+//  Sources:
+
+//  DTIC - History of the Redstone Missile System:                     http://www.dtic.mil/docs/citations/ADA434109
+//  Alternate Wars - Historic American Engineering Record AL-129-A:    http://www.alternatewars.com/BBOW/Space_Engines/HAER_AL-129-A.pdf
+//  Heroic Relics - Redstone Rocket Engines (A-6 and A-7):             http://heroicrelics.org/info/redstone/redstone-engines.html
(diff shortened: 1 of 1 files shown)


NathanKell:
Good stuff! My only comment is that adding an HTP requirement means we need to add appropriate tank volume to the FASA Redstone and Jupiter-C/Juno tanks since the required volume will be higher (or, as with the A-4, put the HTP in the engine model, which might be a better idea. Although it will require the player lowering the HTP amount when using the engine in Redstone config, since the full Jupiter-C amount of HTP should be put in the engine, ModuleEngineConfigs doesn't support changing resource quantity)

PhineasFreak:
Yea, should have also added these notes for FASA (i only thought about the actual inert masses and not the pre-made craft propellant requirements).

Even it it made a lot of sense (the model has the HTP tanks), i wanted to avoid adding it to the engine because of that reason: the player may opt to use any of the 4 LV configurations and filling the tanks automatically upon craft "fueling" is easier than trying to fiddle with a slider. For the same reason i did not add HTP to the RD-100 series engine either.

Since i "broke" the FASA stuff i will take a look tomorrow and fix them.

NathanKell:
Ok, that makes sense! I'll go ahead and merge this once I finish testing (couple minutes)

NathanKell:
Merged into update branch.

PR #1710, opened 2017-07-16
[Bug Fix] Part max pressure tolerances

PhineasFreak:
**Change log:**

* Add a pre-RO pass to increase the pressure limits of all parts from 4 to 20 MPa (plenty enough to land on Venus and probe the lower atmospheres of the gas giants).
* Patch comments!

Hope that i did good by basing this PR against the RO-Pap branch...

### DIFF
--- a/GameData/RealismOverhaul/RealismOverhaul_Global_Config.cfg
+++ b/GameData/RealismOverhaul/RealismOverhaul_Global_Config.cfg
@@ -1,4 +1,8 @@
-// Temp fix: apply crew mass in VAB/SPH
+//  ==================================================
+//  Allow the wheel modules to be deployed even if they
+//  are shielded by other parts.
+//  ==================================================
+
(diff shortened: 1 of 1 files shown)


raidernick:
@PhineasFreak is this going to overwrite any of my parts that already have a gtolerance and maxpressure set? I have several probes that need extreme pressure and g tolerances way beyond what you have here to function properly.

PhineasFreak:
@raidernick This runs "BEFORE" any RO part patch is applied, so if your custom values apply "FOR" RO then you will override them.

I did something similar previously for the Coatl Aerospace Venera parts (example): https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/CoatlAerospace/RO_CA_ProbesPlus_Command.cfg#L506

(and now that i think about it...i should also add to them the missing custom pressure limits...)

PR #1712, opened 2017-07-16
HG-3 and HG-3-SL Plume Fixes

Ash19256:
Making the plumes actually work for more than one engine model, and making the sea-level plume noticeably different from the vacuum plume.


NathanKell:
I'm confused--there's both vac and SL configs on both the vac and SL parts?

Ash19256:
My response to the SHIP parts was to make it so that both of the SHIP HG-3 parts used the universal config. This had the effect of making it so that you could theoretically configure the "Vacuum" model as if it were a sea level variant and vice versa. As such, plume configs would logically be necessary.

GameData/RealismOverhaul/RealPlume_Configs/SXT/SXTHG-3.cfg
```diff
@@ -0,0 +1,42 @@
+@PART[SXTLT80]:FOR[RealPlume]:NEEDS[SmokeScreen]
```
PhineasFreak: Wrong part config name, should be "RO-HG-3".

Ash19256:
Well, that fixes that small issue, but unfortunately the SHIP plumes don't work for some reason. When in the game, the engines appear to default to what appears to be the SXT plume for some reason. I'm not sure if this is something that would have been fixed by correcting the part that the SXT plume directs to, but I'm not sure.

Ash19256:
Alright, as far as I can tell everything is working correctly. I'll do some testing and if nothing's broken (ie. engines using wrong plumes) I think this is ready to be merged.

PhineasFreak:
This is good for a merge.

PR #1725, opened 2017-07-20
RN fix soyuz descent pods

raidernick:
-I don't know what happened with this somehow the module got duplicated.
-Removed duplicate engines modules, restored solid fuel, and fixed
plume.
--Pod should autoburn at 7m off the ground if player has rnmodules.dll
somewhere in their install.


raidernick:
@NathanKell try this to fix the pods

PR #1731, opened 2017-07-21
RN fix soyuz pod landing engines staging

raidernick:
-Fix it to not activate when staged


raidernick:
@NathanKell ok while this does remove the engine from the staging, it also breaks the auto activation of the landing engines. I'm not sure why it's doing this. It turns out the pod has the RCS enabled by default(you don't need to stage it) and staging enabled for the engines so they actually are activated by the plugin. Once you disable the staging it breaks the plugin from working.

This needs to be looked into further before being merged.

raidernick:
ok @NathanKell I rewrote the way the plugin works, made some more config changes. To make these parts work correctly you will need this PR as well as the updated plugin for 1.2.2 from here: https://www.dropbox.com/s/upihkucunf0r6wt/rnmodules_1.1.5.0.zip?dl=0

Replace all instances of this plugin in your gamedata folder with this one.

raidernick:
ok gonna merge this @NathanKell so you can just pull to get it. You can download the fixed plugin above.

PR #1733, opened 2017-07-22
Add TestFlight config for RD-180 (and 170/191)

abortz:
Burn time estimated from Atlas user guide, plus allowance for documented 1 in 70 loss of performance in late burn.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD170_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD170_Config.cfg
@@ -220,3 +220,55 @@ PARTUPGRADE
 	
 	deleteme = 1
 }
+
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[RD-170]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
+	{
+		name = RD-170
+		ratedBurnTime = 129
+		ignitionReliabilityStart = 0.95
+		ignitionReliabilityEnd = 0.999
+		cycleReliabilityStart = 0.94
+		cycleReliabilityEnd = 0.995
+	}
+}
+
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[RD-171]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
+	{
+		name = RD-171
+		ratedBurnTime = 129
+		ignitionReliabilityStart = 0.95
+		ignitionReliabilityEnd = 0.999
+		cycleReliabilityStart = 0.94
+		cycleReliabilityEnd = 0.995
+	}
+}
+
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[RD-171M]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
+{
+	TESTFLIGHT
+	{
+		name = RD-171M
+		ratedBurnTime = 129
+		ignitionReliabilityStart = 0.95
+		ignitionReliabilityEnd = 0.999
+		cycleReliabilityStart = 0.94
+		cycleReliabilityEnd = 0.995
+	}
+}
+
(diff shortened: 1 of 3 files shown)


PhineasFreak:
Did you take into account the throttling capability? The Atlas V User's Guide has a wide range of possible BECO times (401, 521, HLV) and none of them match the 220 s figure (401: 242 s, 521: 252 s, HLV: 367 s).

abortz:
Yeah @PhineasFreak that's what I did.

I roughly integrated their graphs of launch throttle vs time (see pgs. 2-17 through 19, 2-21). I got very close results from all of them, around 226s. The 220s figure is by taking the one documented loss of performance near the end of a burn and figuring they are running slightly beyond ideal burn time. ;-)

Does that seem reasonable?

PhineasFreak:
If the TF failure tale can reach the HLV burn time then yes, it is OK. If not then the value should be increased to safely reach that rated burn time. If you can also incoporate the same values for the RD-170 and RD-191 engines then you will be golden!

For the record, the one Atlas V RD-180 failure was not a performance loss, it was a mixture ratio error. The engine avionics used a higher ratio and was left with excess RP-1 at BECO.

jwvanderbeck:
Burn time can be scaled by throttle by the way

abortz:
Oh, didn't know the RD-180 failure was external to the engine. Interesting! I'll bump the burn time back to 226s, which has the benefit of safely handling the HLV core burn as well.

Re RD-170/191 engines:
1. RD-170 series should probably have lower reliability and burn time, given earlier construction, placement in the tech tree, and history of failures and lower burn time as used in Zenit first stage. I'm thinking burn time of 129s as per http://www.spacelaunchreport.com/zenit.html
2. RD-191 series having same stats as RD-180 seem reasonable. Newer engines used in practice for slightly shorter burns (e.g. 215s 100% thrust) with great reliability record so far.

abortz:
I gave the RD-172-173 the stats of the RD-180 since it was the "baseline" for that engine (according to the config description at least)

SirKeplan:
Burn times for for the RD-170/171 are too low. every RD-170 had a 140-150s qualification firing before use. http://lpre.de/energomash/RD-170/index.htm
Also seems they burned slightly longer than that in flight(over 150s) http://www.russianspaceweb.com/energia.html though that is presumably with throttling.

i would personally be inclined to set the burn time higher still though, as this was designed from the beginning as a reusable engine that would be recovered, so must therefore be very durable.

abortz:
Hmm. According to http://lpre.de/energomash/RD-170/index.htm they only fired for 142s as part of the Energia LV block A, with substantial throttling. I'd also be inclined to rely more on data from the Zenit where there have been 83 launches vs Energia where there have been 2?

Of course I have no idea what I'm talking about here so open to suggestions. :-)

abortz:
Ping cc @pap1723, would like feedback and to get this in master so I don't have to maintain a local fork :-)

pap1723:
@SirKeplan is this one good to merge?

raidernick:
@PhineasFreak what is going on with this? Can I merge it? I don't use testflight so I don't want to just merge without knowing. I'm trying to merge the rest of the PRs that don't have any conflicts.

PhineasFreak:
To be honest, i have no idea if these burn times actually make any sense. All TF configs use the absolute rated burn time, even for engines that throttle. As an example, the RD-180 burn time would be 367s and not 220s as specified.

The user has also not been online since the date that this PR was made so i guess we cannot ask for clarifications. At this point it might be better to close it and either reopen it if he returns or create a new one.

raidernick:
Ok then closing it for now. @PhineasFreak can you look at any other PRs that you have expertise in or already chimed in on and post there if they are good to merge or not.

jwvanderbeck:
> All TF configs use the absolute rated burn time, even for engines that throttle.

Mainly because that has been the way it was done in the past, but TF does have the ability to scale burn time with throttle.

PhineasFreak:
@jwvanderbeck So if an engine is used mostly in the 80% throttle setting do we multiply the max burn time by a factor of 0.8? Or is it more complex than that?

jwvanderbeck:
By default the burntime is 1:1 with time.  1 second of flight time = 1 second of burntime.  But you can set the configs to scale time based on current throttle setting, so that at lower than 100% throttle it accrues slower and higher then 100% it accrues faster.

PR #1737, opened 2017-07-23
[Maintenance] AJ10-190 global engine config

PhineasFreak:
**Change log:**

* Add some default engine module patching.
* Add a generic TestFlight config.


pap1723:
I rolled these into the massive global Engine Config edits that are going on. We are good to go on this.

PR #1739, opened 2017-07-24
Fix TestFlight for RN's R-7 engines.

ahmedcharles:
This doesn't work yet, but I'm not sure what else could be going wrong.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_R-7_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_R-7_Engines.cfg
@@ -22,7 +22,7 @@
 	!RESOURCE[ElectricCharge]
 	{
 	}
-	
+
 	@MODULE[ModuleEngines*]
 	{
 		@minThrust = 930.28
@@ -57,7 +57,7 @@
             amount = 1
         }
 	}
-	
+
 	MODULE
 	{
 		name = ModuleEngineConfigs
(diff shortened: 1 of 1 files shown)


raidernick:
What exactly did you change here other than the order of a couple things? It doesn't look any different.

GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_R-7_Engines.cfg
```diff
@@ -853,171 +1011,13 @@
...
-		{
-			name = ElectricCharge
-			amount = 0.002
-		}
-	}
-	MODULE
-	{
-		name = ModuleEngineConfigs
-		type = ModuleEngines
-		configuration = Core/booster Verniers
```
ahmedcharles: This is a typo, should be Core_Verniers.

GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_R-7_Engines.cfg
```diff
@@ -843,7 +1001,7 @@
 {
 	TESTFLIGHT
 	{
-		name = RD-108A_14D21
+		name = RD-108A-14D21
```
ahmedcharles: This seems like a typo, though I'm not actually sure if it matters or not.

ahmedcharles:
I commented on the two actual changes that aren't order. As for the changed order of things, I'm not actually sure if it matters. It actually doesn't seem to make a difference but the test flight config for the vernier engines doesn't get properly applied and I thought changing the order might matter.

ahmedcharles:
This should work, though the TF configs don't share data with any of the other RD-107/8 engines. I'll probably work on that over the next few days and try to unify them much like RCS works.

This should be good enough for now though.

NathanKell:
We're fixing this by moving RN's stuff over to the global engine configs system \o/

So we can go ahead and shut this. Thanks for spurring us on to that! :)

PR #1757, opened 2017-08-20
RealScaleBooster Ariane V fix - VEB decouple node

Theysen:
- VEB currently works as interstage although it is being kept on the ESC-A
- fixed decouple node so the VEB stays attached to the upper stage and provide the fuel and guidance

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Ariane.cfg
@@ -499,6 +499,7 @@
     @MODULE[ModuleDecouple]
     {
         @ejectionForce = 100
+        @explosiveNodeID = bottom
     }
 
     !MODULE[ModuleRCS*],{}


PhineasFreak:
Thanks for catching that one! I have some more updates to do on the Ariane 5 in the near future, especially for it's intricate double attitude control system and to fix some stupid errors of mine.

Theysen:
Welp I should actually tell NecroBones about to fix in the core mod but that one single thing isn't worth a whole new release so I thought that we're patching them anyways.

Just tell me about it and I can see what I can fix and add. Thanks though :)

PhineasFreak:
I have made the changes, i'll just upload the changes to GitHub for you to pull and preview once i am sure that everything is OK. Mostly mass distributions (since the adapter is now part of the upper stage) along with some RCS updates for the roll control authority.

PhineasFreak:
OK, the updates are live: https://github.com/PhineasFreak/RealismOverhaul/commit/d1efa43495a67caa64601b4d81fb59162ea41592

Do note that the RCS combination may seem very weird at first but it is actually a very clever setup: the roll thrusters (Hydrazine) control the vehicle from the moment of the booster separation till the separation of the ESC-A. After that, the SCAR thrusters (GH and GOX) activate and take care of the 3 axis stage + payload attitude control (until payload separation and CCAM).

The only "problem" (realism) is that the user may need to manually disable the roll RCS thrusters after ESC-A separation. Nothing will break if they are not disabled though.

Do not be alarmed by the lack of RCS plumes for the roll thrusters. For some reason both the "emptyEngine" and "lp_rcs" models (just empty engine and RCS thrust transforms - no actual model attached) do not accept any kind of FX. The thrusters work perfectly (can be also checked via the PAW RCS info).

Theysen:
Great I'm loading up right now and will check it out. 
Since you also included this very, very difficult fix of mine I think I delete this here and you just PR your complete fixes, right? 

Sad they don't emit a plume but welp, thanks for implementing the complete correct setup of the thrusters and even the stages they start working!

PhineasFreak:
No problem at all! Do some checks though to validate the changes and ping me back. I may have missed something obvious, as it happens to me all the time...

Theysen:
Alright, I think I might have goofed my understanding of all this. Due to a bad source what I mistakenly took for the VEB (the interstage) doesn't decouple at the bottom. 
I mistook the part as the VEB (as it was earlier entitled like this) to not decouple from the ESC-A. This is of course wrong and can be safely reverted to decouple on the bottom. They do not carry deadweight of the interstage with them on the upper stage.

For the plumes of the RCS thrusters I get a FX for every thruster. Even when disabling the one pair you mentioned before.?

PhineasFreak:
Yes, i was the one that goofed in the first place. I though that the VEB was part of the EPC-E (to provide roll control - the ESC-A provides it's own thrusters for that) and interstage structure...

Now: the VEB is mounted on the top of the ESC-A (like the Saturn IU) and the interstage is mounted on the bottom of it and it is a part of it. I got these from the Ariane 5 user's manual (has a graphic indicating the various parts of the vehicle). Other infographics though (launch video coverage) do not show that. And using some common sense, as you point out, they would not carry the interstage with it (it may also mess with the thermal properties of the HM7 engine - it is radiatively cooled). God, so much confusion!

For the RCS FX: it may just be my bad computer that i am currently stuck with...

Theysen:
Yes the ariane launch manual shows that the ISS (and actually many many pictures) is part of the ESC-A assembly at least and gets attached under the actual stage and manual also refers to "part of ESC-A which could lead to the assumption that it stays attached).

Then there is this: **Pyrotechnical expanding tube at the top of the ISS**
and 4 ullage rockets - where the fat part would let me make the assumption it pushes away from the ESC-A after ullage? 

I can't dig up any sources on that specific part but common sense tells me to not carry it with you after sep and ullage. I try to dig up more and deeper!

PhineasFreak:
That would make even more sense. Sub-assemblies of a stage are usually stacked separately before being integrated together (horizontal or vertical stack integration - the Ariane 5 uses the second one) so being "part of the ESC-A" may mean that it is part of the ESC-A **stack** (with the EPC-E and the payload adapter + SYLDA + fairing being the rest of them) and not of the **stage**.

Another thing that we should be added (by cloning RSB parts): the ullage motors and the Hydrogen vent valve of the EPC-E (to spin the stage after separation - similar to the Soyuz boosters).

Theysen:
So the solution would be to give the interstage a double decoupler top and bottom where one decouples from the bottom stage, then the ullage happens and upon good HM-7B ignition the ISS decouples from the top and sets ESC-A free. 

Unfortunately NecroBones stopped development before he added the ES variant, too.

PhineasFreak:
Does the interstage carry the ullage motors or are they part of the ESC-A? I cannot find any more information about them in the webs.

BTW, having a double decoupler in a **single** part does not allow the user to stage each of them on **separate** staging events.

Theysen:
That's why I would put on decoupler into the EPC-E on the top node +1

Ariane Manual states that the ISS has 4 ullage motors (I'd suspect they come from Nammo as the separation motors for the EAP do).

PhineasFreak:
Oh correct, the EPC-E does not have an integrated engine module! Nice!

I could give the motors some generic stats for now until we can find the proper ones.

Theysen:
Whatever you think works better, really! I'd prefer a complete "manual" sequence to get the timings correct (yes, sometimes I take it that seriously..) but that is just me!

PhineasFreak:
Masses are currently wrong but they are based around the 19440 Kg mark (14900 Kg propellants, 165 Kg HM7, ~70 Kg Hydrazine). The inert ESC-A mass should then be 4.3 (and not the stated 4.87).

PR #1770, opened 2017-08-27
NTR fixing, and RD-108

Bornholio:
did something before that shifted some {}
noticed RD108 was missing module

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD108_RD118_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD108_RD118_Config.cfg
@@ -42,6 +42,14 @@
     }
 
     !MODULE[ModuleEngineConfigs],*{}
+    
+    MODULE
+    {
+        name = ModuleEngineConfigs
+        type = ModuleEngines
+        origMass = 1.19				// Needs correct origMass Value
+        configuration = RD-108-8D75		// Verify
+	modded = false
 
     CONFIG
         {
@@ -98,7 +106,7 @@
         CONFIG
         {
             name = RD-108-8D75PS
-			description = Used on Sputnik 8K71PS
+	    description = Used on Sputnik 8K71PS
             maxThrust = 918.3
             minThrust = 918.3
             massMult = 0.9774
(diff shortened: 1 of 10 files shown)


GameData/RealismOverhaul/Engine_Configs/RD108_RD118_Config.cfg
```diff
@@ -47,7 +47,7 @@
     {
         name = ModuleEngineConfigs
         type = ModuleEngines
-        origMass = 1.19				// Needs correct origMass Value
+        origMass = 1.278			// Needs verify origMass Value (value taken from VSR)
```
PhineasFreak: This mass value is invalid. The VSR RD-108 engine differs from the **base** global engine config since:

* It has integrated vernier engines (more mass).
* It has integrated structural fairing around the actual engine (more mass).
* The inert mass values of all the RD-108 engine variants are known from the sources listed at the top of the config file.

I do not know why this comment was there, probably a leftover...

Bornholio: I added those since i was trying to repair is without reference.  Right now its broken. (not able to be config'd)

PhineasFreak: You are missing a closing bracket at line No. 574: https://github.com/Bornholio/RealismOverhaul/blob/32ba1e8f21606af31dbd9256b3d541a0b654253c/GameData/RealismOverhaul/Engine_Configs/RD108_RD118_Config.cfg#L574

For some reason you reported that as an issue and it got removed: https://github.com/KSP-RO/RealismOverhaul/pull/1767/files. But if you take a closer look it is obvious that the bracket is actually missing (extra brackets are ignored by KSP).

I'd suggest Notepad++ as an editor. You can directly see where the brackets are opening and closing very easily.

Bornholio: Ah thanks, yeah that was me rushing a fix, and not noticing that at some point the whole module disappeared.  I don't normally use the Russian engines.  I'll see if something In paps spreadsheet doing ECM fixing is mangling the configs once I can talk with him.

PR #1771, opened 2017-08-28
RN_Soviet Probes

DocRockwell:
Request: 
-increase antenna range for Luna 1 and Luna 3 probes for lunar use. Current extended range is 1.9 Mm, suggest an increase to 4.0 Mm to match the communitron 16 antenna.

-decrease energy cost of antenna on Pioneer 5 probe from 0.15 to 0.0015, bringing it in line with Communitron 32. Currently the Pioneer 5's solar panels can't come close to keeping up with the power drain.


raidernick:
So what is going on with this PR? It has been open so long now that it is completely useless, look at all the files that are being changed now due to all the updates by other pr's. If this gets merged it's going to revert a ton of changes and cause massive problems. It should be closed and resubmitted.

blowfishpro:
@DocRockwell this should be rebased against the latest master/dev ... it can't be merged as it is

raidernick:

PR #1773, opened 2017-09-12
CommNet Antenna changes

smartdummies:
Changes to RN IS/Soviet proes, FASA antennas, Squad antennas, AIES
antennas, Coatl antennas, and BlackLeg antennas

### DIFF
--- a/GameData/RealismOverhaul/RO_DSN_Ranges.cfg
+++ b/GameData/RealismOverhaul/RO_DSN_Ranges.cfg
@@ -1,0 +1,7 @@
+@CUSTOMBARNKIT:AFTER{RealSolarSystem]
+{   
+      @TRACKING
+      {
+            @DSNRange = 200000000000, 2000000000000, 100000000000000
+      }
+}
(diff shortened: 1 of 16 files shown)


raidernick:
This needs to be redone. Every single one of my parts that has an antenna has a NEEDS[Remotetech] pass after the realism overhaul config pass. You made all the changes in the regular config instead of the remotetech config right under that. This is going to cause problems and I don't want anyone to merge this until that is fixed.

smartdummies:
@raidernick The changes for is for for stock CommNet and not for RT which is why I placed these in the initial pass. The antenna change for your parts follows other places such that the first pass updates the stock ModuleDataTransmitter module and then an NEEDS[RemoteTech] config removes the stock module and configures the ModuleRTAntenna module. If this is not correct then I can move my changes into a NEEDS[!RemoteTech] section. Let me know how this should be handled an I will make my changes accordingly

raidernick:
I would prefer they be in their own pass yes they should not be in the main pass

smartdummies:
I will make it so

smartdummies:
@raidernick check out these confgs and if they are correct I will apply these changes to the other files.

smartdummies:
Can somebody take a look at these and let me know if the changes are good so I can go on and apply the same style to the remaining file.

Thanks

NathanKell:
LGTM.

PR #1774, opened 2017-09-16
Tweak Gloster Meteor cockpit temp limits

aw1621107:
Change temp limits to match those of the other starting cockpits


NathanKell:
I'd like to keep the Meteor cockpit's skin max temp higher, since it's the only starting jet cockpit. But maybe that's silly gameplay?

aw1621107:
How much higher?

NathanKell:
I'd say 450 would be good, should let you get to Mach 1.5 or even 2.0 in steady state?

aw1621107:
Done!

PR #1799, opened 2018-01-27
[Bugfix] Update the FilterExtensions patch

PhineasFreak:
**Change log:**

* Update the MM patch for compatibility with the new FilterExtensions releases.


PhineasFreak:
Damn, i forgot that the PRs should point to the dev branch...

PR #1807, opened 2018-01-30
Correct AJ10-190 fuel ratio & SSTU Orion masses

ctiberious:
From what I can find, the AJ10-190 did not use MMH/MON3 for fuel.  It used MMH/NTO (N2O4).

I'm also finding that the version being used on the Orion MPCV has a higher thrust rating than the version used on the Space Shuttle.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
@@ -8,17 +8,18 @@
 
 //  Sources:
 
-//      Norbert Brügge - US Liquid Rocket Engines:                  http://www.b14643.de/Spacerockets/Diverse/U.S._Rocket_engines/engines.htm
-//      Alternate Wars - Aerojet General Space Engines:             http://www.alternatewars.com/BBOW/Space_Engines/Aerojet_Engines.htm
-//      NSTS Shuttle Reference Manual - ORBITAL MANEUVERING SYSTEM: http://science.ksc.nasa.gov/shuttle/technology/sts-newsref/sts-oms.html
+//  Norbert Brügge - US Liquid Rocket Engines:                  http://www.b14643.de/Spacerockets/Diverse/U.S._Rocket_engines/engines.htm
+//  Alternate Wars - Aerojet General Space Engines:             http://www.alternatewars.com/BBOW/Space_Engines/Aerojet_Engines.htm
+//  NSTS Shuttle Reference Manual - ORBITAL MANEUVERING SYSTEM: http://science.ksc.nasa.gov/shuttle/technology/sts-newsref/sts-oms.html
 //      Orion Service Module:                                       https://en.wikipedia.org/wiki/Orion_Service_Module
 //                                                                  http://oai.dtic.mil/oai/oai?verb=getRecord&metadataPrefix=html&identifier=ADA036741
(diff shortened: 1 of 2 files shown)


ctiberious:
I've also included a correction for the engine on the SSTU Orion SM so it'll use the same engine configuration as the AJ10-190.

PhineasFreak:
MON3 is exactly the same as NTO but it is classified as being less corrosive to metal parts. None of the modern (1980+) hypergolic propulsion systems use NTO for that reason. IIRC the STS used MMH/MON3 for both the OMS and the ACS thrusters.

Orion will also use MMH/MON3 for both the engine and it's ACS: https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20150023030.pdf (since it uses ATV-derived hardware).

ctiberious:
Interesting.  The sources and all the wikis I came across just referred to it as N2O4.  One thing I do notice is that, at least in KSP, the density of NTO and MON3 are not the same.  Since even your sources say the same 1.65:1 fuel mixture is used, at the very least the ratios should be altered slightly.  The latest changes return to MMH/MON3 but adjust the ratios so we keep the proper 1.65:1.  I've also added the two sources you cited.

PhineasFreak:
MONs are generally less dense than pure NTO. The density is set by the percentage of Nitric Oxide in NTO (NO: 0.0013402 g/cm3 - NTO: 1.44246 g/cm3), so they are generally going to be less dense.

Yep, i had originally set a 1.6 O/F ratio, since i was not sure about the correct ratio. Later, and from R-4D documentation, if found out that the 1.65 ratio is the expected value for such propulsion systems (makes sense from R&D, manufacturing and heritage points).

I am also sure that i have many more sources about the Orion MPCV propulsion systems buried in the CMES configs - the first link was copied from them...

ctiberious:
With the new 1.65:1 fuel ratio, it seems like a waste of space to have a 50/50 mix of MMH/MON3 on the Orion SM so I've slightly tweaked those number.

I can't find too many places that talk about the actual expected mass of the Orion CM, SM and LAS but I have found a couple and those indicated masses are a bit higher than what we currently have.  What I've found lists the Orion C/SM totaling at around 25.8t (10.4t CM + 15.4t SM), and the LAS at around 16000 pounds.  Normally I'd put these mass changes into a separate PR but I know the two would conflict so I've left everything in this one with an updated title.

pap1723:
@PhineasFreak @ctiberious 
This should be removed and not implemented, correct?

ctiberious:
@pap1723 That depends.  If you'd prefer that I submit this as two separate PRs then I can do that.  As I mentioned above, the only reason I am submitting it as a single request is that the SSTU Orion SM and the AJ10-190 both need the same engine modifications, and I can't submit a separate PR just for the Orion mass adjustments because it would conflict with the engine PR.  I can submit them as separate PRs and just assume that the engine changes are going to be implemented on the AJ10-190 engine config if that is preferred.

PhineasFreak:
I agree with @ctiberious. The changes are valid (and pretty much required) for the SSTU compatibility.

I would also like the updated global engine config so i can fix/update the CMES Orion parts (all of them are based on the old 1.6 ratio).

raidernick:
Is this good to merge?

raidernick:
@ctiberious please fix the conflicts and I will merge this

raidernick:
@ctiberious still has conflicts as the base branch has changed again with the new release of RO

ctiberious:
I can't seem to figure out how to get the conflicts to resolve in the existing PR.  I will create a new PR based on the latest files.

PR #1817, opened 2018-02-22
SSTU RCS

DocRockwell:
Updated for RP1


PhineasFreak:
It would be better if you used the global RCS config instead of redefining the RCS module for each of the parts (less clutter and better compatibility with the rest of the parts). [Example](https://github.com/KSP-RO/RealismOverhaul/blob/dev/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg#L144-L146).

[The global RCS config can also guide you through](https://github.com/KSP-RO/RealismOverhaul/blob/dev/GameData/RealismOverhaul/RO_SuggestedMods/RO_RCS_Config.cfg).

DocRockwell:
How does it look now?

PhineasFreak:
Looks OK. I will also tag @pap1723, since he did some work on SSTU.

raidernick:
Same question here as the previous PR about antennae. Why is this being done through a branch that conflicts with changes being made in master? Look at all the commits and files changed, this can't be merged like this...

raidernick:

PR #1818, opened 2018-02-22
Various SSTU tank fixes for 1.3

DocRockwell:


raidernick:
As I posted in the other 2 pr's this is not right, this pr should not be merging these branches.

raidernick:

PR #1832, opened 2018-03-20
Recompile for KSP 1.4.1

aw1621107:


raidernick:
Branch has conflicts now and KSP version already exceeds this. Should be resubmitted when we are actually updating RO to KSP 1.4.x or whatever verison it is on at the time. Closing

Issue #1819, opened 2018-02-23
X-405-H - adjustments and upgrades


blowfishpro:
Been doing a bit of research on the canceled X-405-H engine intended for the Vega upper stage of Atlas

Sources:
1) [Space Launch Report](http://www.spacelaunchreport.com/atlasnot.html) (Ed Kyle's site, I believe we've considered this to be reliable in the past)
2) [Alternate Wars](http://www.alternatewars.com/BBOW/Space_Engines/GE_Engines.htm) (Big Book of Warfare / Space Engine Encyclopedia)
3) [_History of Liquid Propellant Rocket Engines_](https://books.google.com/books?id=s1C9Oo2I4VYC&q=vega#v=snippet&q=vega&f=false) (book by George Paul Sutton)

Current stats in RO:
* Thrust: 156.3 kN
* Isp: 311s
* Ignitions: 2
* Burn time: 195s

1) Stats from Space Launch Report
* Engine Name: 405H-2
* Thrust: 156.3 kN
* Isp: 311s

2) Stats from Alternate Wars
* Engine Name: 405H
* Thrust: 150.5 kN
* Isp: 300s
* Expansion ratio: 25
* Burn time/ignitions: design conditions not specified, but was ground tested at 2000s and 12 ignitions

3) Stats from _History of Liquid Propellant Rocket Engines_
* Engine Name: 405H
* Thrust: 150.5 kN
* Isp: 300s
* Expansion ratio: 25
* Ignitions: 3 (using hypergolic ignition fluid)

There are definitely two sets of thrust/Isp numbers.  So this either means that one set is wrong, or that there was more than one variant.  The multiple variants hypothesis seems more plausible to me for 3 reasons:
* Both sets of numbers correspond to about the same fuel flow, ~51 kg/s (i.e. turbomachinery is the same with a different nozzle)
* Space Launch Report refers to the engine as 405H-2, which suggests a variant
* Pictures I can find do seem to show two nozzle variants: [#1](http://www.alternatewars.com/BBOW/Space_Engines/GE_Vega_Engine.png) and [#2](http://www.b14643.de/Spacerockets_2/United_States_4/Atlas-Vega/Gallery/Vega_1.jpg)

I think it makes sense to up the base ignitions from 2 to 3 given that (3) states specifically that this was the number of ignitions.

But would it be reasonable to have both variants in RO?  It seems like any changes happened before the planned first flight in 1960, but it does seem like both actually existed at some point in time.

[...]

PR #1836, opened 2018-03-25
Configure BDB X-405H

blowfishpro:
For just the one config, it doesn't seem worth using engineType = X405
and then having another patch to remove the unneeded configs and fix the
thrust.  If we ever end up having multiple X-405H configs per #1819 I
will implement it that way.


blowfishpro:
Worth noting that this requires SolverEngines v3.5 to work, but that's been out for almost a month now

PR #1837, opened 2018-03-31
DISALBED -> DISABLED

glilienthal:
I don't know what that lines does, but it looks very much like a typo.


raidernick:
thanks for finding this

PR #1843, opened 2018-04-01
[Mod Support] KTDU-425A global engine config

PhineasFreak:
**Change Log:**

* Add the KTDU-425A global engine config (moved from the CoatAerospace ProbesPlus KTDU-425A part config).


raidernick:
@PhineasFreak can this be merged into the 1.3.1 release despite being part of the changes for that other pr?

PhineasFreak:
Yes. It is an engine config so i guess that anyone that wants to use it can do so (not restricted to ProbesPlus only).

PR #1853, opened 2018-04-06
[Maintenance] RSB updates & fixes

PhineasFreak:
**Change Log:**

* Normalize all temperature values to account for the DRE nerfs.
* Fix the RCS thrusters of the Ariane 5 (they are part of the upper stage, not the interstage).
* Fix the inert mass of the first stage.
* Minor updates to the part descriptions.

**Notes:**

* The Ariane 5 interstage should also carry ullage/separation motors but these are not documented very well (if at all...). They also cannot be integrated with thr interstage, since KSP cannot handle two separate modules in one staging action in the staging list (only via action groups). So, the solution for now is to use either the generic RO or the RSB PSLV separation motors.

* Older discussion with @Theysen about the above note: https://github.com/KSP-RO/RealismOverhaul/pull/1757
* EADS Astrium -  Ariane 5 ACS: https://web.archive.org/web/20110719143310/http://cs.astrium.eads.net/sp/spacecraft-propulsion/showcase/ariane5-attitude-control-system.html
* Arianespace - ESC-A: http://www.arianespace.com/?popup=ariane-5-4

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Agena.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Agena.cfg
@@ -40,8 +40,8 @@
     @crashTolerance = 12
     @breakingForce = 250
     @breakingTorque = 250
-    @maxTemp = 473.15
-    %skinMaxTemp = 573.15
+    @maxTemp = 673.15
+    %skinMaxTemp = 773.15
 
     @MODULE[ModuleRCS*]
     {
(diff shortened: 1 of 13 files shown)


Theysen:
Ah yikes, the explosive node for the Ariane 5 Interstage Decoupler is set to bottom now, it needs to be on top as Ariane doesn't carry its interstage as dead weight with it once rid of the core. I think this is just a slight mistake, I'll PR the fix asap 😄

raidernick:
@Theysen that was my fault I only checked the syntax, didn't actually test the parts as @PhineasFreak does a good job with that

PhineasFreak:
Wait, didn't i fix that?

PR #1861, opened 2018-04-11
Smaller starting camera

Schnobs:
Rescaled starting camera to be more usable with sounding rockets / not require a V-2. Before 1.25m, now 0.675m (in line with bio sample container).


pap1723:
I disagree with this push. The camera is the size that it is, because the first images from space came from a V-2 sized (an actual V-2) rocket. The camera technology (mostly the film) needed that much shielding from what it was going to experience on launch and in the upper atmosphere / space.

Bornholio:
Agree with Pap1723  https://history.nasa.gov/SP-4401.pdf you need to include the payload structure in its size and looking at the V-2 that carried cameras while its a bit conic it ends in about a 1.5-1.6m diameter so the current on is compact enough. superstructure above included the recovery systems, so that is not forgotten.  https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19650006344.pdf This isn't just a 35mm camera strapped on, its a high rate (for the time) movie camera with mechanical stabilization and shock compensation.  Camera systems for reconnaissance get bigger not smaller.  ECC serves as a fine photo bucket in these later systems for recovery.

raidernick:
Since neither @pap1723 or @Bornholio think this is a good idea I'll take their word for it and close this.

PR #1865, opened 2018-04-19
redone Vanguard solids

Schnobs:
Based on a single but solid-looking source, new configs for the Vanguard third stage:
GCRC: bog-standard star-cut SRB in metal casing
Altair: Fiberglass case, more propellant, better ISP

In it's current form, this isn't as much a pull request as a request for help. How do I do this right?

I'm adding a whole new part, although both engines have virtually the same dimensions and somewhat comparable performance. However, their propellant loads are different and I can't figure out how to deal with this in an engine config.

Besides, I'm not sure about naming. What I have here are two dumb SRBs, which my source refers to as GCRC and ABL / X-248. Limited research suggests that the "Altair stage" was based off a X-248 but had 3-axis control. I didn't find anything I'd depend on, though, only brief and contradictory data. 

Still, it seems intriguing that most sites give different stats from what I have here -- even if they just copy off each other, these figures must have originated somewhere. Thus I wonder if it wouldn't be better to rename the vanilla SRB to "X-248", and leave "Altair" available for something more advanced.


PhineasFreak:
About propellant loading: it is impossible to have a global engine config referencing the same part and have different resource loadings. This is also the reason why we have different engine configs for the two STAR 37 SRMs.

PhineasFreak:
The Altair really was a later addition to the Vanguard launch vehicle (initial launches used this older GCRC motor to launch the early Vanguards, before the 20 inch versions came into view).

I'd like to at least request:

* separating the two global engine configs (these two motors are not the same). You can use the TF parameter **techTransfer** to transfer a percentage of the TF data between the GCRC and the Altair (this percentage should be a high value, at least 75% since only the casing was changed).
* moving the required propellant resources directly to the global engine config (so they will be available directly for other parts - propellants are also integral parts in SRMs).

For the rest i agree. In fact, it will make a fine addition for the early solid rocket motors (before Altair).

Schnobs:
I've now split off the GCRC into a separate config /RO/Engine_Configs/GCRC_Config.cfg, this step has also solved a few other issues I've been having. Engine now works as intended.

> moving the required propellant resources directly to the global engine config

Not sure if I understand? There's a new +PART[] for the GCRC specifically for the purpose of giving it a smaller amount of propellant. Both engines use PSPC. 

There's only one other part referencing the Altair config: RO_SuggestedMods/RaiderNick/RO_RN_Vanguard.cfg. However, it seems to override the RO configs, setting it's own propellant and thrust curve. I won't touch it unless @raidernick tells me to.

--

For the record / RP-0 integration:
The GCRC is basically a standard JATO rocket of the time. The specific dimensions may be unique, but items of similar size and performance were in wide use. It should be cheap to unlock and not provide much if any ECM to the Altair.

The Altair uses "plastic" casing (=resin+fiberglass) which was new and experimental.

Re: Testflight, I've given the GCRC a much higher cycleReliability than the Altair; it doesn't seem right that it carries over, but then again, AFAIK none have ever blown up, so why worry? Ignition reliabilty starts equally low for both, and, given that this most likely reflects integration errors, it should carry over.

PhineasFreak:
What i mean about "moving the resources" is to move the ModuleFuelTanks that defines the solid propellant from the unique part config inside the global engine config. Example: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/Star37_Config.cfg#L307-L326

Doing so will allow any future parts that require this engine config to also have the correct propellant loading. It is similar the the TEATEB resources defined inside many of the liquid propellant engines (considering that an SRM without propellant is as useless as a liquid engine without it's ignitors).

Definitely do not touch any RN configs! @raidernick will take care of them and move them over to the global engine config setup as he sees fit.

The carry over from GCRC is because, as you said, the mechanical differences are minimal and Altair is heavily derived from it. And it's true, the GCRC is probably the test version of the Altair but with more "conventional" manufacturing techniques/materials (IIRC none of them ever blew up on a mission but some of them failed due to second stage failures).

raidernick:
yep @PhineasFreak is right, the fuel amount should be in the engine config itself since it will always be the same on all parts that use it. 

Also yes, don't ever change my configs. I generally use the global configs now, but some of my parts work in such a way that using those configs breaks things, or adds more work than it's worth modifying them for a specific part than just manually inputting the specs directly into the RO config for it.

I have certain parts that the global configs just don't work at all with or break things. I do have some parts that probably COULD be moved, but I haven't gotten to it yet. I do slowly go through things as I update stuff and if I think it's ready to be moved over I do it. No need to worry about it yourself.

raidernick:
I think @PhineasFreak meant to separate it into it's own individual config file, not into it's own config inside the same file?

Schnobs:
yeah, something's wrong. The changes are fine, but there should also be a new file GCRC_Config.cfg which for whatever reason isn't part of the PR. No longer at home, will check tomorrow.

GameData/RealismOverhaul/Engine_Configs/GCRC_Config.cfg
```diff
@@ -0,0 +1,106 @@
...
+// given dV figures @ ISP only work out if end mass is "inert" mass +3lb or -8lb, respectively
+// hence drymass + propellant
+// 50 + 420 lb  -- 23 + 190 kg  109.1 l
+// 42 + 463 lb  -- 19 + 210 kg  120.65l
+
+
+@PART[*]:HAS[#engineType[GCRC]]:FOR[RealismOverhaulEngines]
+
+{
+	@title = GCRC
```
PhineasFreak: It would be better if these editor fields used the % modifier. It may sound weird and stupid but i have found a lot of mod parts that do not include these in their part configs.

GameData/RealismOverhaul/Engine_Configs/GCRC_Config.cfg
```diff
@@ -0,0 +1,106 @@
...
+				key = 0.91 1.052388 1.565168 1.565168
+				key = 0.9730649 1.160696 0 0
+				key = 1 0.93 -4.794868 -6.284284
+			}
+		}
+
+	}
+	!MODULE[ModuleGimbal] 
+
+	!MODULE[ModuleFuelTanks]
```
PhineasFreak: Both of the above fields are missing their closing ",*{}" statements.

Schnobs:
@PhineasFreak thanks for your patience.

raidernick:
looks good now

PR #1875, opened 2018-04-28
X-15 resources fix

Schnobs:
ServiceModule doesn't work in RP-0 (no food, oxygen, ...). Changing to Fuselage works around the issue.


raidernick:
yes it was done on purpose to prevent abusing the parts in early career, the cockpit should stay as it is in this case

Schnobs:
This PR fixes an obvious misconfiguration. No more, no less. As it is, the tank is simply not working at all. It might as well be not there.

PR #1876, opened 2018-04-28
unpatch DRE part upgrades

Schnobs:
A DRE patch ( GameData/DeadlyReentry/DeadlyReentry-Upgrades.cfg )  lowers the cockpit's heat tolerance to 450/450K, rendering it unsafe at any speed. Same patch would also add a part upgrade which doesn't work on the RP-0 tech tree.

This here cleans up the fallout.

On second thought, maybe it would be better to add a NOT[RealismOverhaul] to the original DRE patch?

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Spaceplanes.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Spaceplanes.cfg
@@ -1974,6 +1974,19 @@
 	}
 }
 
+@PART[Mark1Cockpit]:BEFORE[RealismOverhaul]
+{
+	//undoes DRE stock patches that conflict with RO
+	!MODULE[PartStatsUpgradeModule],* {}
+	@MODULE[ModuleAeroReentry]
+	{
+		!maxOperationalTemp = dummy
+		!skinMaxOperationalTemp = dummy
+		!UPGRADES,* {}
+	}
+}
+
+
 // Resized parts.
 // because why not, and 1.25m is too narrow a diameter for crewed spaceplanes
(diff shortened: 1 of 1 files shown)


raidernick:
This looks like it changes the same thing as the other PR? Can you remove the patch for the tank changes because as @pap1723 said it should be that way on purpose. Also, not sure about the heat tolerances for this part as I never use stock parts at all, @PhineasFreak is the heat tolerance even broken on this, or was it also done to prevent people from using it as an orbital vehicle in early game?

PhineasFreak:
The specific cockpit is definitely not intended for LEO reentries. We have specific (cloned) parts for LEO-rated spaceplanes, like the **RO-Mk1Cockpit**.

Schnobs:
(Sigh) let me outline the problem once more:
DRE applies patches to the Mark1Cockpit which override the RO config. That's override, not overwrite: the part stats still show the RO values. But in flight, the invisible DRE values are in effect.

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Spaceplanes.cfg
```diff
@@ -1696,7 +1696,7 @@
 		name = ModuleFuelTanks
 		volume = 100
 		basemass = -1
-		type = ServiceModule
+		type = Fuselage
```
raidernick: This is what I'm talking about. Why is this here, it changes the same thing the other pr does which is resource related for tacls. That pr is not going to be accepted anyway based on what pap said.

PhineasFreak:
If all parts are affected then if will be much easier and cleaner to just change the base DRE configs so that they will not apply when RO is installed.

PR #1884, opened 2018-05-21
Improved config for SABRE engines.

winged7:
Better isp and minimal thrust.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RO_B9.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RO_B9.cfg
@@ -3588,7 +3588,7 @@
 {
 	%RSSROConfig = True
 	@title = SABRE S
-	@mass = 2
+	@mass = 1.75
 	@MODULE[ModuleEngines*]:HAS[#engineID[AirBreathing]]
 	{
 		@name = ModuleEnginesAJEJet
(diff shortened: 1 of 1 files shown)


Bornholio:
https://en.wikipedia.org/wiki/SABRE_(rocket_engine)
smoke up the butt says your ISP is good.  IF specifications are derived from Wiki Sources TWR should be 14:1 with vacuum thrust of 2940kN

My comment on that is given the nozzle size i can't see that as possible unless thrust is far lower than specified. See RL-10's for ratio's that make sense (280:1 Nozzle = ISP 465). Or the chamber pressure is double or more an RS-25. RS-25 is ~69:1 ratio and only gets 452 ISP

winged7:
14:1 is air-breathing TWR which has nothing to do with rocket mode (btw. current airbreathing TWR is too low and we need to increase that number as well).

Regarding nozzles I've found only this: 
"Nozzle profiles similar to those of the Space Shuttle Main Engine are assumed"
https://www.nas.nasa.gov/assets/pdf/papers/Mehta_Unmeel_Skylon2015.pdf page 6

winged7:
Some info about mass of the engine:
https://web.archive.org/web/20140127114808/http://www.reactionengines.co.uk/tech_docs/JBIS_v57_22-32.pdf page 32
Mass of the engine, inlet, bypass is equal to 7400 kg and after subtracting the mass of inlet (because it works in RO as a separate part) it will equal to 7000 kg.

winged7:
This page states that vacuum thrust would be only around 1800 kN
https://info.aiaa.org/tac/PEG/HSABPTC/Public%20Documents/Skylon%20Spaceplane.pdf

PR #1887, opened 2018-05-24
SSTU MUS tank fairings to work with small tank sizes

Relic13:
Allows MUS fairings to be able to size correctly with very small tanks.  Old min defaults to 0.625, but tanks can go down to 0.1.


Bornholio:
Didn't see any problem, is there any weird math in the smaller region?

Relic13:
I'm not sure, but what I noticed was that if I was using a MUS tank that was very small (for probes, say 0.3m diameter), the MUS fairing diameter could not be resized below 0.625m.  Also if I had an upper stage and wanted to have the upper fairings be small for matching the kickprobe.

Basically:
You currently cannot resize the MUS fairings smaller than 0.625, but you can use tanks down to 0.1m.

I don't see why these tank fairings shouldn't be able to match at 0.1m.

Bornholio:
My SSTU intall has some issues but this seems to work fine

PR #1888, opened 2018-05-25
Fixing amount of electricity produced by Orion SM

winged7:
After applying the fix it produces as much electricity as stated here: http://www.esa.int/Our_Activities/Human_Spaceflight/Orion/Power


Bornholio:
Agree with the correct number being 11.2kW for the module (O-ESM).
https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20170009574.pdf page 5
"In  nominal  condition  the  four  SAW  can  supply  a  total  11.2  kW.  The maximum CMA/CM power demand is 7.3 kW, this figure being applicable even with one wing failed"
Is there something multiplying it by 4?  

https://i.imgur.com/22C2sHN.png

PhineasFreak:
SSTU probably counts all four of the sunCatcher transforms as valid, that's why production is four times larger than expected (the config specified the overall power generation and not the per array one).

PR #1894, opened 2018-06-10
Added Launchers Pack Configs

jeremy-penn:
I made config files for the [KK Launchers Pack](https://forum.kerbalspaceprogram.com/index.php?/topic/71323-143-launchers-pack-spacex-pack-v53-released-april-18th/) mod.

Things present in the mod I did not address:

- The ITS parts (sans the gridfin) because it does not appear like the parts are finished. I also do not know if the ITS ship is compatible with FAR. That would require some further testing.

- The Raptor engines because I couldn't find reliable data for them and, as they're connected with the ITS launch system, I ignored them.

- The dragon capsule. I have some code for its present state, but it's a WIP on the creators end. So, I left it commented out for now. I'll revisit it if/when the creator finishes the model.

Please feel free to ping me on discord (Nightcat) with feedback or questions.


PhineasFreak:
We already have configs for the Raptor engines if you want to also configure the parts: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/Raptor.cfg

jeremy-penn:
Okay, I used the existing Raptor configs as a basis to configure KK's raptor parts and the ITS booster stage. Thanks for the heads up, @PhineasFreak

PhineasFreak:
There is no need to recreate the global engine configs locally, that's why i linked you to them (you also lose all other engine variants). You just need to define a **%engineType = Raptor** parameter in the engine part and RO will take care of everything else.

Example: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg#L36

jeremy-penn:
Okay, I can do that. But will it work with KK's new models? I made my own because their raptors aren't single engines. They made clusters of raptors. Will it properly scale the part for multiple raptors in a single part?

raidernick:
I would like the files to be renamed to follow the naming convention I already have for the rest of the kk parts. It helps to keep things organized and easy to find. For instance, kk_engines.cfg tells me nothing about what's in the file, all of the engines?, some of the engines?, which rockets. Please fix to use the current naming system for this to be merged.

jeremy-penn:
I renamed the files to properly fit your naming convention. That was a very silly oversight on my part. Sorry!

jeremy-penn:
Update:

The mod now contains configs for all the raptor engines, using the existing global raptor configs (thank you @PhineasFreak for the help) & ITS Booster Stage fuel tank. The dragon capsule will remain unsupported until KK finishes it.

Finally, I tested the ITS Ship and it does, indeed, appear to be incompatible (really the flaps) with FAR. Therefore, I will not be supporting it.

Bornholio:
The best way to handle the ITS in my mind is have a patch that removes the aero from it if FAR and use separate parts, probably proc wings control surfaces or all moving wings to provide control.

PR #1902, opened 2018-07-06
SSTU SRB

DocRockwell:
Fixed configs for SSTU solid rocket motors. RealFuel engineConfig switching between variants doesn't seem work, so the ISP is fixed, but the thrust switches at least. I set the minimum thrust to ~50% of max thrust, instead of being locked at max, so the performance can be tuned in the VAB.
The GEM booster needs its plume position adjusted, it's currently above the nozzle. The thrust on the GEM boosters might also need some tweaking, the 1.05 m diameter GEM40 looks alright, but increasing the diameter for the GEM46/60 causes thrust to scale to higher values than the real versions.


DocRockwell:
- Got the RealFuel configs working
- changed the GEM variants from generic to GEM40/46/60/63 with ground and air lit variants (not sure if the non-gimbaling variants can be put to use with SSTU controlling the mass and gimbals, could give them a reduced cost though)
- Made RealPlume configs for the SRB's

DocRockwell:
The dry masses need some adjusting, they're to high. Tweaking the generic part mass and SSTU settings has no affect, I assume RealFuels is controlling the part mass. Is there a way to modify the mass with ModuleFuelTanks? @NathanKell

PhineasFreak:
Use **basemass = -1** to disable the RF tank mass management.

DocRockwell:
Can you use basemass = x * volume in ModuleFuelTanks, or does that only work for TANK_DEFINITIONS?

PhineasFreak:
Looks like it is possible: https://github.com/NathanKell/ModularFuelSystem/blob/master/ModularFuelTanks/TVPP_modularFuelTanks.cfg#L11

DocRockwell:
Hmmm basemass = x * volume didn't work when I tried it on these parts. Getting the volume and drymass dialed in for these modular parts is proving to be tricky.

DocRockwell:
Updates:
-split GEM40/46/60 into two separate parts, made them less customizable 
-GEM40/46: can no longer switch length, only 2 diameter options, added config for GEM40 GL/TVC
-GEM60/63/63XL: 2 diameter options, 3 length options corresponding to the SRB model. Volume and mass for these parts are good (mass set by engineconfig, but had to estimate GEM63 and 63XL mass from the dimensions of the boosters, no data in the Orbital ATK 2016 catalog, and the spacelaunchreport sight estimates seem spurious (63XL could not possibly have the same drymass as GEM46))
-UA120: pruned models that didn't match the length of UA1204,5,6,7 boosters. Engine configs restricted to the actual produced boosters
-still struggling with drymass of UA120 boosters
-RSRM still pretty customizable
-Renamed parts from SSTU-RO-ENG-SRBA, -SRBC, SRBD to SSTU-RO-ENG-RSRM, -UA120, -GEM40, -GEM60. Seeing as these parts haven't been working for a while, I figured now is a good time to give them more sensible part names. Also updated the RO Tree parts spreadsheet to reflect this (will make PR later).

DocRockwell:
The RSRM boosters might be nice to keep fully customizable, especially if part tooling can be added. I think engineconfigs will have to be removed to let SSTU handle thrust scaling if diameter is to be adjustable though.

raidernick:
This needs to be redone with no conflicts, as it stands it can't be put into the latest RO release since it's been sitting too long. Reopen it when ready.

PR #1903, opened 2018-07-06
SSTU Engine Cluster Mass Fix

DocRockwell:
Fixed the 2n-1 clustering mass bug by ignoring origMass and setting the stock mass parameter to the appropriate value. Mass switching for engines with variants configurable in the GUI is no longer possible on these engines. Added origMass back for the RD-107 and RD-108 as I personally never cluster these engines and would rather have mass switching with tech upgrades.


MikeOnTea:
Can you test whether the following line/patch in RO_SuggestedMods/SSTU/SSTU_Engines_GLOBAL.cfg can also fix the problem? I think it did for me and it should thus allow variants and clustering:

    @MODULE[SSTUModularEngineCluster]
    {
        @diameterIncrement = 0.1    // Allow increments of 1 cm
+       %adjustMass = false         // Fix mass calculation for clusters
    }

PR #1912, opened 2018-07-10
[Bug Fix] LMAE global config updates

PhineasFreak:
**Change Log:**

* Remove the part mass parameter since it breaks the **ignoreMass** global engine config operator.
* Add some default engine module parameters patching.
* Make sure that no previous engine module configs exist before applying ours.
* Make sure that all alternator and gimbal modules will be removed from the part config.
* Convert spaces into tabs.


Bornholio:
Looks good, didn't crash anything

PR #1920, opened 2018-09-02
Use global configs for Castor 1

Tidal-Stream:
This unifies all Castor 1 variants to use the global configs. It also moves the mass, max temperature, and solid fuel to the global config, so individual configs have to provide just the model.


PhineasFreak:
Two requests:

* Remove the **mass** parameter since it breaks some global engine config calculations (not really applicable for SRMs but just to be on the same page).
* Reduce the maximum temperature to something saner, like 773.15 K (500 °C). 1800 K (1526 °C) is far too high for a rocket motor.

Tidal-Stream:
* I will remove the mass although SRMs having fuel are already different than liquid engines.
* I picked the temperature from RealFuels since it seemed to be the most up-to-date config (took the fuel type and load from there as well). I will change it and check if it overheats. I assume this is external and not internal temperature because in one of my tests of Procedural solid rocket motors they were at [over 3,000 K][1].

[1]: https://user-images.githubusercontent.com/34808675/44946948-60b43c80-ae0e-11e8-9e7d-3e66fd158e80.png

raidernick:
do a += description on my part to leave in the directions about how to radially place these and to remove the fins as this is important for proper use with my rocket parts.

Tidal-Stream:
I fixed the issues pointed out. I am sorry for the two commits, I erroneously pushed the first before proper testing. When you merge the pull request, use the squash option and leave their descriptions out.

raidernick:
I'll fix this stuff on my end for my parts make a new pr for the others

PR #1923, opened 2018-09-09
[Mod Support] STAR 24C global engine config

PhineasFreak:
**Change Log:**

* Add a global engine config for the STAR 24C solid rocket motor.


Bornholio:
Why the massive thrust curve?  Looks good though.

PhineasFreak:
Excuse me? What's different on the thrust curve when compared against the rest of the SRM global configs?

PR #1925, opened 2018-09-23
RN fix castor and crafts

raidernick:
-Fix craft file for scale fix of castor 1


MikeOnTea:
Tested it, scale looks good to me now.

PR #1930, opened 2018-09-26
Fix SSTU engine cluster masses

MikeOnTea:
SSTU's engine clusters calculate their masses wrong. For example, if you change from one engine to two, the mass triples instead of doubles iirc. Fix this by setting adjustMass to false, which i think was made for realfuels compatibility.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Engines_GLOBAL.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SSTU/SSTU_Engines_GLOBAL.cfg
@@ -12,6 +12,7 @@
 	@MODULE[SSTUModularEngineCluster]
 	{
 		@diameterIncrement = 0.1	// Allow increments of 1 cm
+		%adjustMass = false			// Fix mass calculation for clusters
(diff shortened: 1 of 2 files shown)


stratochief66:
RealFuels is a requirement for RO though.

MikeOnTea:
Sure - i was trying to say if it was made for us, and if it works, we should start using it.

Bornholio:
diameter change worked okay, whatever is messed up in my install didn't seem like the =false gave a reasonable result, not sure if it was RF interaction

blowfishpro:
I don’t really know much about this area unfortunately

MikeOnTea:
adjustMass = false disables SSTU's engine cluster mass calculations, because apparently RealFuels has its own calculation, and they don't play nice along each other. SSTU passes the number of engines in a cluster to RF, so RF can and does calculate the correct engine cluster mass just fine on its own.

Bornholio:
tested a bit more and mike is right, seems to be appropriate

raidernick:
So is this ready to merge or not I want to release the new RO and I'll leave this out if no responses?

PR #1934, opened 2018-10-14
CO₂ scrubbers

ec429:
Our MM patches for CO₂ scrubbers using LithiumHydroxide were broken in various ways.
Note that I haven't tested patch 3 (c97b09c), because it's mildly game-breaking: previously any crew cabin with one of these scrubbers would run a water surplus, so you didn't need to take much water with you; now you get much less water (about a tenth of what a 'naut drinks) and it's not potable.
Also, that patch was created with a horrendous `grep`/`tr`/`sed`/`bash` script; hopefully it's correct.

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/TACLS/RO_TACLS_Tanks.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/TACLS/RO_TACLS_Tanks.cfg
@@ -89,23 +89,25 @@
 	@description = A life support module that filters Oxygen out of the air. Requires IntakeAir and oxygen atmosphere. Rated for 3 people.
 	@MODULE[TacGenericConverter]
 	{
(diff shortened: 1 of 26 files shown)


MikeOnTea:
I have not tested it, but i get to roughly the same values so calculations seem to be correct (except that i think your mol values in the commit message are off by a factor of 1/10 :))
Everything else also looks good to me.

ec429:
> except that i think your mol values in the commit message are off by a factor of 1/10

That's as a result of *two* errors that *nearly* cancelled out... I stripped a couple of leading zeroes off all the densities while calculating, because reasons (involving `dc`), and then forgot to allow for the fact that KSP measures mass in tonnes rather than kg.

Ah well, it doesn't matter.

raidernick:
Can this be merged did anyone else test it?

pap1723:
The conservation of math has been confirmed.

PR #1936, opened 2018-10-15
[Maintenance] DECQ R-7 Pack

PhineasFreak:
**Change Log:**

* Added RO configs for the missing parts (Sputnik 1 and launch gantry).
* Fixed the separation issues of the Soyuz FG fairings.
* Added a missing bottom attachment node to the RD-0124 engine (this also fixes an old bug).
* Fixed the attachment node sizes of the interstage truss adapter.
* Added command capabilities and increased the carried battery size of the R-7 core for compatibility for the Sputnik launch vehicle configuration.
* Reduced a bit the thrust and increased the VAC Isp of the RD-107/RD-108 verniers (sources define thrust values anywhere between 24 kN - probably an ASL value - to 34 kN and VAC Isp of 313 s).
*  Fixed some major bugs regarding the Syntin variants of the RD-107/RD-108 engines (previously it was impossible to use them since the vernier engines lacked compatible engine module configs).
* Added support for the stock CommNet system.
* Tweaked the heat production values of the engines.
* Added some extra part tags.
* Tweaked the part temperature values to be in-line with the other parts.

**Notes:**

One unfortunate note of the new DECQ R-7 core model is that for some reason the RD-108 vernier engines got separated from each other, resulting in one engine module for the X axis and one for the Y (i really wonder why this change was made, it makes no actual sense).

This is not much by itself until someone tries to use the Syntin variants of the RD-108 engine: due to a (probable) bug within RealFuels one vernier pair gets the correct propellant combination but the other does not (remains at KeroLOX). The result: a stage that is impossible to control after booster separation, since ony one pair of verniers work.

I hope that someone can confirm that this is a bug caused by the RO configs and not RF!

**This pack also requires some updates to it's RealPlume configs. Until PR https://github.com/KSP-RO/RealismOverhaul/pull/1935 can be merged then this PR should not be merged into master.**

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/RO_DECQ_R7.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/RO_DECQ_R7.cfg
@@ -276,12 +276,6 @@
             pressureFed = False
             ignitions = 1
 
-            IGNITOR_RESOURCE
-            {
-                name = ElectricCharge
-                amount = 0.25
-            }
-
             PROPELLANT
             {
                 name = HTPB
--- a/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/RO_DECQ_R7.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/DECQ/RO_DECQ_R7.cfg
@@ -1,19 +1,189 @@
 //  ==================================================
 //  Sources:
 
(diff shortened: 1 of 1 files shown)


raidernick:
@PhineasFreak I want to include your open PRs with the 1.5.1 RO release. Can all of them be merged yet?

PhineasFreak:
Yes but do note that this specific PR should be merged after #1935, as i cannot push the updated RealPlume configs that are required by it.

PR #1941, opened 2018-11-03
RO v13 for KSP 1.5.1

raidernick:
-Updating RO stuff for ksp 1.5.1


Bornholio:
Is there a way to do an _OR_ on the FAR dependency?

raidernick:
I don't think ckan allows that. You can tell it not to install one if the other is installed but I don't think you can tell it to pick a specific one over the other. If they are set as "needs" then it will try to install both, but fail since they conflict.

pianojosh:
@raidernick I'm not 100% on how it all works, but can RO depend on some meta requirement like "FerramAerospace", and have both "FerramAerospaceResearch" and "FerramAerospaceResearchContinued" provide "FerramAerospace", similar to how RSSTextures are provided in multiple resolution options?

raidernick:
I have no clue, I know next to nothing about ckan. You'd need to ask on their git.

raidernick:
@Olympic1 you'd need to show exactly how to do that in the RO ckan file, I have no idea how to do it and don't want to mess with it and potentially break it.

jrodrigv:
Hi guys, just wondering. Which is the "Hold on" reason? Are you waiting for a dependency to be released?

raidernick:
Note this is missing change to checker.cs for 1.5 version and this PR will likely be closed anyway as we are skipping 1.5 but is on hold for now.

PR #1946, opened 2018-11-24
RCS mass fix and rebalance

### LINKED ISSUE #1769: Broken 2 kN thruster in DEV BRANCH
Some changes seems to have broken most of the configs for the generic 2kN thruster and caused incorrect or no thrust. Awaiting further investigation.

MikeOnTea:
PR #1842 broke the rcs mass calculation in an attempt to fix #1769, resulting in single nozzle thrusters having a mass of 1/8 of a quad thruster with the same thrust, for example. The real issue was just poor balancing though, the calculation itself worked as intended before the patch.

Here's my attempt to fix this. Doing some research for balancing, it seems rcs thruster mass scales roughly proportional to sqrt(thrust) instead of linearly with thrust (see [here](https://www.rocket.com/files/aerojet/documents/Capabilities/PDFs/Bipropellant%20Data%20Sheets.pdf), [here](http://www.rocket.com/files/aerojet/documents/Capabilities/PDFs/Monopropellant%20Data%20Sheets.pdf) or [here](http://www.space-propulsion.com/brochures/hydrazine-thrusters/hydrazine-thrusters.pdf)), so this is what i did as well. 
I also reduced the mass savings for multi nozzle thrusters, according to one real world example i found [here](http://www.rocket.com/files/aerojet/documents/Capabilities/PDFs/Monopropellant%20Data%20Sheets.pdf).

These changes result in the following masses (for TL2 thrusters):

Thruster|Mass
---------|-----------------
**28/45 N**|1.7 kg
**138/223 N**|3.8 kg
**138/223 N quad**|11.3 kg
**275/445 N quad**|16 kg
**550/890 N**|7.5 kg
**1.1/1.78 kN**|10.7 kg
**2.2/3.6 kN**|15.1 kg

I also fixed and changed the scaling of the parts slightly (for rcs parts) or not so slightly (for the 1/2 kN thrusters, they were a bit on the small side i think, the latter now got a nozzle diameter of 45 cm, a little larger than the slighly lower I<sub>sp</sub> AR R-40B).

Edit: As siimav suggested i now changed the 1/2 kn thrusters' models to OMS Large if VSR is installed, those look better for a vacuum thruster i think:

image

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RCS_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RCS_Config.cfg
@@ -13,7 +13,10 @@
 		configuration = Nitrogen
 		modded = false
 		
-		origMass = 0.028
+		// The Apollo R-4D thruster was about 3.4 kg - 3.6kg [1, 2], i estimate the quad's mass to be roughly 16 kg
+		// 1. https://www.rocket.com/files/aerojet/documents/Capabilities/PDFs/Bipropellant%20Data%20Sheets.pdf
+		// 2. https://www.apolloartifacts.com/2013/11/marquardt-r-4d-apollo-spacecraft-attitude-control-engine.html
+		origMass = 0.016
 
 		CONFIG
 		{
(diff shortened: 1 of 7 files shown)


pap1723:
This is a good fix, HOWEVER be aware that these changes can be craft breaking as the sizes are pretty large for the single Thruster engines. These are a lot more accurate.

siimav:
I have tested these changes and the new balancing looks good.

PR #1947, opened 2018-11-30
Fix gimbal on BDB LR79

lpgagnon:
Don't delete a non-existent "second" ModuleGimbal, it ends up deleting the first. Also don't delete the non-existent ModuleEnginesFX.

BDB has 2 engines mapped to LR79: juno_engines3d and thorEngine. thorEngine has built-in verniers, so it has 2 ModuleGimbals instead of 1. The RO config strips out the verniers and their gimbal setting. The RO config for juno_engines3d tries to strip out the same verniers, but there aren't any. I don't speak ModuleManager, but it would seem `!MODULE[ModuleGimbal],1` is intended to match "the second one", but that "second of one item" is that one item.


pap1723:
I have noticed this issue as well with the LR79 from BDB. It causes some difficult ascents.

PR #1949, opened 2018-12-01
Add Speculative NK-9 Configurations

Tidal-Stream:
Add NK-9 upgrades in line with NK-15 and NK-33 with respect to TWR and
ISP. It should give players more options for light Soviet launchers.


Tidal-Stream:
ISP is taken directly from the corresponding engine - NK-15, NK-33, or AJ26-62. For the ones with increased thrust, TWR is taken from NK-33 and AJ26-62. NK-33's and AJ26-62's mass is 1459 = 1222 * 1.1937. NK-33's TWR is 1766 / 1459. We multiply this by 360 (the mass) and get 436. AJ26-62's TWR is 1815 / 1459, we multiply by 360 and get 448.

PR #1950, opened 2018-12-06
[Bugfix] KSP 1.5.X ModuleManager compatibility

PhineasFreak:
**Change Log:**

* Fix two ModuleManager errors that appear when using the newer ModuleManager releases (KSP 1.5+).


Bornholio:
Saw this with MM 3.1.1 but not 3.1.2  thanks.

PhineasFreak:
Weird, i saw these errors even though i had the latest MM assembly (v3.1.2). But anyway, it does not really matter as these are typographical errors and should be fixed regardless of the MM version.

pap1723:
I am surprised that those didn't cause errors before.

PR #1953, opened 2018-12-09
[Maintenance] J-2T gimbal module

PhineasFreak:
**Change Log:**

* Update the J-2T engine to use a better way for patching it's gimbal module (reference: https://github.com/KSP-RO/RealismOverhaul/commit/4e46064248c40152651b3b85491d0ed0443ed841)


pap1723:
This is the better way to fix this issue, thanks @PhineasFreak

PR #1961, opened 2018-12-23
BDB fixes

raidernick:
-Remove pointless NEEDS for BDB as these won't work without the parts
installed anyway, and they are causing mm warnings.


Bornholio:
awesome, thanks every little warning that goes away makes me happy

PR #1988, opened 2019-02-05
[Bugfix] DECQ R-7 vernier engines

PhineasFreak:
**Change Log:**

* Fix the TestFlight integration of the RD-0110, RD-108 and RD-107 vernier engines (they will now fail if the main engine fails).
* Fix the propellant switching between RG-1 and Syntin breaking the RD-108 and RD-107 vernier engines.

**Notes:**

* The vernier plume config for the RD-108 engine will not work correctly due to a patching issue within RealPlume.


pap1723:
@PhineasFreak what is your self-imposed rule for line length that you like to use for comments?

PhineasFreak:
Really, nothing specific. I always want (and try) to fit the text inside the 50 char limit of the part patch headers but other than that...whatever fits the purpose/feel.

(I do think that the formatting should be locked down though. Currently, the "Sources" header is the worse as the hyperlinks are long and you cannot really LF them).

PR #1989, opened 2019-02-05
[Bugfix] RealEngines updates

PhineasFreak:
**Change Log:**

* Change **all** internal part names to fix compatibility issues with RP-0/1. **WARNING: save & craft-breaking update!**
* Update all engines with integrated tubopump exhausts/vernier engines to use a single engine module in order to fix various TestFlight interop bugs.
* Simplify the RealPlume patches of the A-7 engine.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_American.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealEngines/RO_RealEngines_American.cfg
@@ -1,41 +1,38 @@
 //  ==================================================
 //  Sources:
-
-//  NASA - Vanguard: A History (chapter 5):         https://history.nasa.gov/SP-4202/chapter5.html
-//  Alternate Wars - Aerojet General Space Engines: http://www.alternatewars.com/BBOW/Space_Engines/Aerojet_Engines.htm
-//  Designation Systems - Martin Vanguard:          http://www.designation-systems.net/dusrm/app4/vanguard.html
-//  Norbert Brügge - Vanguard SLV:                  http://www.b14643.de/Spacerockets_2/United_States_7/Vanguard/Description/Frame.htm
-//  Norbert Brügge - Thor Able SLV:                 http://www.b14643.de/Spacerockets_2/United_States_4/Thor/Description/Frame.htm
-//  Encyclopedia Astronautica - AJ10-40 engine:     http://www.astronautix.com/a/aj10-40.html
-//  Encyclopedia Astronautica - AJ10-104 engine:    http://www.astronautix.com/a/aj10-104.html
-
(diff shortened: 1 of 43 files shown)


pap1723:
@PhineasFreak are you of the opinion that I can remove the duplicates I created in RP-1 for the old parts to save people's crafts? Should we just go with a craft breaking switch like this?

PhineasFreak:
Three considerations that i had in mind before pulling this trigger:

* The craft-breaking changes that will occur anyway for RO in the future to be fully compatible with KSP 1.4+.
* The fact that most parts that RP-1 requires (like the avionics and the custom tanks) are not available under RO, breaking crafts anyway.
* The fact that RP-1 still breaks careers/craft designs almost daily due to other reasons.

Now, part name changing will occur only once in the lifetime of the parts, with or without RP-1 installed.

siimav:
I noticed that the new part names are different to the ones currently used in RP-0. [See here.](https://github.com/KSP-RO/RP-0/blob/master/GameData/RP-0/Parts/RealEnginesFixes.cfg) Wouldn't it be better to use already existing ones so that one less round of craft-breaking updates would happen?

PhineasFreak:
Craft compatibility is already broken, both by the KSP version and the recent RO solids updates (see my notes above). So, i decided to use names that are a better fit (see: various KDTU part clones).

BTW, not all part names break RP-1 but i went and changed them anyway.

pap1723:
Reviewed, waiting to push changes to group together some craft breaking ones at the same time.

pap1723:
I will also need to change all of the RP-1 files to match these.

PhineasFreak:
So, i guess that this PR is now useless and not going to offer anthing, right?

pap1723:
Not necessarily, why?

PhineasFreak:
All these are part of/will be replaced by RO Engines, no?

pap1723:
Oh, yes they will be. Don't know if we want to remove them, I will listen to ideas. ROEngines will not be mandatory.

PhineasFreak:
Then what's the point over having a minum mandatory mod list (as the current situation is)? I though that RO Engines would **replace** all these heavy mods with a more streamlined, 100% RO-curated solution.

pap1723:
That is exactly what it is going to do. It is just not forced on players to download. If players don't want the whole thing, they can just get RealEngines for example. (I do not know why anyone would not want it, but I don't like forcing people to only go one path)

raidernick:
@PhineasFreak RO is not a parts mod and never will be. It can support official RO "approved" packs as replacers but I am NOT adding in any models or meshes to this mod(or approving of anyone else doing it either). The whole point is to keep RO as small as possible with as little overhead as possible. @pap1723 is correct it will be suggested only and not required or packaged or even linked directly as part of the RO download, it will just be listed as recommended along with the rest of the mods that are but not required.

PR #1993, opened 2019-02-18
Add basic kerolox plume for S2.253 engine 

Zantigar:
We have a new engine from the Scud missile (see 5730a6c7f17b55aa1769a61fc5379fe31fb73d57), but it was without a plume.  This adds a basic plume using the kerolox prefab, sized and placed to fit the temporary (RD-10x) model.


NathanKell:
It's kero and acid, right? Not sure if hypergolic or kerolox is more appropriate. Merging so we have *something* tho.

PR #1994, opened 2019-02-20
[Bugfix] LMAE/LMDE/LR89 engine masses

PhineasFreak:
**Change Log:**

* Fix the incorrect mass definitions of the LMAE, LMDE and LR89 global engine configs (introduced by https://github.com/KSP-RO/RealismOverhaul/commit/117754e48dfb8d6b9ffff28e1ea6bf7ca9680ed3 - breaks the custom mass overrides of specific parts like the DECQ and SXT Lunar Modules).


pap1723:
Is it a better idea to remove the global mass amount or to just add an ignore mass to those parts? Or to add offset mass to those parts? 

What is the advantage to removing them here vs setting those other values I mentioned above? What is the disadvantage?

PR #1997, opened 2019-02-25
Rebalanced nuclear reactors to their real life counterparts

thesciencenut:


thesciencenut:
Adjusted figures to better fit the following:

///Units are Kelvin(k), kilograms(kg), metric tons(mT), kilowatt thermal(kWt), killowatt electric(kWe), years and days (y,d)

///Sources
///https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/20020066158.pdf
///https://www.world-nuclear.org/information-library/non-power-nuclear-applications/transport/nuclear-reactors-for-space.aspx
Reactor Name	=	NASA Safe-400
Nominal Temp(k)	= 	1154k
Critical Temp(k)	= 1154k		///Initial Guess
Fuel type	=	97% Enriched UN
Fuel Mass(kg)	=	130kg
Reactor Mass(mT)	=	0.512mT
Reactor Heat(kWt)	=	400kWt
Reactor Power(kWe)	=	100kWe
Run Time(y,d)	=	10 years

///Sources
///https://en.wikipedia.org/wiki/TOPAZ_nuclear_reactor
///http://enacademic.com/dic.nsf/enwiki/5757685
Reactor Name	=	Topaz-2
Nominal Temp(k)	= 	824k
Critical Temp(k)	= 	1574k		///Initial Guess, also the original RO configs number adjusted to Kelvin
Fuel type	=	96% Enriched UO2
Fuel Mass(kg)	=	12kg
Reactor Mass(mT)	=	0.320mT
Reactor Heat(kWt)	=	135kWt
Reactor Power(kWe)	=	5.5kWe
Run Time(y,d)	=	3-5y		///Supposedly on the same fuel mass at full load... Needs to be re-examined

///Sources
///https://www.nuscalepower.com/technology/technology-overview
///https://en.wikipedia.org/wiki/NuScale_Power
Reactor Name	=	NuScale
Nominal Temp(k)	= 794k		//Initial guess, original RO config number adjusted to Kelvin
Critical Temp(k)	= 	2474		///Initial guess, original RO config number adjusted to	Kelvin
Fuel type	=	4.95% Enriched Uranium of Unknown type
Fuel Mass(kg)	=	140000kg		///Initial Guess, subtracted 450mT(Original config weight) from actual mass of 590mT
Reactor Mass(mT)	=	450mT		///Unknown actual value, but total weight is 590mT which I guess includes fuel mass
Reactor Heat(kWt)	=	200000kWt
Reactor Power(kWe)	=	60000kWe
Run Time(y,d)	=	2y

///Sources
///https://www.ipen.br/biblioteca/cd/genes4/2003/papers/1039-final.pdf
///https://en.wikipedia.org/wiki/RAPID-L
Reactor Name	=	USDOE RAPID-L Nuclear Reactor
Nominal Temp(k)	=	654k
Critical Temp(k)	=	2374k
Fuel type	=	40% and 50% Enriched UN
Fuel Mass(kg)	=	1000kg		//unknown fuel mass, initial guess (subtracted from estimated full mass)
Reactor Mass(mT)	=	6.6mT		///Unknown full actual mass, but estimated at 7.6mT, only specified as less than 10mT
Reactor Heat(kWt)	=	5000kWt
Reactor Power(kWe)	=	200kWe
Run Time(y,d)	=	10y		///at 80% of full load, or 7.5y at 100% load

Bornholio:
I'd have to look at each, US NRC and Canadian NSC are good sources but take some digging.

I've a good hatch report for CNSC that includes the IAEA summary of Nu-Scale http://www.nuclearsafety.gc.ca/eng/pdfs/RSP-0299-eng.pdf 
Its a 37x  17x17 264 rod Western PWR fuel assembly these weigh about half a ton. less than half of this is UO2 
so my estimate is U235O2 = ~462kg and ~7800kg U238O2 (~9.25tonnes eU
Its shutdown temps should be pretty low given its fail intrinsically safe values are supposedly air cooling after boiling off its coolant tank. How you tune that in NFE i've no idea.

The hatch report is bit out of date since i'm sure NRC got a submittal for 60MWe

I've not looked into the others deeply.  Maybe @Starwaster (look at Real ISRU)

thesciencenut:
@Bornholio Do you happen to have a rough estimate for the temperatures? I have no idea what makes sense. 1200k for critical?

Bornholio:
Design pressure of the vessel is 12.8MPa, thats ~330C steam temp, probably some Fos safety ~3x or 36MPA gives off the chart for steam tables, so lets just say ~700F or~ 370C (~ 640K)
This is why i don't understand the critical temps of these, you are past operation by a lot, assuming this is when the thing turns to slag.

Starwaster:
I just mentioned on Discord but I'll comment here too. Real ISRU has KiloPower 10 and KiloPower 40 implemented. I'm looking at maybe doing MegaPower but those would be for bases and would require reactor masses in the 10s of tons (like 10-40 tons for a 1-2 MW reactor) .... which for RO is going to be tough to transport and land. I'm experimenting with having RocketParts as a required resource which scales reactor output if the parts aren't there. (not a consumed resource but it has to be present on the part) - think of it in terms of shielding I guess? The more shielding is installed the more you can ramp up the output.

KiloPower is more forgiving in terms of mass. Heavier than an RTG but still light enough to transport. (3 and 7 tons for KP10 and KP40 respectively)

Starwaster:
Also I agree that temperatures (core temperatures) should be based on operating temperature and not when it slags. KiloPower is designed passively where if it overheats, core geometry expands which automatically cools it down and I tried to reflect that in its config but it needs work still because ModuleCoreHeat

thesciencenut:
@Bornholio @Starwaster I suppose it's possible that I'm misunderstanding what it actually is, but the plugin treats the critical temp as the point where the reactor will start to meltdown and fail. 

As for the core temperature, I kept finding a couple of different numbers and I really wasn't sure which one to use. Essentially the NFE plugin only cares about two temperatures for operation (three if you count the automatic shutdown temperature): the core temperature, which is also displayed as the nominal temperature; and then the critical temperature, which is the point at which it starts to take damage. In practice, what you see is a decreased output from the power generation of the reactor regardless of thermal output.

I can update that to be whatever makes more sense. 

It's worth noting that I did notice some parts in the source that could potentially be used to "overdrive" the reactor (I apologize if this is the wrong terminology) and generate more power at the cost of increased wear and tear on the reactor and increased heat output. I couldn't quite figure out how to properly utilize that and I also wasn't sure how realistic that was because I didn't see anything that mentioned doing that in real life, nor anything mentioning that as being something that you could actually do.


I'll also just put these here for future reference:
http://web.mit.edu/22.33/Chapters/Radiator/Radiator_Chapter-new2.doc
http://stuff.mit.edu/afs/athena/course/22/22.33/Pete/Radiator_Chapter-MRP1.doc

thesciencenut:
I should also add that initially the default value that seemed to be used for most of these reactors was 1300k, though I think this was actually put in Celsius despite saying otherwise due to the temperatures of everything else... regardless, this leads to a value of 1574k if we assume that the 1300k was actually in celsius. 

I couldn't really figure out why that particular value was used so often,  but then someone pointed out that 1574k was pretty close to the melting temperature of beryllium. As a result, I assumed that the critical temp was the point at which the casings of the fuel rods or the reactor would start to melt. 

Again, this is a simple fix if there is a better number to use. (also, things tend to explode when you reach that point)

That being said... having the failure point at the same as the critical point makes a lot of sense... as if the reactor melts down, the crew would probably die, right? I don't think that there is  presently a way to cause death due to radiation exposure, so maybe an explosion is the next best thing?

thesciencenut:
Brief update: last night I fixed the attachment node placement and continued to do more research on the NuScale, I noticed a discrepancy of the reactors mass from various sources. 

NuScales website lists the total mass as 700t while most other technical documentation lists 590t. My assumption is that the 700t figure may include other things such as the shipping containers or other things that aren't necessarily a part of the reactor. 

In either case, I'm presently going to use the 590t figure because that seems to be more wide spread and 700t seems obsenely high.

I can easily update that figure later on if anyone can help provide more accurate information. 

I will also update the fuel mass and temperatures as suggested. 


I will try to push these updates out later tonight when I get home.

thesciencenut:
@Bornholio 


WOOO More sources! 

I'm going to keep editing this comment from here on out with updates on the specs as needed (there are a lot of comments by me already)
UPDATED SPECS FOR NuScale

Reactor Name = NuScale
Nominal Temp(k) = 560k ///550F-575F operating, adjusted to kelvin
Shutdown Temp(k) = 640k
Critical Temp(k) = 2474 ///Initial guess, original RO config number adjusted to Kelvin
Fuel type = 4.95% Enriched Uranium of Unknown type
Fuel Mass(kg) = 8260kg ///as per Bornholio's numbers. 
Reactor Mass(mT) = 581.74mT ///590mT-fuel mass of 8.26mT
Reactor Heat(kWt) = 200000kWt
Reactor Power(kWe) = 60000kWe
Run Time(y,d) = 2y




Sources: 
https://www.blankrome.com/publications/nuscale-power

PR #2004, opened 2019-03-08
Thesciencenut nfe rtg fix

thesciencenut:
Fixes decay of RTGs over time

The decay of RTGs over time doesn't appear to be working properly, this fixes that and removes the Resource Converter module and replaces it with NFE's Radioisotope Generator module. 

Decay rate was taken from actual values taking both the decay of the fuel and the thermocouples. 

Note: This patch will only apply if both RO and NFE are installed.


thesciencenut:
Why was this closed?

raidernick:
it had conflicts, same as the other one. I agreed with pap to close them both. You need to work with the latest versions of these files if you want to open a PR.

raidernick:
if you want it reopened you need to fix the conflicts, that's why these have both been sitting and nobody has been testing them. Also it is why they weren't included in the latest releases.

PR #2018, opened 2019-03-29
RN kosmos 3 engine fixes

raidernick:
-Fix kosmos 3 configs for merging of engine and verniers


raidernick:
cancel this crap because I screwed up...

PR #2019, opened 2019-03-30
Create RO_UniversalStorage2.cfg

DocRockwell:


DocRockwell:
Configs for Universal Storage 2. 
All parts should be configured, balanced, and working. Some balance changes are anticipated though.
Baked in RP-0 support for the time being, with appropriate/reasonable part costs and tech node placement. 
The US switchable models aren't compatible with RealFuels, so I made some predefined tank configs for a few common fuels (ratio may be imperfect for your engine). The tanks aren't pressurized either (will RF recognize a pressurized flag on non-RF fuel tank modules?), so attach a pressurized tank and fill it with helium.
Parts all have their stats balanced according to equivalent RO certified parts. Some of the pressurized tanks should have been balanced to Tank-III-HP mass instead Tank-IV.

pap1723:
Integrated, thanks!
But I did remove the techs, that has to be put into RP0 instead.

PR #2024, opened 2019-03-30
Z titan plumes

Zantigar:
Fix LR87 and LR91 plumes for RN US Rockets versions of the engines--they were not appearing.


raidernick:
This is not needed, these were just updated for changes I made to the stock engine models and configs and works fine IF you are using those changes, which you clearly aren't. You are using the released version of the mod for 1.4.5, get the git master, I'm not merging this as it will break the mod once I do officially update it, which will be at the same time I update RO to avoid actually breaking things like you are seeing.

Zantigar:
Mea culpa.  I'll add this to the list of repos that I need to get from git and keep updated.

PR #2037, opened 2019-04-06
Add Kerbalism support (WIP, 1.4-1.6.1)

Standecco:
Numbers should be pretty much correct, part configs are still underway. Adding a GameData/Kerbalism folder is the only way to edit Kerbalism settings that worked. Not ready to be merged yet, I'm going to update it as more people (hopefully) give feedback.

IMPORTANT: Be sure to use the latest kerbalism master; without it, the profile isn't going to work


raidernick:
Besides the conflicts you have 2 files here that are not acceptable. RO CAN NOT be installing files anywhere but the RO directory. You have a rosettings.cfg that is being installed directly to kerbalism folder, that's not going to be allowed. You also have an xlsx file being installed to a notes? folder. That's also a big nope.

Standecco:
Conflicts aren't actually a thing, so don't worry about that. I didn't add the xlsx file (I don't even use excel) nor the notes folder, so that's gone. As I wrote in the OP, the Kerbalism folder with ROSettings is at the moment the only way to make this work in the first place, since Kerbalism (like most mods) doesn't read files from outside its folder. I'd be happy to delete it, but there is no other way at the moment; it doesn't replace anything in the kerbalism folder and can be deleted just fine if you want to uninstall RO, and everything will be back to normal.

raidernick:
That doesn't change the fact that I'm not going to have RO be installing a separate folder other than itself. If you want to add something to the kerbalism folder, give it to the person who makes kerbalism to distribute with that mod.

Standecco:
Moved to its own ROKerbalism mod. It will be an additional configuration on ckan installation, TBD for manual installation. For the moment, it's a standalone mod that can be tested by downloading it from here: https://github.com/Standecco/ROKerbalism

PR #2051, opened 2019-05-06
Ro 1.6.1

smartdummies:
Add plumes to the gas generator exhaust form the vernier engine for the MH LR-91


pap1723:
Can't accept this with all the changes added from before, please re-submit with an up to date branch.

PR #2059, opened 2019-05-13
Ua1205 thrustcurve

Schnobs:
UA1205_thrust
Old engine performance blue, new performace data in red.

Source: https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19760021170.pdf
Thrust curves on pp51-52, but also note comments about this being "slow burners" (p36).

Fun fact: now that I'm done, I notice that even the old curve may still be within the specification limits. I'd still like to get rid of it's long tail, though, where it's running at miniscule thrust for >10 sec.


raidernick:
Is the long tail not correct? These boosters are separated similarly to how the shuttle boosters were separated. If they didn't have a long tail and you separated them they'd immediately fall back and possibly damage the core. By having a small amount of thrust(i.e. around 2.5% for 5-10 seconds) it allows you to separate them and have them slowly fall back and ass the boosters rotate due to the sep motors the thrust actually helps to push them away from the core more.

MikeOnTea:
According to https://www.alternatewars.com/BBOW/Space_Engines/UA-1205_SRM_Thrust_Curve.gif which is apparently taken from NTRS 19750004937, the new curve looks closer to the real one. I suspect the old curve might've been a time curve directly entered into KSP again without proper conversion. However, there's a spike in the tangents, so i wouldn't merge this but redo the curve without the spike which i have done. Will do a test run tomorrow or give it schnobs to test it.

Schnobs:
The old curve's thrust during that phase is not sufficient for a shuttle-style separation, and I think it wasn't even attempted. In the plots from the post-flight report, one can see the jerk when the SRMs were separated. It looks as if they'd clung to them until the very last.

That said, I, too, would like to refine that curve some more, mostly to properly fill out the hump in the 20-30 second area. Problem is that I still cannot squeeze as much impulse out of the SRB as it's supposed to have had. I'm either short of thrust, or shut down too soon.

Schnobs:
UA1205_thrust
Redone the curve. 

Total impulse delivered works out after I've increased the ISP to 266 seconds, which is what my main source states.

raidernick:
Hey you wouldn't mind doing this for the other titan boosters like the ua1206, ua1207 and srmu if those also need an overhaul would you? I lot of the srb curves we have are just placeholders or copied from other boosters.

Schnobs:
I'll have a look, no promises though. I can probably stand only one or two more before I'm fed up.
If anyone happens to have *good* sources, please post links. Post-flight summaries usually are treasure troves, did they still do that in the 90's?

raidernick:
This has thrust curves for ua1205 and ua1207(and several other srb types): https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19720007150.pdf

For the SRMU I really can only find an image of the thrust curve here: https://www.northropgrumman.com/Capabilities/PropulsionSystems/Documents/NGIS_MotorCatalog.pdf

Schnobs:
Just added a curve for UA1207. Compares well to p.247 of the NTRS doc RN gave above. Could still need some refining, but is serviceable and an improvement over the current one.

PR #2062, opened 2019-05-17
Correct SRMU ThrustCurve

lukecologne:
SRMUComparison

Made using @MikeOnTea 's thrust curve script 

Data extracted from here : https://www.northropgrumman.com/Capabilities/PropulsionSystems/Documents/NGIS_MotorCatalog.pdf

with this tool: https://apps.automeris.io/wpd/


MikeOnTea:
@lukecologne any reason you removed the tangents? Was there something wrong with the direct output of my script?

lukecologne:
@MikeOnTea The values after the first two? I didn't know what to do with them, since all the other thrust curves don't have them, thats why I removed them. But the curve seems to work without it. Thanks for the info, I will include them in the future.

lukecologne:
@MikeOnTea By the way, you said you would collect the input data, if you want, I can post them here

Issue #2070, opened 2019-05-19
UA thrust curve overhaul


Schnobs:
WIP: provide two or three curves for each engine.
Current state:
![thrustcurve-overview](https://user-images.githubusercontent.com/8556997/57987556-316c5b00-7a83-11e9-9535-023eaeeb8452.png)
The UA120x curves have been based on NTRS 19720007150.

The saddle-shaped UA1563 is from DOI 10.1016/B978-0-08-006931-9.50010-6, a study for MLV-SAT-IB-27 (Saturn SIVB upper launched by SRMs). That's the only 1563 curve I could find. The regressive one is made up (and not even very well ATM).

I plan to make better curves for UA156x based on what I know about the motor, and the configuration possibilities outlined in the 120x doc.

But in the mean time, I'd like to have some input:
- I guess I'll set up a wiki page outlining the thrust curves, but IMO it's desirable that the names make sense without looking things up. "Config-XX" isn't very helpful. Can you think of a universal naming scheme that can be used across all large SRMs?
- Should run time be part of the config name, and if so, which? Docs usually use "web action time" (roughly, from ignition until the bend to tailoff), and/or "action time" (ignition to 10% thrust).
- what thrust levels do you want?
For example, my current UA1205-config-2 has higher thrust and shorter runtime compared to what I found in an actual post-flight summary report. Overall shape is virtually the same, though -- one could arrive at the real thing simply by setting the SRM thrust slider to 90%. This works for me, but Raidernick mentioned that it might not work for users of particular mods. Comments?
- if each thrust level has to be represented by a seperate engine config, how many are we willing to provide?

About ISP: the thrust/time curves often give thrust provided along the longitudinal axis. Nozzle cant is taken into account when stating ISP. This hardly matters for the 120" motors which had the nozzles canted at 5-6°, but the 156" ones came with cants of ~15 degrees. By 1970 the assumed ISP would've been >270s, but only 263s when considering useful thrust provided to the vessel. Needless to say, I'm going with the higher value for purposes of engine config. It will be up to modelers to cant the nozzle accordingly.

raidernick:
@Schnobs you can close that PR as you are right it's now not needed, we DO need something done with the 15xx ones though. You can probably make generic configs of those as long as you point FASA to the new configs.

PR #2071, opened 2019-05-19
UA SRM thrust curve overhaul

Schnobs:
Current state of UA1205 and 1207 configs.
Still WIP, but serviceable. PR provided if you want to merge it for the upcoming release.

Please have a look at https://github.com/KSP-RO/RealismOverhaul/issues/2070 -- I'd prefer if all substantial talk was there, rather than on a PR that might soon be closed.

I've got curves for the UA1563 as well, but can't find the config file...?

EDIT: I just notice that there's duplicate work from Pap and me. Is my contribution even desired?

EDIT2: found the 156x configs in GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_SRBs.cfg -- should I try to bring them into mainline? How much of that FASA-specific file is even needed anymore, shouldn't most of it be overwritten by generic configs?


pap1723:
This was duplicated as it appears we both worked on them at the same time. Sorry about that!

PR #2104, opened 2019-07-18
New comms master merge

rsparkyc:


raidernick:
little late on this huh lol

MikeOnTea:
Updated my branch so it's up to date. (Doesn't contain an updated dll but i think i prefer to package releases with a dll).

PR #2106, opened 2019-07-18
Add support for integrated RCS in B9 HL and S2 cockpits

whale2:
Some B9 cockpits have support for integrated RCS, why not add it to RO?
Besides, without this patch, cockpits have RCS FX always on without actual RCS

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RO_B9.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RO_B9.cfg
@@ -38,7 +38,9 @@
 		@basemass = -1
 		@volume *= 2
 	}
+	!MODULE[ModuleFuelTanks],1 {}
 	!MODULE[ModuleReactionWheel] {}
+
 }
 
 @PART[B9_Aero_HL_Adapter_Front]:FOR[zzzRealismOverhaul]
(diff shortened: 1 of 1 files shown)


raidernick:
This needs to be RCSFX and you need to define EFFECTS for it. The older way of doing it you have here will work in 1.3.1 and before versions of KSP but will newer versions it's going to cause issues.

whale2:
@raidernick Not sure I understand. B9 already has RCSFX:
https://github.com/blowfishpro/B9-Aerospace/blob/master/GameData/B9_Aerospace/Parts/Cockpit_HL/HL_Aero_Cockpit.cfg#L112

So the first thing I encountered after starting up 1.6.1 RO, that B9 cockpits have their RCS effects always on, but without having actual RCS.
So either RO should disable RCSFX on B9 cockpits that have it, or add actual RCS. Where's my mistake?

raidernick:
You changed all the ModuleRCSFX back to ModuleRCS by searching for ModuleRCS* and then doing @name = ModuleRCS.

whale2:
Oh. I feel so stupid now. But apparently, that line didn't work, I still have ModuleRCSFX in MM config cache. Ok, I'll fix it for sure.

whale2:
Hmm, all other RO patches for B9 have exactly that style - 
 @PART[B9_Control_RCS_Block_R5]:FOR[RealismOverhaul]
{
 ...
        @MODULE[ModuleRCS*]
        {
                @name = ModuleRCS
...

I guess this comes from times when B9 didn't use ModuleRCSFX, doesn't it?
https://github.com/blowfishpro/B9-Aerospace/commit/5b133e6fa3a193c9bd133bcae252ad13da0a60bf

raidernick:
yes it does and that's wrong. You HAVE to use rcsfx in ksp 1.4.5+ otherwise the plumes will be missing, reversed or generally messed up

whale2:
Ok, I'll prepare a patch for all other B9 RCS thrusters then.

PR #2108, opened 2019-07-20
newComms antenna patches (attempt 2)

Gordon-Dry:
Please provide your edits and additions.

The dedicated patches come first, using BEFORE[zzzRealismOverhaul]
- CoatlAerospace
- CommNetAntennasExtension
- CxAerospace
- DMagicOrbitalScience
- FASA
- JX2Antenna
- KerbalGPS
- MechJeb2
- NAU-ORIGAMI
- NearFutureConstruction
- ReStockPlus
- RN_Cygnus
- RN_Salyut
- RN-Skylab
- RN-Soviet_Probes
- RN_Soyuz
- RN-US_Probes
- ROCapsules
- RP-1
- Stock_RO
- SXT
- Taerobee
- TokamakIndustries
- UniversalStorage2
- VenStockRevamp

As you can see, the list is far from finished. (These are only the parts I got in my install plus a few part packs just added for creating the config and removed afterwards)
- Please provide your own patches in the same manner, so paste them into the config at the proper place (by alphabetical sorting by mod names) for easy maintenance

The generic patches come after the dedicated patches, using FOR[zzzRealismOverhaul] for the main part and AFTER[zzzRealismOverhaul] for the finalization part.

They provide some simple formulas to use the antennaPower of the ModuleDataTransmitter as a calculation base and in some cases even the mass of the part.
This is to avoid all generic patched antennas having the same values.


Gordon-Dry:
From my side this looks good.
If the force push is good or bad - I can't tell.

DRVeyl:
I don't think it's useful to configure hundredths of a dB of gain for all of those integrated omni antennas.  Specifically, all of those reference gains at 0.05-0.1, and generally anything < 1.0.  They should be less than the stock "SurfAntenna" part (2.0?).  1.0-1.5 is IMO most appropriate for omni antennas integrated into the avionics.

I assume your diameter values come from measuring the part versus a procedural tank in the VAB.  That's a useful method.

Anything with "HGA" or even "LGA" in the name is probably not an omnidirectional antenna and should get a diameter value, rather than a low referenceGain.  (Some of the ca_landv_* parts)

Do the DMagicOrbitalScience SIGINT antennas actually have a diameter of 52m and 105.4m?  That seems incorrect, and I imagine they should be a science experiment if anything, but not a comms antenna.

Is jx2LDA really 21m diameter in the VAB?

I don't think it's practical or maintainable to individually patch all of these extremely similar parts (all the integrated omni antennas).

Gordon-Dry:
@DRVeyl 
Well, a generic patch could be a better idea for many parts.
But then it needs to be a really good generic patch.

The DMagic parts really have that big dishes, unfoldable. But yeah, they're experiments.

I measured all dishes vs. a proc part, yes. And I also got infos from the real life counterparts.

The jx2LDA should really have a 21m dish, it's unfolded then. Perhaps this is the wrong value, because the unfolded part is not a 100% circular shape.

I'll check the ca_landv_* parts again.

I'm not sure about your first paragraph.

Gordon-Dry:
@DRVeyl The ca_landv_* HGA parts are rectangular shapad flat discs. So I assume they're omni.
The LGA antenna is definetely not a dish. It's a stick.

DRVeyl:
Can't just add diameter.  You can determine the **area** of each section, sum those up, then compute the diameter of a circle that would correspond to that area, and that's a reasonable equivalence.

Can you screenshot the parts, as I do not have them handy?  Flat panel arrays aren't omni and should be treated akin to a dish.  ("Dish" is being used loosely to correspond to any directional antenna.)

I'll back off that "LGA" could still be omni.  It's a bizarre naming pattern to me, but ok.

Gordon-Dry:
@DRVeyl 
can_landv_* are Surveyor parts

ca_landv_hga is a combined antenna and solar panel:
Sideview extended
screenshot_2019-07-23--16-27-45

Antenna side
screenshot_2019-07-23--16-27-37

ca_landv_orbiter_HGA
Just the antenna
screenshot_2019-07-23--16-30-57


And again, never has a chance to get my fingers to RA, so I'm really new to this in KSP.
You wrote:
> I don't think it's useful to configure hundredths of a dB of gain for all of those integrated omni antennas. Specifically, all of those reference gains at 0.05-0.1, and generally anything < 1.0. They should be less than the stock "SurfAntenna" part (2.0?). 1.0-1.5 is IMO most appropriate for omni antennas integrated into the avionics.

What do i have to do now?

Gordon-Dry:
@DRVeyl I would like to know which are the cases when it's allright to have a referenceGain < 1

DRVeyl:
@Gordon-Dry Thanks for the screenshots, that clarifies for me.  Determine the area of that square antenna.  Since area=pi * r * r, divide by pi, take the square root, and then double the result to get an effective diameter if it were circular.  You can generally use that process for a directional antenna.  ... We went with diameter rather than area because we thought it would be more direct most of the time.

I would not use referenceGain < 1.0 unless you can find specific documentation showing that.  (I don't think you'll find it often, and that's fine, it's early and we can still paint with broad brush strokes, so to speak).

I would suggest you divide the omni-directional antennae into two sets: those that are dedicated antenna parts, and those that are integrated into something else eg a probe core.  Those integrated antennas I'd give a referenceGain=1.0.  The dedicated parts I would assign referenceGain=2.0.  I'd also suggest a syntax like @PART[item1|item2|item3...] so that when we decide to rebalance either set, we can do so without changing hundreds of patch lines.

All the directional ones will remain part-specific patches, because they probably have fairly unique diameters/areas.  @MikeOnTea or someone else can comment on if they prefer this single-file patch support for multiple mods, or if they prefer it broken up into multiple files for each mod.

Gordon-Dry:
That force push forces me to fiddle even more. Too much of a hassle.

PR #2126, opened 2019-08-22
Update SSTU_Tanks.cfg

Starwaster:
* Changed targeted part names where applicable to keep up with SSTU development
* Added config changes to SSTUModularPart to force it to not make mass/cost changes. (Real Fuels will handle this)


Starwaster:
Only gave SSTU-TANK-MFT-R MFT type = ServiceModule; someone check that please to see if you want to give it one of the other SM types.

Starwaster:
btw, these updates are critical or there will be negative mass issues on SSTU tanks due to both SSTUModularParts and ModuleFuelTanks trying to null the mass on the part.

siimav:
Maybe you'll one day merge my PR as well 😞

PR #2130, opened 2019-09-13
Adding the Space Transportation Main Engine to the RS-25

Ash19256:


Ash19256:
This is just me posting this pull request for Scorpu on the RO Discord. It adds a gas generator version of the RS-25 SSME, intended for expendable rockets.

vader111:
Theres no need to create a engine config for the STME, i have already created one that has been merged with master

Ash19256:
One, I didn't create this, Scorpu on the RO Discord did. Two, your STME configuration appears to be in use with no parts that I am aware of - thus, this configuration edits the SSME config to add the STME as an available configuration.

vader111:
@Ash19256 thats true, but instead of writing out a config, you just need to attach the engine config to the engine you want it on

siimav:
Ideally someone would create a model similar to SSME but with gas generator hardware and we'd add that to ROEngines

vader111:
even better idea, since yes it is a gas generator engine and bigger then the rs 25

pap1723:
I don't agree with the numbers here. According to the official NASA studies on it, which can be seen here: https://www.alternatewars.com/BBOW/Space_Engines/STBE_STME-481_Details.png

It should be more like this:
Max Thrust = 2139.59
Vac ISP = 461
SL ISP = 380.4

vader111:
@pap1723 Bit different then what i found on the engine 
https://www.alternatewars.com/BBOW/Space_Engines/STME_Data_Excerpt.pdf
http://www.astronautix.com/s/stme.html

PR #2143, opened 2019-10-29
Add RA modules to FASA and BDB parts

Capkirk123:


lukecologne:
You should probably add a :NEEDS[RealAntennas] so it only gets applied when RA is installed

PR #2144, opened 2019-10-30
Add RA modules to FASA, BDB and DECQ Apollo

Capkirk123:
Improved RA patches so they only apply when RA is installed, added DECQ patches and commented justification for antenna stats


Capkirk123:
Forgot how Github worked, sorry. Reverted changes

PR #2145, opened 2019-11-02
Recompile for KSP 1.7.3

raidernick:
-Recompile for KSP 1.7.3
-Update version file


siimav:
Pretty sure that RO doesn't need a recompile to work on 1.7.3. @pap1723 can correct me if I'm wrong here. 
The next RO release should target both KSP 1.6.1 and 1.7.3. We need at least one more update on 1.6.1 for the upcoming RP-1 release.

raidernick:
@siimav it does to bump the file version and it DOES target 1.6.1. I recompile using the 1.3.1 binaries so it supports all versions from 1.3.1 to 1.7.3 and officially in the version file it supports 1.4.5 to 1.7.3. The only reason it doesn't go back to 1.3.1 in thee version is that some config changes were made for 1.4.5 that don't fully support 1.3.1 anymore

siimav:
It's all good then.

PR #2152, opened 2019-11-22
Make DRE heatshields available even with ReStock

lpgagnon:
Because sometimes you want to build a .craft you can share
with people who don't have ReStock


lpgagnon:
(also because _installing_ restock shouldn't break .craft any more than it has to)

PR #2164, opened 2019-12-04
Add Navaho XLR43 booster

Capkirk123:
Add engine Config for XLR43, the Navaho booster and predecessor of all american kerolox engines. It started as an upgraded, single chamber A-4, the XLR43-NA-1. The NAA 75-110 was derived from it, and was very similar, although much more reliable. It later was upgraded with a bipropellant gas generator, much higher chamber pressure, and switched to burning Ethanol 90 as the XLR43-NA-3. It flew on extensively on Navaho tests (in pairs as the XLR71) and became the Altas boosters when upgraded to burn kerolox (as LR43, already in game).

I believe this engine should be added because it adds an American equivalent to the RD-100 series, and was a major stepping stone to large, high performance boosters. 

RP-1 ECM configs (please commit this first so players don't get it for free): https://github.com/KSP-RO/RP-0/pull/1158
ROEngine part: https://github.com/KSP-RO/ROEngines/pull/34


lpgagnon:
tested alongside ROE PR; both configs working as expected for me. (those short burn times make it tricky to use effectively, but they're sourced)

PR #2179, opened 2020-01-03
RealPlume Template System

lukecologne:
This adds a template system to improve the current workflow of configuring complex RealPlume plumes, by doing part of the work for you. The template sets up the `PLUME`-Node with properly scaled values, and sets up the effects in the engine. 
It requires a `scale` value, an `offset` value, and a `transformName`. Optionally, you can add a `setupEngine = false`, to prevent the template from setting up the effects in the engine module, if you want to use a combined plume, per config plumes, etc.
Examples can be found in the ROEngines repo.

I still have to decide about hydrolox templates, so I left it as a draft PR.


lukecologne:
So, since I didn't have much time in the last week, I think it should be good to merge. If I have more time, I will add more templates as needed in future PR's.

siimav:
@lukecologne What's the status of this PR? Am I correct that the templates do not change anything by themselves and require a mod like ROEngines to explicitly take usage of them?

Btw, looks like some of the files have mixed tabs and spaces.

lukecologne:
Yes, this PR changes nothing by itself, and changes nothing about the way the old plume stuff works. If you want to use it, you need to use a PLUME_TEMPLATE node instead of the normal PLUME node (should probably be documented in the wiki or somewhere appropriate).

About the tab/space thing: I'll go through the files and try to replace the tabs with spaces.

PR #2182, opened 2020-01-05
Early Soviet Hypergolic Engines

Capkirk123:
The RD-211 series was an early Soviet four combustion chamber rocket. Developed from the RD-100 for the R-12 IRBM in the mid 1950s, it provided a significant increase in power over earlier Soviet engines. It was later used for the Kosmos-2 launch vehicle.
Engines:
RD-211
RD-212
RD-213
RD-214
RD-214U

The RD-215 series was a soviet dual combustion chamber gas generator engine. Designed in the late 1950s to use storable propellants for ICBMs. It was used in pairs in the R-14 and R-26, and in threes in the R-16 and R-36, and later in the Kosmos and Tsiklon launch vehicles. After initial issues, it proved extremely reliable.
Engines:
RD-216 (2x RD-215)
RD-216M (2x RD-215M)
RD-218 (3x RD-217)
RD-224 (2x RD-225)
RD-251 (3x RD-250)
RD-261 (3x RD-250PM)

The RD-109/119 was a Soviet high performance semi-cryogenic upper stage engine. Originally developed as an upper stage for the R-7, it was later adapted to to Kosmos-2 Launch Vehicle, which needed a high performance upper stage due to the low performance of its R-12 first stage.
Engines:
RD-109
RD-119

The RD-219 was a vacuum variant of the RD-215, featuring a slightly larger nozzle and higher performance. It was only ever used by itself on the R-16, R-36 and Tsiklon as a second stage engine. The initial versions were unreliable, but later versions, such as those used on the Tsiklon, are extremely reliable.
Engines:
RD-219
RD-252
RD-262

The RD-215/RD-219 was the Soviet equivalent to the American Titan-series rocket, and I found it a little odd it hasn't been added yet despite having a history nearly as long as the RD-107/RD-108. There are not any engine models for the RD-215/216 series to the best of my knowledge, but if anyone does find one or just wants to use a stock model I made this.


raidernick:
@Capkirk123 I have engine models for both the kosmos 2 and kosmos 3 rockets I made so they do exist.

Capkirk123:
Oh I didn't see those. Is the engine model a single engine or an engine pair?

raidernick:
It's a pair and it has built in thrust vanes as the engines were only ever used with those since they don't gimbal.

Capkirk123:
Well that should work too. I'll just set the cluster multiplier to 2. It'll make some engines configurations that technically didn't exist IRL (a cluster of 2 rather than 3) but it should otherwise work fine. I am also working on the RD-214 for the Kosmos-2.

PR #2190, opened 2020-01-16
Fix RD-119 engine config

Capkirk123:
Fix RD-119 engine config so it uses the global configuration rather than hardcoded settings.


raidernick:
@Capkirk123 I'll deal with this when I have a chance. It requires more than just changing it to the global config to work because of how I have the verniers work using the rcs module. I'm going to close this for now as it's already on my todo list.

PR #2192, opened 2020-01-19
Add AJ10 Transtar

Capkirk123:
The AJ10 Transtar was a proposed pump-fed upgrade for the AJ10. Developed in the 1980s as a high performance upper stage and tug. Based heavily on shuttle OMS, using the same nozzle and combustion chamber, with turbomachinery installed in place of the pressure-fed system. It was tested extensively, but never flew, as program was cancelled following the Challenger disaster and changing priorities of NASA.


GameData/RealismOverhaul/Engine_Configs/AJ10_Transtar_Config.cfg
```diff
@@ -0,0 +1,317 @@
...
+//	Chamber Pressure: 350 psi
+//  =================================================================================
+//
+//	AJ23-154 (1997)
+//	Ready by 1997, for OTV
+//	Burn Time: 72000 s
+//	Thrust (vac): 13.3 kN
+//	ISP (vac): 483 s
+//	O/F Ratio: 6.0
+//	Propellant: MMH and NTO (MON3)
```
siimav: Should be hydrolox

Capkirk123: Fixed

PR #2196, opened 2020-01-25
Update STME Config & Add STBE Engine Config

vader111:
Updated STME Engine with better source, Reduced the rated burn time to the same as the rs-25.

Added Space Transportation Booster Engine (STBE) Config, The only values that were not available was the Vacuum ISP and any reliability values, so since the engine has a decently sized engine bell i have given it a reasonable vacuum isp, i have also given it F-1A testflight values and burn time. The engine was designed to burn a 50-50 mix of kerosene and gasoline & LOX, but decided to have it burn pure kerosene & LOX.


nepphhh:
Page 24 of the source document contains chamber pressure & mixture ratio information: https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19860001749.pdf

vader111:
@nepphhh just tryed to use rpa and i really don't think we have enough stats to run the engine through it. If RO has JP-4 resource to use i will switch it over.

PR #2198, opened 2020-01-28
NewTanks

nepphhh:
This is an overhaul of the RO procedural tank system, created by [Kirk ](https://github.com/Capkirk123) and Neph. Rather than the current abstracted tank system, this implements a new tank system based on historical construction and technology. The old tank system is retained, but marked as depreciated, for compatibility purposes.

In this new system, there are three types of tanks: separate structure tanks, integral structure tanks, and balloon tanks. 

*  **Separate structure tanks** are based off of stringer style tanks, in which the structure is welded to the inside or outside of the tanks. They are simple and cheap to build, but have relatively poor mass ratios. Most early rockets like V-2, Aerobee, Redstone, and R-2 used this style of tanks. Later Soviet rockets like the R-7, R-36, and Proton also used this style of tankage. In our system, these tanks are starting tech. They gain slightly better performance with improved tech, while maintaining a low price, making them attractive for dense, cheap, Soviet style LVs.

* **Integral structure tanks** are are based off of isogrid tanks, in which the structure of the tank is machined into the tank walls. They are very light, but expensive and difficult to machine. They are based on the tank isogrid tank structure pioneered by Thor, and used by almost all American LVs. In our system, they unlock with satellite era materials science, and diverge from separate tanks at that point. They become much lighter and much more expensive over time, making them useful for small, highly efficient LVs, similar to most US and European LVs.

* **Balloon tanks** remain largely unchanged from the current system. They also unlock with satellite era materials science, and minor upgrades are available over time, making them slightly lighter and more efficient.

These changes are based off of extensive research of real rockets, which has been compiled here:
[New Tank Definitions.xlsx](https://github.com/KSP-RO/RP-0/files/4120889/New.Tank.Definitions.xlsx)

Additionally, these historically-obtained numbers have been "stupid-checked" in the RP-1 environment by comparison of minimal stages (composed of engines + tank sized & filled to rated burn time + contemporary avionics configured to tonnage) with historical dry stage masses. Discrepancies were revised.

Descriptive titles and historical & factual information have been provided for each configuration in the case that future UI iterations will be able of delivering additional information to the player about the tank they are using.

The RP-1 PR, containing the rest of this overhaul, is located here: https://github.com/KSP-RO/RP-0/pull/1195


siimav:
I would prefer the _Seperate Structure_ and _Integral Structure_ to be combined into a single part instead of having 2 different parts for them. This makes it a lot easier for the player to compare the different tank definitions on a launch vehicle they are building. Having to dismantle half of a launch vehicle just to replace a tank part gets tedious really fast. This also makes upgrading an existing stage a lot easier when better tank tech becomes available along the line.

siimav:
Looks like this PR removes `Tank-III-Balloon` which is going to break existing vessels that still use it. All the existing parts and definitions should still be available during the transition period.

GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
```diff
@@ -414,7 +414,7 @@
 		utilizationTweakable = true
 		maxUtilization = 88
 		utilization = 88
-		type = Tank-I
+		tank = Tank-I
```
siimav: Pretty sure that isn't correct

GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
```diff
@@ -460,8 +559,10 @@
 		utilizationTweakable = true
 		minUtilization = 99
 		utilization = 100
-		type = Tank-III-Balloon
-		typeAvailable = Tank-III-Balloon
+		type = Tank-Balloon-SteelAl
+		typeAvailable = Tank-Balloon-SteelAl
+		typeAvailable = Tank-Balloon-SteelAlCu
+		typeAvailable = Tank-Balloon-SteelAlLi
```
siimav: You hijacked the existing balloon tank part which means that vessels from old saves will be broken.

nepphhh: Fixed

GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
nepphhh: Fixed, thanks!

PR #2206, opened 2020-02-16
Lunex RL10

Capkirk123:
Add the Lunex RL10 configuration to the RL10. The Lunex RL10 was intended to have a 6:1 throttle ratio. It is based of the RL10A-3-3, since that was the most recent RL10 available at the 1966 first flight test goal. Some performance loss is assumed due to hydrogen bleed and valves required to make throttling possible.


lukecologne:
It seems like you got the descritptions added in your other PR in here as well. I think that this may cause merge conflicts.

Capkirk123:
I removed the descriptions from the Descriptions! PR to prevent conflicts (hopefully).

Capkirk123:
Fixed conflicts with BDB

PR #2207, opened 2020-02-16
RN Soyuz Update

Capkirk123:
Add improved Kerbalism support, fix descent modules using monopropellant instead of HTP/bipropellant, added RealAntennas support, and added more descriptive descriptions.


raidernick:
this is wrong, very wrong. This is why I don't want people editing my configs when I'm actively working on them. The descent modules no longer HAVE rcs on them, the rcs was separated into it's own part. Please don't edit my configs with wrong settings while I'm current making changes to the models. The RCS configs were removed from the SA for a reason you are working with both an outdated version of the configs AND the base mod. This will not be ok to merge and will need to be redone. This is the same issue I was discussing on the discord the other day about changes being made that break things.

Capkirk123:
I didn't realize I was working on the incorrect version. This is put on ice until further notice

PR #2218, opened 2020-03-01
Leave ATK Propulsion Pack SRB parts

DocRockwell:
This was deleting various fully RO/RP1 configured SRB's from the ATK Propulsion pack.


raidernick:
Yes, it does this on purpose as the delta ii pack has these exact same srbs in it as the atk pack does except the models and textures are better. This is intended and not a bug.

nepphhh:
Wouldn't removing "worse modeled" parts be something more suitable to a mod like ROEngines? Are there any other cases in which RO makes a decision for the player about what parts they should have to use?

raidernick:
the parts are literally the exact same part. The models are exactly the same except that the delta ii pack versions have updated colliders as that was made at a later date. Yes it should absolutely make this decision as having an exact visual duplicate of the same part that has the exact same stats and name because they both use the generic config is not only confusing, but it can cause issues with craft files as you will have no idea which srb was used because they look exactly the same.

DocRockwell:
I only have ATK Propulsion pack, not the Delta 2 pack, and that line is removing the ATK parts.

nepphhh:
If you build a vehicle using ATK & then install Delta II, won't this break your craft file and anything in flight?

raidernick:
Then it somehow broke in a newer change of mm or ksp. Try changing FOR to NEEDS. If that doesn't work I need to look at it myself

raidernick:
@nepphhh if you want to have both parts loaded into the game you need to make a distinction between them then. They can't have the exact same name as then there is no way to tell where they came from. If you want both to be visible then make a PR that changes the ATK ones to be prefixed ATK and the LP ones to be prefixed LP or something. I'm not accepting a PR to just have both copies with no distinction as it will cause more confusion.

PR #2222, opened 2020-03-06
Hide deprecated parts from the editor

Standecco:
No need to keep showing them if they're deprecated and shouldn't be used anyway. This will make the part list less cluttered.


nepphhh:
would suggest also deprecating the FF1, 2, & 3 Aeroshell fairings, they seem completely broken in 1.8 & proc fairings are in all ways preferred

also some timing is going to have to happen to this slew of new deprecated parts covered yb your pr: https://github.com/KSP-RO/ROTanks/pull/27

PR #2240, opened 2020-03-19
Updated Some SSTU Plumes to Newer Plumes

IronKerbal:
I updated some SSTU engine plumes with Newer effects, the old ones are nearly 4 years old now.


siimav:
@lukecologne are these even relevant in the context of the plume templates you created?

IronKerbal:
I tested them Rigorously, they work 100% fine

PR #2247, opened 2020-03-25
More BDB configs.

mcd0uble:


stratochief66:
There is a lot of erratic formatting in the RealPlume configs. Could you regularize those a bit?

You can probably remove the thrust curves for the UA1205 and other UA12XX engines and put in an engine tags like `%engineType = UA1206`

The bluedog_GATV_SPS_LFO looks like it has mismatched fuels in the tanks vs ModuleEnginesFX ?

The desciptions in RO_BDB_engines.cfg appear to overwriting with `abc` rather than text, which doesn't seem right.

Down here: https://github.com/KSP-RO/RealismOverhaul/pull/2247/files#diff-1037d07c037348ddd7c28803c060117eR1371
Could you leave the engine as a tagged `engineType = LR87`?

The only part I ran into obvious trouble with was the mercury parachute. The realchute config for it seemed to be broken, crashed right into the ocean full speed.

I've made it down to the Saturn file, but haven't reviewed that yet.

mcd0uble:
Yep good catch on the GATV SPS, one of them should have been changed to MON3. 

I know there's a reason I didn't use the UA12xx engineType but now I can't fully remember what it was... I think I wasn't getting correct burn times so I adjusted some of the numbers to come in line with real world performance.

The descriptions of engines doesn't really matter if they have an engineType because it gets overwritten by that anyway. I think the abc thing was a placeholder I was using to make sure the configs were applying properly. Oh and some engines weren't able to use engineType configs without the plumes breaking, for example the Vanguard X-405 and the Titan engines. Zorg did configs for those that I'm pretty sure are complete.

The mercury parachute just doesn't want to deploy, either with stock or realchute configs, don't really know what to do about that.

stratochief66:
Fair, if the descriptions get overwritten, can you remove the `abc` descriptions then? 

We can punt figuring out the mercury parachute issue, if it doesn't seem to work with either stock or realchutes in RSS/RO. Could you remove any chute configs from it, maybe update the description to note that it doesn't work (to avoid surprise failures for people) and apply the landing chute config to the drogue chute part? That way, the Mercury can still be successfully landed (since the drogue part is successfully deploying)

stratochief66:
The saturn iv tankage isn't scaled up sufficiently. With this it seems to be 3.5m diameter but sources suggest it was 5.5m in diameter. Odd I know, but 1.6x rescale doesn't seem sufficient.

mcd0uble:
Yeah I'll remove those descriptions and put a note with the parachute. As for the Saturn sizing, that's an issue where BDB (and a lot of other "stockalike" mods) choose to use standardized part diameters, so you can either rescale them to their real-world sizes (and then parts like adapters will not fit properly) or you can just scale everything by a standard amount so everything fits together, but then parts won't exactly fit their real world sizes.

I'm not really sure what to do in those cases, I personally prefer to have realistic proportions and just use proc interstages and things like that if necessary, but I know others feel differently. To be honest I am kinda just ignoring Saturn for now because it;s one of the oldest parts of the mod and is due for a revamp soon (and I believe Cobalt was going to change the diameters).

Oh and last thing is I know some of the formatting especially in the RealPlume configs was a bit messy but all of those configs were just copy-pasted from Zorg's already working configs, so I feel like it's not really worth going in there and fiddling with things as long as it's functioning.

stratochief66:
Sounds good, I'd say just remove the configs for the Saturn parts. If someone wants to come along and scale them to precise size and volume to be realistically usable they can, but partially scaled up they aren't really useful to either camp.

If the RealPlume configs work, formatting is less important yeah.

PR #2252, opened 2020-03-26
Stock mk2 bicoupler

whale2:
Stock bicoupler converted to highly pressurized => heavy. This patch adds structural variant of it.


raidernick:
FOR[RealismOverhaul] should be changed to :BEFORE[RealismOverhaul] when you are doing an addition. Once you fix that I will merge this.

whale2:
Done.

PR #2255, opened 2020-03-29
Restore original ejection force on payload bases

### LINKED ISSUE #2254: Why are procedural fairing payload bases being reset to 0 ejection force?
https://github.com/KSP-RO/RealismOverhaul/blame/master/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg#L112

The commit where this was changed was a long time ago, but it was part of a commit to change node sizes, and no rationale was given for the change.

Interstages are left at their default 250 force, but payload bases are reset to 0.

jwvanderbeck:
Instead set ejectionForcePercent to default to 0%.  This way if the player wants to have some ejection force, they can by increasing the percentage in the PAW, but by default, the behaviour will be as before.

**NOTE**: While this does not break existing craft, those crafts would most likely have been saved with 100% ejection force, so in most cases, the player will want to set that to 0% manually before using the craft,

Fixes #2254

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
@@ -109,7 +109,7 @@
 	%MODULE[ModuleDecouple]
 	{
 		%name = ModuleDecouple
-		%ejectionForce = 0
+		%ejectionForcePercent = 0
 		%explosiveNodeID = top
 		%stagingEnabled = True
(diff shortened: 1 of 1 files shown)


NathanKell:
Makes sense! Don't think we can avoid the save incompatibility without a
code change that isn't worth it.

Sent by my thumbs, slowly.

jwvanderbeck:
Tested locally on my own install. It would be nice to have additional testing but its a pretty simple change, but other than that its good to merge whenever you want.

PR #2256, opened 2020-03-30
Fix issue with KSPWheel when VSR is installed, lower steering response

lukecologne:
This fixes an issue with Vens Stock Revamp, where the smallest rover wheels were breaking instantly. This was caused by a collider on the VSR side, which blocked the KSPWheel collider. Along with that fix, the torque of said rover wheel is lowered to be inline with the other ones, and the steering response for all wheels is lowered to a sensible level.


whale2:
Tested in 1.8.1 with small rover wheel - works ok.

PR #2273, opened 2020-04-23
Titan ascent engine

Katniss218:
Neph wanted me to make a PR with my custom (RPA-valid) titan ascent engine, so here it is.

The original, official engine used in my Othrys program


raidernick:
Speculative engines are already riding the line of acceptability. RO is supposed to have REAL WORLD engines only. Sure, you could conceivably do the math and come up with any working engine design you wanted, but this is not the mod for that kind of thing sorry.

PR #2276, opened 2020-04-25
Increase internal SM tank volume

### LINKED ISSUE #2275: Advanced Command Module, a Gemini alternative, has considerably less tank volume
Given that the Advanced Command Pod (`ROAdvCapsule`) is presented as an alternative part option for a Gemini style 2 crew program, uses the same training in RP1, costs the same as Gemini in ROCapsules (`ROC-GeminiCM`), and shares the unlock cost, and both more massive and physically larger, it seems really odd that it has nearly **1/10th** the internal SM volume.  Especially when you consider that the Gemini part also has an external SM in addition to the CM yet still has 525l of internal SM volume compared to the Advanced Command Pod's measly 60l

tl;dr why does the Advanced Command Pod have only 60l internal SM volume while the Gemini from ROCapsules has 525l?

jwvanderbeck:
Increase internal SM volume to 530l even without TACLS, for Kerbalism players

Fixes #2275

### DIFF
--- a/GameData/RealismOverhaul/Parts/AdvCapsule/ROAdvCapsule.cfg
+++ b/GameData/RealismOverhaul/Parts/AdvCapsule/ROAdvCapsule.cfg
@@ -392,7 +392,7 @@ PART
     {
         name = ModuleFuelTanks
         type = ServiceModule
-        volume = 60
+        volume = 525 // match volume of ROC Gemini
         basemass = -1
 
         //  Batteries 4 kWh.


Capkirk123:
Seems reasonable, but you might want to have kerbalism patch the volume instead.

jwvanderbeck:
Well its not really a Kerbalism patch so much as a "don't only do this for TAC-LS" patch I think.  Plus I don't think Kerbalism itself patches any of these, it just uses whatever RO sets, and right now RO only sets it up if you have TAC-LS

DRVeyl:
Will merge this.  Note that the volume is still different between if TACLS is installed versus not (525 vs 530).  The PART config needs a maintenance pass in general, I will probably look at that after the heatshield support patches merge.

PR #2278, opened 2020-04-25
CryoEngines update

undeadfake:
Just an update to make this config work with the latest version of CryoEngines.

Apart from the changes in part names, I also added new parts (RL10A-5, RL-60 and RL-60 cluster) and made some tweaks.

A couple of notes:

1. Edited the Vulcain engine to only use the Vulcain 2 config since that's what the part is modeled after. I can change this if it's not desired.

2. Even though the RS-68 roll control exhaust animation is broken, it works just fine. Couldn't find what the deflection angle is, so I used 15 degrees (copied from ROEngines).

3. Rescaled the RS-68 based on the boattail diameter (so it fits the Delta IV's CBC) instead of the nozzle exit diameter. As a result, the engine is slightly smaller than the real RS-68.

4. Since there's no LE-9 default config available, I used the LE-7 one for the Fuji, but kept the scale of the engine according to the LE-9 nozzle exit diameter (1.59m according to the NF_Realism team).


Capkirk123:
Looks good, everything works fine

PR #2281, opened 2020-04-29
Reduce ejection force from Procedural Fairings and interstages

leudaimon:
Procedural Fairings and Interstages have an ejection force of 250, which amounts to roughly 2.5kN over a second. That is a lot of force, especially if you apply it to a small probe. This PR changes it to 25, equivalent to 250N. Until something similar to the scaled impulse implemented in procedural decouplers is adapted for PF, this should keep these values more sane and less exploitable.

### DIFF
--- a/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
+++ b/GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_pFairings.cfg
@@ -109,6 +109,7 @@
 	%MODULE[ModuleDecouple]
 	{
 		%name = ModuleDecouple
+		%ejectionForce = 25
 		%ejectionForcePercent = 0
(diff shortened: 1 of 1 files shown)


jwvanderbeck-Unity:
What is the rationale behind this change, and is it applied to both interstage bases and payload bases?

Limiting the ejection force limits creativity and options to builders.  If you simply want less ejection force, that is what the percentage slier in the PAW is for.

Capkirk123:
Looks good. Explosive bolts can't provide lateral forces at all, and spring-loaded deploy mechanisms tend to be only enough to push things apart a little. Separation motors and RCS are used for a reason, decouplers being able to generate strong forces is unrealistic

jwvanderbeck:
yes but sep motors are all butt f'ing ugly in the game and no one wants to actually use them.

PR #2282, opened 2020-04-29
Adjust tank masses

Capkirk123:
Adjust masses of HP tanks to be closer to real life.


lpgagnon:
Can't speak to the 'real life' aspect; but tested a few configs, looking good. A basic sep-steel tim+wac loses ~3% dv; an Int-Al AJ-10-138 loses ~2% TLI payload; an Int-Al XASR-1 stage goes from being slightly better than an equal-mass GCRC to being slightly worse.

PR #2283, opened 2020-05-16
Heat Shield Config Restructure

DRVeyl:
Introduce a heat shield configuration tool into RO patching process.  Instead of every heat shield part, or command module with an integrated heatshield, needing to declare everything about that shield, it can set a few parameters and let an RO pass take care of the settings.  If it wants to customize the parameters, it can hook a later pass and adjust only the relevant bits.

New pass: RealismOverhaul_HeatShield
Main tool functionality is in new file RO_Heatshields.cfg
Flags / tags:
 * resetHeatShieldMass  - the tool will configure the part's mass if set
 * resetHeatShieldAblator - the tool will calculate and configure the maximum Ablator if set
 * heatShieldTag   - current valid values are LEO and Lunar
 * heatShieldDiameter - diameter (meters) if recalculating mass/ablator

LEO: Values correspond to the configurations for DRE's parts
Lunar: Values correspond to the configurations for stock heatshield parts.

--
Cleaned up a lot of the patching process for Squad and DRE and Restock parts.  Eliminated a lot of copy-paste (and some copy-paste errors).  Introducing the tool simplifies repeated configs of the same parameters like in those files.  Did this additionally for ProbesPlus_Thermal and some FASA capsules.

Updated configs to use the tool for various mods, including FASA, VSR, ProbesPlus, and some Tantares.  Enough to get going and demonstrate uses.

This tool is "opt-in": it does nothing unless the appropriate tags (above) are added to the PART.  The old patches are not affected / any less valid.

The tool sets the default temperature for the ModuleAblator display to 3000K for Lunar shields.  The default 2000K value is not particularly useful for parts with max skin temp ratings of 3600K expecting to return from beyond LEO.

I modified very few actual parameters.  Intent is not [yet] to rebalance the various parts, though it IS to make understanding the unique and extremely variable configurations easier to notice and understand, and to later apply better / more consistent rules to them.  Some ablator totals [mostly on heatshield parts] may have changed by +/- 5 through the use of the formula instead of the rounding by hand.  Most capsules have highly-variable ablator totals that didn't correspond to similarly-sized heatshields, and so keeping with "don't change things yet" I preserved their ablator quantity.

I tested these parts by examining the parts in the VAB.  I have not checked ReStock [due to current issues with its patch timing]


DRVeyl:
Reconsidering if `emissiveConstant` and `skinMassPerArea` should be set as part of the heatshield rule.  Started with it from the ubiquitous comment: `emissiveConstant = 0.6 // not too absorptive for reentry` splattered across many files, but it does seem more appropriately per model.

Capkirk123:
I've been looking into adding a 3rd class of heat shields, "hypervelocity" shields (name W.I.P.). They will be rated for 50 km/s reentries, for stuff like the Galileo Atmospheric probe (so you don't have to install RN US Probes to recreate it), or incredibly fast interplanetary trajectories in the inner solar system (Venus has a nasty habit of eating through even Lunar heatshields if you have an unusually fast transfer).

DRVeyl:
@Capkirk123 That would be easily doable by creating a new heatshield class in RO_Heatshields.cfg, and then tagging specific parts appropriately.

Validated these patches by comparing differences in MM cache before and after, in a build including ReStock, DRE and FASA.  Some accepted differences include:
- Smaller attachment node gizmos because of rescaling the overall heatshield part
- Not-as-pretty ablator / charred ablator quantities in the cache because we are not rounding our math in MM patches.
- Some very small mass differences (<1kg) for the same rounding reason

As this is not a balancing pass, few part stats changed.

PR #2285, opened 2020-05-17
Globally replace ignitions = -1 with ignitions = 0

Standecco:
From @jwvanderbeck's suggestion:
> tl;dr of the whole investigation was that 0 = infinite unless literalZeroIgnitions (or something similarly named) is true in which cases 0 = 0.  -1 will technically work but only in some situations and it was never intended for -1 to be used for infinite.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Athena.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Athena.cfg
@@ -3,7 +3,7 @@
 
 //  Athena Launch Vehicle Family:  https://www.sprsa.org/sites/default/files/conference-presentation/Athena%20Launch%20Vehicle%20Family-Kehrl_PR.pdf
 //  Space Launch Report - Athena:  http://www.spacelaunchreport.com/athena.html
-//  Norbert Brï¿½gge - Athena:       http://www.b14643.de/Spacerockets_2/United_States_2/Athena/Description/Frame.htm
+//  Norbert Brügge - Athena:       http://www.b14643.de/Spacerockets_2/United_States_2/Athena/Description/Frame.htm
 //  Propulsion Systems Data Sheet: https://www.rocket.com/files/aerojet/documents/Capabilities/PDFs/Propulsion%20System%20Data%20Sheets.pdf
 
 //  ==================================================
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Athena.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Athena.cfg
@@ -3,7 +3,7 @@
 
 //  Athena Launch Vehicle Family:  https://www.sprsa.org/sites/default/files/conference-presentation/Athena%20Launch%20Vehicle%20Family-Kehrl_PR.pdf
 //  Space Launch Report - Athena:  http://www.spacelaunchreport.com/athena.html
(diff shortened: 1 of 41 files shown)


jwvanderbeck:
@NathanKell

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Athena.cfg
```diff
@@ -3,7 +3,7 @@
 
 //  Athena Launch Vehicle Family:  https://www.sprsa.org/sites/default/files/conference-presentation/Athena%20Launch%20Vehicle%20Family-Kehrl_PR.pdf
 //  Space Launch Report - Athena:  http://www.spacelaunchreport.com/athena.html
-//  Norbert Br�gge - Athena:       http://www.b14643.de/Spacerockets_2/United_States_2/Athena/Description/Frame.htm
+//  Norbert Br�gge - Athena:       http://www.b14643.de/Spacerockets_2/United_States_2/Athena/Description/Frame.htm
```
jwvanderbeck, line 6: Stomping of unicode.  Egg would murder you.

jwvanderbeck:
Semi Obvious note: This is a large change and should be signed off by a few people as tested before it gets merged.

GameData/RealismOverhaul/RO_SuggestedMods/RealScaleBoosters/RO_RSB_Athena.cfg
Standecco: Fixed. Hope VSCode hasn't messed up anything else

jwvanderbeck:
> So I stated slightly wrong.. it really depends on what method of configs the engine is using
> however MOST engines use MEC I think?
> Hence my 0 is correct statement
> Changing them to 0 should be safe because the literalZero would have already been set or not set previously and not affected by this change
> I guess the caveat though is, are any of those engines proposed to change NOT using MEC?
> This is also probably why the confusion happened.. if there is an engine with just one config, and thus probably not using MEC, then the -1 works.  If it has multiple configs, and thus uses MEC, then -1 doesn't work.  It needs to be 0
> Are we all on the same loony train now

Capkirk123:
Ignitions shouldn't be set to 0 in ModuleEnginesRF, since it interprets 0 as 0, and -1 as infinite, although it doesn't really matter since ModuleEngineConfigs overwrites it anyway. All of the ModuleEnginesRF patches should simply be removed, as ModuleEngineConfigs does the same thing, and overwrites anything you put in ModuleEnginesRF.

derlaft:
@Capkirk123 could you please tell a little bit more details about what should be done?

I would love to help (because the issue still persists, for example ROENgines Raptor Vacuum has only 1 ignition, which makes 0 sense), but not sure what to do.

pap1723:
Is this still valid? Was it fixed with the other push for the engine fix?

Standecco:
I have no idea. The issue was that not all engines are using ModuleEngineConfigs (even though they should be), so this might break some engines. There wasn't much time put into testing this properly (or at all), so it's still uncertain.

DRVeyl:
I didn't close this issue earlier, but it's not valid when being done this generally, because most of the patches target various parts and not just the EngineConfigs.  Each part change we would need to confirm what it is patching.  (ModuleEngineConfigs would be correct, but ModuleEngines* would not.)

Standecco:
Could be a good use of my engine config parser; would there be any interest in turning it into a RP-0 part browser analog?
I'd only have to add de-serialization to it, should be doable.

pap1723:
I was thinking that using some of the same technology that the Parts Browser uses would make sense.

DRVeyl:
Going to close this now.  Moving an engine to 0 ignitions needs to be intentional.  I'm not the one to search the 41-ish parts to check, but individual changes can be merged in another PR if needed.

Issue #2237, opened 2020-03-17
Pheobus 2 Nuclear Engine Plume Detached


DegenerateGroundState:
The LV-NPT100 "Phoebus Two 100" nuclear engine configured in RO (I think by @Bornholio) has its plume detached quite a bit from the engine, see screenshot:

![plume_error - Copy](https://user-images.githubusercontent.com/23085959/76815301-24301200-67f5-11ea-8afc-a1099e636cfc.png)
Edit: Oh, and I just noticed the right click menu's title spells Phoebus incorrectly, which is why I spelled it incorrectly in the title of this issue.
`@title = LV-NPT100 "Pheobus Two 100" Atomic Rocket Motor` presumably should be 
`@title = LV-NPT100 "Phoebus Two 100" Atomic Rocket Motor`

DRVeyl:
![image](https://user-images.githubusercontent.com/25906300/83934018-79881980-a77b-11ea-8a6c-607ac0cf1806.png)

Bornholio:
"Phoebus"You are correct sir. sloppy work on my part.  Not sure why those NTR's settings have decayed so much they worked a few generations ago. notice that some are in the wrong nodes again also.

Snip from a development chart and baseline estimate for deployment of flight models if that had been developed (damn the politicians).  the "Number" at the end of the name is the effective expansion ratio of the nozzle
prototypeNuclearPropulsion | 59-60 | 1967 | KIWI A3
prototypeNuclearPropulsion | 59-60 | 1967 | KIWI A24
prototypeNuclearPropulsion | 60-64 | 1969 | KIWI B12
prototypeNuclearPropulsion | 60-64 | 1969 | KIWI B48
earlyNuclearPropulsion | 64-67 | 1972 | NERVA NRXA6
earlyNuclearPropulsion | 64-67 | 1972 | NERVA NRX
earlyNuclearPropulsion | 68-69 | 1974 | NERVA XE'
earlyNuclearPropulsion | 68-69 | 1974 | NERVA XE
earlyNuclearPropulsion | 68-69 | 1974 | design proposal 1961
basicNuclearPropulsion | Notional 70-73 | 1978 | NERVA I
earlyNuclearPropulsion | 65-67 | 1972 | Phoebus 1
basicNuclearPropulsion | 68-69 | 1981 | Phoebus 2
improvedNuclearPropulsion | Notional 71-75 | 1986 | NERVA II 
[...]

PR #2286, opened 2020-05-17
Update Nuclear Engines

Capkirk123:
Clean up the configs of all nuclear engines, in bring their stats more in line with available sources.


DRVeyl:
Does this fix https://github.com/KSP-RO/RealismOverhaul/issues/2237 ? (If so, can you mark that in the PR comment?)

Capkirk123:
It does not, it simply cleans up the configs and makes them more consistent.

Issue #2061, opened 2019-05-15
Coatl Aerospace battery balance


fat-lobyte:
Hi, it seems that the balance for Coatl Aerospace batteries is a bit off.

For reference, the stock Z-100 has an energy density of 1800kJ / 8kg = 225 kJ/kg.
The densities are inconsistent (CA300i seems especially wrong) and the help texts are a little (or a lot for CA300i) off.

Part | EC [kJ] | Weight [kg] | Energy Density [kJ/kg] | Orange Help Text
------------ | ------------- | ------------- | ------------- | -------------
CA25i  | 1600 | 5 | 320 | 440 Wh -> ‭1 584‬ kJ
CA100i | 2900 | 10 | 290 | 800 Wh -> 2880 kJ
CA300i | 1800 | 20 | 90 | 1500 Wh -> 5400 kJ

PR #2288, opened 2020-05-20
RTG Updates

DRVeyl:
Updated some patches for mostly RTGs from VSR, Coatl, SXT, and NearFutureSolar

Fixes https://github.com/KSP-RO/RealismOverhaul/issues/2061

Primarily the patches are the better ensure consistency in configs across parts.  Some bugfixes.  Transition from changing the model scaling and manually placing attachment nodes to using rescaleFactor and letting the system move the nodes.  That is, trust that the model maker puts them in the right place, or fixes it when they don't.


Capkirk123:
Tested in game, everything seems to work correctly.

Issue #2168, opened 2019-12-16
B9PS Serious Warnings on BDB Parts


Omnipius:
Current RO config for BDB parts results in a large number of errors. This has been reported to BDB, but they identified it as an issue with RO.

RO's BDB configs need to be updated to remove unused B9PartSwitch Modules and Subtypes. I've started work on this and eliminated errors on main part types. However, errors are still being thrown for subtypes.

Example warning:
[ERR 17:55:21.516] Initialization errors on ModuleB9PartSwitch (moduleID='engineSwitch') on part bluedog.Scout.Castor.Radial subtype 'CastorIVA'
  Could not find matching module

hargn:
Hi,
You can downgrad the B9PS version to 2.6.0 in order to fix the issues with BDB.
This version seems to properly work in 1.7.3 (need confirmation however).

Omnipius:
That does work. However, given that B9PS 2.6.0 is only officially compatible with KSP 1.6.1 and the 1.7.3 versions of BDB and B9PS being just fine without RO, I would call this solution a workaround at best. 

This remains an issue that the 1.7.3 version of RO is not actually compatible with the 1.7.3 versions of it's suggested mods.

blowfishpro:
Yeah, the BDB configs need to be updated.  A lot of the new functionality it’s complaining about is sort of ModuleEngineConfigs-lite which doesn’t make sense in RO anyway

DRVeyl:
Partially addressed by https://github.com/KSP-RO/RealismOverhaul/pull/2199
Though blindly removing the first ModuleB9PartSwitch is maybe not the best method.  May clean some of the patches up to do it in a more targeted manner for `#engineSwitch`.

Still leaves some modules for parts RO doesn't patch.  Slightly confused, tho, why BDB would point back to RO.  RO doesn't add the now-deprecated engineSwitch, it's in the original BDB configs.  I guess that's what @blowfishpro is saying.

PR #2289, opened 2020-05-21
BDB Engine Patch Fixes

DRVeyl:
Move BDB engine patching to FOR[RealismOverhaul] instead of legacy pass.
Add missing NEEDS[TacLifeSupport] to some Apollo parts.
Remove deprecated engineSwitch modules for ModuleB9PartSwitch
Remove left-over part tag meant to signal a different tech tree mod.  Should go away once BDB cleans up after itself, but in the meantime it spams the PartLoader section of the log with warnings.

Fixes https://github.com/KSP-RO/RealismOverhaul/issues/2168

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_Apollo_OLD.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_Apollo_OLD.cfg
@@ -183,7 +183,7 @@
 			maxAmount = 206.8
 		}
 	}
-	MODULE
+	MODULE:NEEDS[TacLifeSupport]
 	{
 		name = TacGenericConverter
 		%converterName = CO2 Scrubber
@@ -698,9 +698,7 @@
 {
 	%RSSROConfig = True
 	%rescaleFactor = 1.5646
-	
 
-	
 	@title = Apollo Service Module
 	%manufacturer = North American
 	@description = The Apollo Service Module that has a complete package of fuel, monoprop, batteries, fuel cells, and storage bays for your service module.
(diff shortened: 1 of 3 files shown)


Capkirk123:
Looks good for the most part. However, many BDB parts use B9partswitch to toggle between several different models/color schemes, and simply removing the B9partswitch module will cause all of these models/textures to load at once. The B9partswitch modules either need to be fixed so they no longer throw errors, or the extra models/textures need to be disabled through ModuleB9DisableTransform

DRVeyl:
The only B9 module I am patching out is engineSwitch.  Most of the RO configs previously were already removing it through a somewhat more awkward `!MODULE[ModuleB9PartSwitch],0 {}` which blindly hopes the first module is the one you want to remove.

An example engineSwitch config is <https://github.com/CobaltWolf/Bluedog-Design-Bureau/blob/master/Gamedata/Bluedog_DB/Parts/Titan/bluedog_LR87_LH2_V.cfg#L192-L274>.  My understanding from @blowfishpro 's comment in the linked issue is this is basically implementing a B9PS version of how RO does EngineConfigs, and we simply don't need the B9PS version of this.  As best I can tell it does not impact the models.

We still leave the B9PS for eg shrouds, as in <https://github.com/CobaltWolf/Bluedog-Design-Bureau/blob/master/Gamedata/Bluedog_DB/Parts/Titan/bluedog_LR87_LH2_V.cfg#L324-L343>

Is there an additional model-changing use of B9PS in BDB that we also need to track down?

Based on other comments / discussion elsewhere, I'll also remove the `@mass = ` patches in the parts that define `engineConfig`.

Capkirk123:
Many of the newer B9 engines switch the engine model when the engine configs is switched through B9PS. If these are left unculled, then all the model variants will load at once. Only a few engines use this technique (so far), so I’ll probably just check to see what parts need special treatment individually.

Capkirk123:
After some testing, it seems the your patch to remove engineSwitch works as intended, but some of the leftover individual patches for B9PS might cause issues. I would recommend removing all B9PS patches other than your engineSwitch patch

PR #2290, opened 2020-05-22
Remove kerbalCrewMass

lamont-granquist:
The way this is implemented in KSP generates bugs in mods.  Technically
this is solvable in mods but they all need to be updated to delay the
code they are calling to FashionablyLate or BetterLateThanNever by
calling FixedUpdateAdd to schedule the analysis after the part is
updated in Part.FixedUpdate().

This is a gigantic pain in the ass.

The Kerbalism author has also found other areas where its deeply
annoying:

https://github.com/Kerbalism/Kerbalism/blob/475866007fb94684336e959a67ae933dc64c8771/GameData/KerbalismConfig/Profiles/Default.cfg#L2396-L2405

It affects both KER and MJ at the very least, and as I'd consider the
way it is implmented in KSP itself to be active hostile to mod
development and all it does is troll users, with the standard
recommendation to users being to turn it off, lets just turn it off.

### DIFF
--- a/GameData/RealismOverhaul/RO_Physics.cfg
+++ b/GameData/RealismOverhaul/RO_Physics.cfg
@@ -1,8 +1,6 @@
 @PHYSICSGLOBALS:FOR[RealismOverhaul]
 {
 
-	%kerbalCrewMass = 0.1
-	
 	// AeroFX
 	@aeroFXDensityScalar1 = 0.00004
 	@aeroFXDensityExponent1 = 0.7


lamont-granquist:
Think the Kerbalism author also mentioned it creates buoyancy issues and kerbals with the double-mass immediately sink and drown?  Something like that.

Standecco:
> Think the Kerbalism author also mentioned it creates buoyancy issues and kerbals with the double-mass immediately sink and drown? Something like that.

that is indeed an issue, but I've already found 2 ways to solve it:
the first (which works flawlessly in my testing) is simply increasing the kerbals' buoyancy;
the second is changing the `initialMass` field in the MODULE[KerbalEVA], which also works but is a lot more hacky and gives very weird results when taken too far.

But even if the sinking issue can be solved easily, the Δv readout issues are much larger. If those are unfixable, then it's a strong argument against having mass. BUT you could also argue that if those do indeed cause missions to fail unexpectedly, then it means that 100kg per crew member are a non-insignificant part of the payload, and thus completely removing them would be unrealistic.

lpgagnon:
Supposing 100kg/crew is significant enough that it shouldn't be removed entirely, how about just bumping up the mass of capsules by a corresponding amount? It'd make non-full capsules too heavy, but that seems niche enough not to worry about.

AnticlockwisePropeller:
My only qualm about adding the 100kg to the base mass of capsules is whether or not station habs, etc. are included in that. They often contain capacity for quite a few kerbals, but would typically be launched empty. Therefore, simply writing a patch to add 100kg to any part that contains crew would not be ideal. Is there a station/capsule-specific tag that already exists, to make applying the extra mass easier? Doing it individually per-part would be a pain.

Also, sat in a grey area between station habs and capsules are the crew cabin parts, which can potentially be used as either, depending on circumstance. Should these have the mass included, or not?

jwvanderbeck:
Personally, I think the 100kg crew mass is pointless and if it causes so many problems just remove it.  I never care about it one way or the other in my own games.  If you are playing Sandbox then the mass is almost surely irrelevant, and in career by the time I'm lifting Kerbals I've already been lifting far heavier things.

Just remove it and be done with it.

lpgagnon:
Another small argument in favor of removal: the "double mass" doesn't just apply to kerbals walking around, it seems to also apply to kerbals sitting in an External Command Seat, increasing rover mass by ~200kg per passenger instead of ~100

PR #2291, opened 2020-05-22
ServiceBay: Update to new Squad part, add 3m version

fat-lobyte:
Uses the new version of the Squad Service bay parts.
Also adds a 3m version of the service bay. Cost, Unlock cost and Mass is scaled as r^3.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
@@ -825,15 +825,7 @@
 @PART[ServiceBay_125_v2]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	@MODEL
-	{
-		%scale = 1.6, 1.6, 1.6
-	}
-	%scale = 1.0
-	@node_stack_top2 	= 0.0,  0.33, 0.0, 0.0, -1.0, 0.0, 1
-	@node_stack_bottom2	= 0.0, -0.33, 0.0, 0.0,  1.0, 0.0, 1
-	@node_stack_top 	= 0.0,  0.41, 0.0, 0.0,  1.0, 0.0, 1
-	@node_stack_bottom 	= 0.0, -0.41, 0.0, 0.0, -1.0, 0.0, 1
+	%rescaleFactor = 1.6
 
 	@maxTemp = 1000
 	@title = Service Bay [2m]
@@ -845,19 +837,11 @@
 }
 
 
-+PART[ServiceBay_250_v2]:AFTER[RealismOverhaul]
++PART[ServiceBay_250_v2]:FOR[RealismOverhaul]
 {
 	@name = RO_ServiceBay_300
 	%RSSROConfig = True
-	@MODEL
-	{
-		%scale = 1.2, 1.2, 1.2
-	}
-	%scale = 1.0
-	@node_stack_top2 	= 0.0,  0.55, 0.0, 0.0, -1.0, 0.0, 1
-	@node_stack_bottom2	= 0.0, -0.55, 0.0, 0.0,  1.0, 0.0, 1
-	@node_stack_top 	= 0.0,  0.68, 0.0, 0.0,  1.0, 0.0, 2
-	@node_stack_bottom 	= 0.0, -0.68, 0.0, 0.0, -1.0, 0.0, 2
+	%rescaleFactor = 1.2
 
 	@maxTemp = 1000
 	@title = Service Bay [3m]
@@ -872,15 +856,7 @@
 @PART[ServiceBay_250_v2]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	@MODEL
-	{
-		%scale = 1.6, 1.6, 1.6
-	}
-	%scale = 1.0
-	@node_stack_top2 	= 0.0,  0.74, 0.0, 0.0, -1.0, 0.0, 1
-	@node_stack_bottom2	= 0.0, -0.74, 0.0, 0.0,  1.0, 0.0, 1
-	@node_stack_top 	= 0.0,  0.91, 0.0, 0.0,  1.0, 0.0, 2
-	@node_stack_bottom 	= 0.0, -0.91, 0.0, 0.0, -1.0, 0.0, 2
+	%rescaleFactor = 1.6
 
 	@maxTemp = 1000
 	@title = Service Bay [4m]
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
```diff
@@ -844,7 +844,32 @@
...
+	%RSSROConfig = True
+	@MODEL
+	{
+		%scale = 1.2, 1.2, 1.2
+	}
+	%scale = 1.0
+	@node_stack_top2 	= 0.0,  0.55, 0.0, 0.0, -1.0, 0.0, 1
+	@node_stack_bottom2	= 0.0, -0.55, 0.0, 0.0,  1.0, 0.0, 1
+	@node_stack_top 	= 0.0,  0.68, 0.0, 0.0,  1.0, 0.0, 2
+	@node_stack_bottom 	= 0.0, -0.68, 0.0, 0.0, -1.0, 0.0, 2
```
DRVeyl, line 860: Unless we have good reason to manipulate the individual model scale and the attachment nodes directly, which seems unlikely if we are copying a Squad part, please set the MODEL's own scale back to the default [or don't even update it from the default if it is 1.0 -- check the original part] and use `%rescaleFactor = 1.2`

fat-lobyte: I don't think that's a good idea. Note that every single other part in this file (and actually most repurposed Squad parts  in `RO_SuggestedMods/Squad/`) does the exact same thing by updating MODEL and not the rescaleFactor

When I deleted the MODEL update and only set `%rescaleFactor = 1.2`, the part became WAY too big. Possible [explanation ](https://forum.kerbalspaceprogram.com/index.php?/topic/70253-solved-0235-and-resizing-in-the-partscfg/&tab=comments#comment-1102263)from NathanKell:

> HOWEVER, there is a bug in scaling. For parts with MODEL nodes, rescaleFactor is applied twice to the mu, *unless* (a) the part is the root part of the vessel and (B) you've reverted to launch (or perhaps switched in flight? Don't recall). All other times, it is applied twice. What this means is that if you have scale = 1.0, 1.0, 1.0 in the MODEL node (or no scale at all), and rescaleFactor = 1.25, then your mesh will actually be 1.5625 as big, but your nodes will be scaled outwards 1.25.

fat-lobyte: Full [post about scaling](https://forum.kerbalspaceprogram.com/index.php?/topic/70253-solved-0235-and-resizing-in-the-partscfg/&tab=comments#comment-1102263) from NathanKell for posterity:


> Three things affect scale.
> 
> First, does the part use MODEL nodes? Skip to part B.
> 
> A (MODEL node not used; instead "mesh = blah" used)
> 
> 1. "scale = foo" determines how the first three numbers for each attach node line (the X, Y, Z coords of the node) are scaled. It does *not* affect anything else. Default = 1.0 if not present
> 
> 2. "rescaleFactor = bar" scales both the model (and its transforms) *and* the attach nodes. Default = 1.25 if not present.
> 
> B (MODEL node(s) used)
> 
> 1. Inside the MODEL node, "scale = blahX, blahY, blahZ" determines the scaling of the model and its transforms. Default is 1.0, 1.0, 1.0. It does not affect node positions.
> 
> 2. Outside the MODEL node, "scale = foo" works as above.
> 
> 3. Outside the MODEL node, rescaleFactor works as above.
> 
> HOWEVER, there is a bug in scaling. For parts with MODEL nodes, rescaleFactor is applied twice to the mu, *unless* (a) the part is the root part of the vessel and (B) you've reverted to launch (or perhaps switched in flight? Don't recall). All other times, it is applied twice. What this means is that if you have scale = 1.0, 1.0, 1.0 in the MODEL node (or no scale at all), and rescaleFactor = 1.25, then your mesh will actually be 1.5625 as big, but your nodes will be scaled outwards 1.25.
> 
> For this reason, when scaling parts using MODEL nodes, it is suggested to leave scale (outside the MODEL node) and rescaleFactor both at 1.0, and instead both change the scale = x, y, z inside the MODEL node and manually scale the positions of the attach nodes.

DRVeyl: I did exactly this reworking all of the heatshield parts, for both stock and DRE. And confirmed at least in the Editor getting the correct size. Including a ton of fighting with the 2.2m Soyuz part.

If you are rescaling a part that had already had it's MODEL scaled, it will of course not be correct. You may need to apply scale inside the MODEL back to 1.

fat-lobyte: I tried to do this:
```
	@MODEL
	{
		%scale = 1, 1, 1
	}
	%rescaleFactor = 1.2
```
But now the nodes are too far away and I have to figure out where to place them.

Is this really better than my patch? Like I said, all parts in that file do the same thing.

DRVeyl: Hm.  You're copying a part that has had its MODEL and attachment nodes manually moved already, before your copy, aren't you.  This is *the* example to me of why we shouldn't manually edit MODEL scale and attachment node position, in general.  Because you shouldn't have to figure out the node positions every time you want to do a simple rescaling.

fat-lobyte: > You're copying a part that has had its MODEL and attachment nodes manually moved already, before your copy, aren't you

Am I? The copy is stated above the modification. I thought that would modify it after the copy.

> Because you shouldn't have to figure out the node positions every time you want to do a simple rescaling.

But I have to do so anyway when using %rescaleFactor. If I remove the `@node_stack_*` lines, the nodes are wrong. 

To fix them, I had to install the NodeHelper mod, go into the editor and move the nodes around. THEN you have to divide the positions you see in the game by the `%rescaleFactor`.

DRVeyl: :Look at the patch order.

+PART[ServiceBay_250_v2]:**AFTER**[RealismOverhaul] {...}
@PART[ServiceBay_250_v2]:**FOR**[RealismOverhaul] {...}

DRVeyl: image

I get ^ by doing only:
```
@PART[ServiceBay_250_v2]:FINAL
{
	%title = Veyl Rescaled Service Bay 250_v2
	%rescaleFactor = 1.6
}
```

fat-lobyte: Fixed in next commit

PR #2295, opened 2020-05-30
Make early LR105-NA versions ground lit only

jwvanderbeck:
Manufacture specifications state the following:
1) Engine is designed to start at launch sites with an altitude below 10,000 feet
2) Engine requires ground support equipment to start, and said equipment is rated for a single start only before refurbishment


Capkirk123:
Looks good.

siimav:
I'm fine with that change as well.
On Discord it came up that maybe we should make X-405, RD-108 and maybe even the A-4 derivates also ground lit?

user1453:
Also LR79/89

DRVeyl:
Ended up conflicting this change with merging Kirk's mass of engine config formatting/organization PR.  Can you re-update?

Also, anyone want to take lead on if X-405 (not X-405H), RD-108, A4, LR79/LR89, RD107 should also be ground-lit only?

jwvanderbeck:
The conflict is trivial to resolve in the web editor, but requires it be commited after fixing.  Do you have a problem with that?

Otherwise i'd have to clone, resolve, and push the fix but that will take far more time that I don't have at the moment but could make time for later.

DRVeyl:
I'm in no hurry.  And it's a great opportunity for someone with more of a dog in the fight (and more expertise on the topic) to apply the same thing to a few more engines if it is particularly appropriate!

PR #2304, opened 2020-06-12
Don't set radial chute model if restock is there

lpgagnon:
(because it sets its own and we'd end up with both)


lpgagnon:
Main reason I went with NEEDS[!ReStock] is that RO already has several dozen instances of it.

Capkirk123:
Seems to work fine.

PR #2305, opened 2020-06-16
Bump up mass of restock lander cans

lpgagnon:
...to match the bump to non-restock versions from 83766f08832b69d8d347adec97ee4431b6a47781
(restock got forgotten)

not entirely clear that the higher mass is "more correct", but the variants should at least match


Capkirk123:
Looks correct, dry mass is set the same for both restock and non-restock versions of the generic 1-man and 2-man lander cans

DRVeyl:
Maybe hold off on this to avoid conflict with https://github.com/KSP-RO/RealismOverhaul/pull/2283

Capkirk123:
These are both Heat shield-less parts, so they should be safe.

DRVeyl:
@lpgagnon Can you re-base this onto the current RO master or otherwise resolve the conflict?

lpgagnon:
The deduplication of restock and non-restock parts in #2283 appears to have taken care of the issue; masses are now correct even without this PR. Closing.

PR #2306, opened 2020-06-16
Fix GEM-63XL thrust

pt1243:
According to [https://web.archive.org/web/20190825004201/https://www.ulalaunch.com/docs/default-source/evolution/190408_ulapanel_all_compressed.pdf](https://web.archive.org/web/20190825004201/https://www.ulalaunch.com/docs/default-source/evolution/190408_ulapanel_all_compressed.pdf), the GEM-63XL thrust should be 2026 kN.


Capkirk123:
Looks good, all sources seem to agree on 2026 kN

PR #2309, opened 2020-06-18
SCANsat update

Capkirk123:
Update SCANsat configs for the newest SCANsat release.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SCANSat/RO_SCANsat.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SCANSat/RO_SCANsat.cfg
@@ -81,6 +81,17 @@
 //	=================================================================================
 //	NEW PARTS
 //	=================================================================================
+//	sensorType keys: add to add sensors
+//	2^0: Lo-res radar
+//	2^1: Hi-res radar
+//	2^2: Lo-res Visual
+//	2^3: Biome
+//	2^4: Anomaly
+//	2^5: Anomaly Detail
+//	2^6: Hi-res Visual
+//	2^7: Lo-res Resource
+//	2^8: Hi-res resource
+//	=================================================================================
 
 //	Lo-res radar
 //	=================================================================================
@@ -183,6 +194,7 @@
 	rescaleFactor = 2.28
 	
 	@title = MODIS Multi-Spectral Imager
+	@description = An early multispectral scanner, used in the Aqua and Terra Earth Observation Satellites. It uses an array consisting of dozens of sensors sensitive to infrared, visible and ultraviolet light to monitor the biosphere of earth. These sensors allow it to create true-color maps of planets, and determine biome types.
 	@mass = 0.228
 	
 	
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SuggestedMods/SCANSat/RO_SCANsat.cfg
```diff
@@ -1,10 +1,17 @@
+//	=================================================================================
+//	DEPRICATED PARTS
```
DRVeyl, line 2: Spelling (DEPRECATED)

GameData/RealismOverhaul/RO_SuggestedMods/SCANSat/RO_SCANsat.cfg
```diff
@@ -53,6 +72,327 @@
...
+//	2^7: Lo-res Resource
+//	2^8: Hi-res resource
+//	=================================================================================
+
+//	Lo-res radar
+//	=================================================================================
+
+// Lo-res altimetry (regular radar), based on Ranger Block II radar
+// Source: https://books.google.com/books?id=51QCAAAAIAAJ&pg=PA1004&lpg=PA1004&dq=ranger+5+radar+altimeter&source=bl&ots=QSh4sZdgyJ&sig=ACfU3U2E6TgvRqYEcohEXOlM6CImanplMg&hl=en&sa=X&ved=2ahUKEwjf76q16onqAhWUHc0KHdQLAfsQ6AEwBXoECA8QAQ#v=onepage&q=ranger%205%20radar%20altimeter&f=false
+@PART[scansat-radar-poseidon-3b-1]:LAST[SCANsat]
```
DRVeyl, line 101: Does SCANSat do a lot of its own patching of its modules in :FOR[SCANsat] that we need/want to run in :LAST instead of just in a typical :FOR[RealismOverhaul] pass?

GameData/RealismOverhaul/RO_SuggestedMods/SCANSat/RO_SCANsat.cfg
```diff
@@ -53,6 +72,327 @@
...
+	
+	MODULE:NEEDS[RealAntennas]
+	{
+		name = ModuleRealAntenna
+		antennaDiameter = 3.7
+	}
+	
+	MODULE:NEEDS[RemoteTech]
+	{
+		name = ModuleRTAntenna]
```
DRVeyl, line 159: Typo in module name

GameData/RealismOverhaul/RO_SuggestedMods/SCANSat/RO_SCANsat.cfg
Capkirk123: RSS has a patch the forces all parts with a scansat module to some incredibly unhelpful defaults, and it runs AFTER[SCANsat]. Since RSS patches run after RO patches, this means only LAST and FINAL will make any difference. I have also opened a PR that moves the RSS patches to a reasonable time (FOR), but in case this PR gets merged first I kept it on LAST

siimav: I merged the RSS PR but we might still need to use the :Last pass since it isn't certain when and for which versions the next RSS release will be.

Capkirk123: Yep, my thoughts exactly

PR #2311, opened 2020-06-23
Update RO Part Modules (Build to 1.8.1+)

DRVeyl:
Compiled against 1.8.1.  Merging this would make the minimum supported version for RO 1.8.1+.

ModuleROSolarPanel:

- Enable direct input of days elapsed
- Automatically update on Ship change events
- Reduce overall planning duration to 20 years to make slider + direct input both useful
- Refactor display/string manipulation
- Remove inappropriate/unnecessary KSPField properties and Unity calls/callbacks

AdjustableCOMShifter:

- Add PAW groups
- Delay until OnStartFinished to allow other parts/PartModules to apply CoM offset
- Allow direct value input via UIFloatRange
- CoM display always available with Thermal Debugging, not just on change


siimav:
@raidernick Are you ok with moving RO support to only 1.8+

PR #2316, opened 2020-07-03
Resize sounding cores

Capkirk123:
Resize the aerobee and WAC science cores to more accurate sizes.

### DIFF
--- a/GameData/RealismOverhaul/Parts/Avionics/Sounding0-3m.cfg
+++ b/GameData/RealismOverhaul/Parts/Avionics/Sounding0-3m.cfg
@@ -10,37 +10,14 @@
 +PART[asasmodule1-2]:BEFORE[RealismOverhaul]
 {
 	@name = RP0probeSounding0-3m
-}
 
-@PART[RP0probeSounding0-3m]:FOR[RealismOverhaul]:NEEDS[!ReStock]
-{
-	!mesh = DEL
-	
-	!MODEL,*{}
-
-	MODEL
+	@DRAG_CUBE
 	{
-		model = Squad/Parts/Command/advancedSasModuleLarge/model
+		rescaleX = 1.0
+		rescaleY = 3.24
+		rescaleZ = 1.0
 	}
-
-	%rescaleFactor = 0.12
 }
-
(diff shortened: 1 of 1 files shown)


DRVeyl:
Can you look at extracting the VSR/Restock-specific elements?  (Can we actually observe the difference between a rescale of .1205 vs .12?)  Similarly, can you normalize the model scale to y=1 and then rely on rescaleFactor so that we can more easily rescale without needing to manually touch attachment nodes?  If we are declaring a part from scratch then we need the nodes right, but if we are copying any part that stock or some mod has made, it's on them to put the nodes where they belong.

Capkirk123:
I just wanna resize the parts man (yeah, I can use rescale factor instead)

GameData/RealismOverhaul/Parts/Avionics/Sounding0-3m.cfg
```diff
@@ -139,34 +118,28 @@
...
-	%node_stack_bottom = 0, -0.063158, 0, 0.0, -1.0, 0.0, 0
+	%rescaleFactor = 0.48336
 
-	@DRAG_CUBE
-	{
-		@rescaleX = 0.15
-		@rescaleY = 0.252632
-		@rescaleZ = 0.15
-	}
+	%rescaleCube = 0.48336
```
DRVeyl, line 142: `rescaleCube` doesn't do what you hoped.  It is just a flag for if the cube should get rescaled by the `rescaleX`/`rescaleY`/`rescaleZ` values inside the DRAG_CUBE node, it does not treat as if those individual values = the `rescaleCube` value.

Please either just delete the DRAG_CUBE node and allow the PartLoader to auto-generate one, OR to preserve the hollow-ness of the part, edit such that 
```
rescaleX = #$../MODEL/scale[0]$
rescaleY = #$../MODEL/scale[1]$
rescaleZ = #$../MODEL/scale[2]$
@rescaleX *= #$../rescaleFactor$
@rescaleY *= #$../rescaleFactor$
@rescaleZ *= #$../rescaleFactor$
```
I _think_ that will do the correct thing.

EagleEyeLogic:
Not sure what you are doing and how you have tested the changes...

For me with a simple RO installation the cores have 0.954m or 1m. Changing RO from master to the latest release solve this.

screenshot0

PR #2319, opened 2020-07-07
Replacing the XRS-2200 config

Its-Just-Luci:
I renamed the XRS-2200 to the RS-2200 in order to fit more with RO's naming scheme, in addition I added the RS-2200 config as the default, and made the XRS-2200 the second config.


Capkirk123:
This now conflicts. It probably also needs to be redone to match the engine config template here: https://github.com/KSP-RO/RealismOverhaul/pull/2310

DRVeyl:
Should we be looking at this PR or #2320 ?  If you can modify it to match the template that Kirk has made, will appreciate it.

pap1723:
@Its-Just-Luci @Capkirk123 any movement on this?

PR #2318, opened 2020-07-06
Update Stock Command/Guidance/Spaceplane Patches

DRVeyl:
Merge ReStock and Stock patching
Update and organize GuidanceRings patches
Organize patches by part, INTERNAL and RemoteTech compatibiltiy segments
Apply common patches commonly
Use rescaleFactor instead of direct model and attachnode manipulation
Organize TAC-LS patching and apply consistently instead of copy-and-paste
Some bug-fixes regarding avilable tank volume
Minor TAC-LS rebalancing of LiOH, CO2 supplies.  (CO2 storage volume was all over the place.  LiOH also had some variance especially around small cockpits.)
TODO: Unify TACLS-related resource supply patching across Command and Spaceplane files -- maybe in the context of ROKerbalism

PR #2323, opened 2020-07-13
IVA tweaks: Mk1 Crew Carrier, MediumCan seating

lpgagnon:
- generate the Medium IVA even if restock is there, the
  landerCabinMedium needs it (no idea why it was this way)
- remove ALL IVAs from landerCabinMedium before adding the medium
  one, to catch the RPM variant
- use the Medium IVA for the 3-seat Mk1 Crew Carrier, instead of
  leaving the 1-seat Small IVA
- add an extra seat to the medium IVA
- do NOT use made-up names for seatTransforms; that makes the
  seat positioning be relative to the root part
  (original commit makes it clear this was known not to work right)
- enlarge the Medium IVA to more or less fill both parts;
  arrange extra seats in a plausible way

Main motivation was adding the missing seats to the Crew Carrier;
it seems Ship Manifest's crew transfer (and EVA?) breaks when there's
no free seat in the target pod.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/VSR/Crew.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/VSR/Crew.cfg
@@ -264,6 +264,11 @@
 			DumpExcess = False
 		}
 	}
+        !INTERNAL[*] {}
+        INTERNAL
+        {
+            name = landerCabinMediumInternal 
+        }
 }
 
 //  ==================================================


DRVeyl:
Some of this might be duplicative of <https://github.com/KSP-RO/RealismOverhaul/pull/2318>.
The VSR part is unique. I was just starting to look at those patches, also, thinking of pulling apart the VNP portion from VSR.  (And someone brought up the whole "can't detect VNP versus VSR" problem that leads to, among other things, the aerobee core scaling being wrong, so I need to fix that also.)

lpgagnon:
not sure about duplicative, but at the very least the two PRs will conflict horribly

DRVeyl:
Please fix the conflict, and I'll merge the remaining portion affecting VSR.

lpgagnon:
push -f: rebased, stripped out everything except the VSR change. Confirmed the other PR took care of the rest, this is now all that's needed to give the crew carrier its 3 seats

PR #2325, opened 2020-07-15
More consistent diameter/mass ratios for fairing bases

lpgagnon:
Give all payload and interstage bases the same coefficients,
instead of making it sort of mass how the model looks

Main impact is actually to apply consistent formulas to new-ish
(SSTU-derived) bases that weren't handled by RO, and ended up
lighter when smaller, but heavier when larger, than the "old" bases.

(also removed some 'costPerTonne' values; RO doesn't care about cost)


lpgagnon:
push -f: a better selector for boattail bases

PR #2329, opened 2020-07-23
Make early first stage engines ground lit only

leudaimon:
The first commit includes the A-4, RD-100-103M, LR79/89, RD-107/8 and X-405.

Any others I should include?


Capkirk123:
LR87-AJ-3 (although none of the early LR87s could airlight, the AJ-3 was the only kerolox version)

RS-68 (It needs a substantial supply of Helium and support from ground equipment)

RS-25 (same as above)

XLR41, XLR43 and NAA-75-110 (All used a pyrotechnic starting system copied from the A-4)

Issue #1955, opened 2018-12-11
Aerobee engines mass/TWR seems too good to be true


leudaimon:
https://github.com/KSP-RO/RealismOverhaul/blob/3b7167693b13ff325e4c6536d69611181f73ca2a/GameData/RealismOverhaul/Engine_Configs/Aerobee_Config.cfg#L13
I couldn't find any sources for the mass of these engines, but the current values seem off. 
I have no idea how that never came to my attention, but TWR of these engines is 97, 135 and 180 for the WAC, XASR-1 and AJ-27 respectively in the current config. 
This would be crazy good for a modern engine, and impossible for a Post War one. If no source for the mass is available, I was thinking about applying the AJ10-37 TWR to all of them, as that one seems realistic (TWR = 41) and it would be easy to adjust the mass based on the respective thrusts.

pianojosh:
This is a good point, WACs in-game vastly outperform real world missions.

That being said, given how tight funds can be early-game, it might be necessary to completely reevaluate early sounding rocket contract requirements and/or rewards with this change.

leudaimon:
I'm not sure this would have such a big effect in the early career. I made a test with the following changes:
* Instead of the uniform TWR I suggested above, I made it so that the engines kept the same mass relationship among them as the current config does (1, 1.3, 1.5 massMult for the 3 engines in order)
* I gave the XASR-1 the same TWR as the AJ-37. My thought is that as the AJ-37 has gimbaling and these engines don't, it could be possible for the AJ-27 to have better TWR than the AJ-37
* I achieved this by making the standard mass of the part = 0.027kg and all else equal

I then built a WAC-Corporal, under non-optimal conditions:
* 0.3m diameter
* empty (fairly large) nose cone with Tank-I HP type
* main body also Tank-I HP with 80% utilization
* 55s burn time
* dry mass of the whole rocket was 136kg

This rocket was able to break the Karmán line, even though it couldn't get to in-game space for science.

I think this seems reasonable and shouldn't hurt the early career. It may actually make it more challenging to get the space science early, which might be interesting.

NathanKell:
Well, that's a screwup on my part. :)
I'm not sure I ever had a mass value for them, so I might have just invented the 8kg value--I think there was maybe a comment to that effect at one point. The only concern I would have here is aerodynamic stability of (newb-) created WACs, making the tail heavier could lead to issues.

leudaimon:
[...]

PR #2335, opened 2020-08-02
Aerobee Mass Adjustment

Capkirk123:
Taking action on https://github.com/KSP-RO/RealismOverhaul/issues/1955. I've gone with the conservative value of 16 kg base (double the previous mass), assuming that the lack of a nozzle extension and gimbals will make up for the slightly lower chamber pressure of the aerobee vs. the AJ10-37. If people believe @leudaimon's original guess of 27 kg is better, it can be changed to that instead.


siimav:
16kg sounds reasonable and is more in line with Bornholio's findings.

leudaimon:
Yes, 16kg sounds very reasonable. My 27kg guess was very gamey.

PR #2337, opened 2020-08-04
Stretch SXT skylab

lpgagnon:
Because 7m is much shorter than the real thing.
Stretched model looks ok, except for a very tall hatch

A bit dubious, but if we're going to pretend it's skylab, making it the right shape is more important than how pretty it is?


pap1723:
This is an ugly hack fix, but it is more accurate, so let's go with it.

PR #2346, opened 2020-08-15
Re-enable long-lost derwent alternator

lpgagnon:
Whatever _the heck_ was going wrong 5 years ago does not appear to be going wrong anymore; alternator is working fine in my tests, and nothing bad seemed to happen.


DRVeyl:
RFC: The current output with this is 500W.  RO's history on this file never adjusted the rate, but RO does set the rate on eg the Bonanza [`SXTTinyprop`] to 200W.  No issue with turning this back on [as I assume all ModuleAlternators are not still broken 5 years later], but would like someone knowledgeable in this area to comment on the settings.

PR #2351, opened 2020-08-18
Don't give 2-axis gimbal to single RD-8 vernier

lpgagnon:


raidernick:
Gimbals for verniers should be set up on a part by part basis and the global configs shouldn't touch them at all. Every part has its axis set up differently, if you flip the x and y what you have here will break things. Gimbals and which axis they move on should always be done in the config for that specific part as every part is set up differently.

lpgagnon:
Does that mean everything in this file that configures gimbals should be removed? The top-level ModuleGimbal and the CONFIG's gimbalRange?

My concern would be that I can and will add a "correct" ModuleGimbal to the ROEngines RD-8 part config, but it may leave any other mod's RD-8 with a broken gimbal

Capkirk123:
EngineConfig should only set GimbalRange. Everything else should be set in the part itself

lpgagnon:
EngineConfig setting GimbalRange is what's causing a problem. It causes the ModuleGimbal's single-axis gimbal to get overridden by a dual-axis gimbal

Capkirk123:
That can be overwritten by patching the gimbal :AFTER[RealismOverhaulEngines], which is the suggested method for doing these things.

PR #2357, opened 2020-08-23
Don't delete RD-107 vernier gimbals

DocRockwell:
!MODULE[ModuleGimbal],*{} stops gimbaling verniers from working on ROEngines rd107 (probably others based on how this is implemented). Removing this line restores functionality.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD107_RD117_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD107_RD117_Config.cfg
@@ -186,7 +186,7 @@
 	!RESOURCE,*{}
 
 	//Verniers need to be set up on a part-by-part basis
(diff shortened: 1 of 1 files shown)


Capkirk123:
As the comment suggests, verniers need to be set up on a part-by-part basis.

DocRockwell:
"As the comment suggests, verniers need to be set up on a part-by-part basis."

They're already set up in the ROE RD107 part config, why remove the module only to have to reapply it? Why is the RD108 engine config different in this regard and doesn't strip the ModuleGimbal?

Capkirk123:
The Verniers should be set up in a patch :AFTER[RealismOverhaulEngines] to prevent the engine config from messing with them. In the case of the RD107/108, they only have verniers so it doesn't cause an issue, although I would prefer the engineConfig strip the gimbals and then let the parts set them up afterwards.

PR #2360, opened 2020-08-27
Bring back dewar flasks in SM tanks

lpgagnon:
Applied to same resources as in the ServiceModule tankdefinition
(in my MM cache)

Tested by putting a LOX SM tank in orbit; heat penetration goes from watts to milliwatts

The question "why do SMs not reduce boiloff anymore?" comes up every once in a while but never goes anywhere; maybe a PR will get more attention

### DIFF
--- a/GameData/RealismOverhaul/RO_RealFuels.cfg
+++ b/GameData/RealismOverhaul/RO_RealFuels.cfg
@@ -1443,6 +1443,23 @@ TANK_DEFINITION
 		maxAmount = 0.0
 		//note = (pressurized)
 	}
+
+	// Cryo resources in SM tanks get a dewar flask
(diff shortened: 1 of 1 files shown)


pap1723:
@MikeOnTea you had said there was some issue with isDewar working properly.

@lpgagnon have you seen that this is in fact working better?

lpgagnon:
I didn't test extensively, but it did appear to work as expected.
> Tested by putting a LOX SM tank in orbit; heat penetration goes from watts [without this change] to milliwatts [with this change]

lpgagnon:
https://discord.com/channels/319857228905447436/531994741378646016/738112477782343690
tweaking the 'insulation' values might also be appropriate

PR #2364, opened 2020-09-17
Rutherford fixes

vader111:
Change & Fix up Rutherford Engine Configs with Latest data available from Rocketlabs


vader111:
This was quite hard to do since even rocket labs website has a lot of contradicting on the engine

vader111:
Most of the available info is based on the prototype engine, and the engine has since gone through a lot of changes since then

PR #2366, opened 2020-09-22
Heating & heatshield changes

siimav:
* Unlock the full convective heating in FAR. Previously exposed area was clamped to some arbitrary percentage because that is what stock code also did for whatever reason.
* Make 1 unit of LEO heatshield ablator a bit more effective to account for the increased thermal loads.
* Reduce the max skin temperature on LEO heatshields to make them explode when subjected to very high peak heat loads.
* Rework Lunar heatshields such that they actually ablate and keep temperatures at levels more similar to LEO shields.
* Reduce conduction for both LEO and Lunar heatshields to give parts with low thermal masses that are directly attached to the HS a better chance of survival.
* Slightly increase the skin temperature limits for Early B9 wings to account for the increased thermal loads.
* Increase the mass multiplier on Early B9 wings to make them more beginner-friendly. Higher thermal mass is helpful for sounding rockets where the fins are subject to short but high thermal loads. Another benefit is that it's harder to snap the wings off from planes with high-G maneuvers.

### DIFF
--- a/GameData/RealismOverhaul/RO_DependentMods/RO_FAR.cfg
+++ b/GameData/RealismOverhaul/RO_DependentMods/RO_FAR.cfg
@@ -1,0 +1,7 @@
+@FARConfig:FOR[RealismOverhaul]
+{
+	@Settings
+	{
+		%exposedAreaLimited = false
+	}
+}
--- a/GameData/RealismOverhaul/RO_Heatshields.cfg
+++ b/GameData/RealismOverhaul/RO_Heatshields.cfg
@@ -28,22 +28,20 @@
 		ablativeResource = Ablator
 		outputResource = CharredAblator
 		outputMult = 0.75
-		lossExp = -40000
-		lossConst = 15000
-		pyrolysisLossFactor = 40000
-		ablationTempThresh = 500
-		reentryConductivity = 0.01
+		lossExp = -25000
+		lossConst = 350
+		pyrolysisLossFactor = 30000
+		ablationTempThresh = 1250
+		reentryConductivity = 0.0025
 		infoTemp = 3000
-		//reentryConductivity = 0.12
-		//@reentryConductivity = #$../heatConductivity$ // if it exists, use it
 	}
 }
 
 // Apply physics properties for LEO-rated shield
 @PART:HAS[#heatShieldTag[LEO]]:FOR[RealismOverhaul_HeatShield]
 {
-	%maxTemp = 1500
-	%skinMaxTemp = 2600
+	%maxTemp = 1300
+	%skinMaxTemp = 2350
 	%heatShieldDensity = 0.01
 	%heatShieldAblator = 50
 
@@ -54,12 +52,10 @@
 		outputResource = CharredAblator
 		outputMult = 1
 		lossExp = -6000
-		lossConst = 0.13
-		pyrolysisLossFactor = 6000
-		ablationTempThresh = 500
-		reentryConductivity = 0.01
-		//reentryConductivity = 0.12
-		//@reentryConductivity = #$../heatConductivity$ // if it exists, use it
+		lossConst = 0.095
+		pyrolysisLossFactor = 8210
+		ablationTempThresh = 750
+		reentryConductivity = 0.0065
 	}
 }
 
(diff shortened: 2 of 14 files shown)


siimav:
According to "Entry Performance of the Mercury Spacecraft Heatshield", ~13 lbs of mass was lost during re-entry. Our LEO heatshields currently do not lose any mass at all.

siimav:
Collected some more interesting stats from *Coming Home: Reentry and Recovery from Space*.
> The Gemini heat shield weighed 347 pounds, compared to the Mercury heat
shield’s 303 pounds, mostly because of a 48-percent increase in area, a 25-percent increase in the ballistic loading coefficient, and a 90-percent increase in the design heat load (per square foot) due to the lifting reentry

About Gemini GT-3 flight:
> After the heat shield was removed from the spacecraft and dried, engineers noted
it had lost about 20 pounds due to ablation, which was about average for the
subsequent missions.

Gemini-B test flight:
> For the heat shield qualification (HSQ) flight, the RF-300 insulation used
between the NASA heat shield and the capsule structure was not included
since the reentry was expected to be more benign than the standard Gemini
mission. In fact, the backside of the heat shield reached only 133 °F on the
HSQ flight, compared to the Gemini’s 430 °F design limit

On Apollo:
> Despite the information provided by Project FIRE, researchers believed that
large unknowns remained concerning superorbital reentry, so NASA, North
American, and Avco conservatively overdesigned the heat shield. The heat
shield weighed almost 1,500 pounds, over 10 percent of the entire commandmodule weight. In fact, none of the lunar-return missions used even 20 percent
of the available ablator, and engineers later decided they could have cut the heat
shield thickness in half and still have had an adequate margin

Bornholio:
**Looks good**, We need TL or "Modern Tier" PICA, PICA-X, NEXTEL and AVCOAT.
Also a set of Walnut and Oak versions. Pretty sure the nose of my local nuclear missile is Walnut :)

siimav:
### TODO list for this task:
* Separate LEO preset into Mercury and Gemini
  * duplicate the LEO HS parts also into two tiers?
* Since the Lunar HS is based on Apollo - and Apollo used up only ~20% of the available ablator - buff up the Lunar ablator again

NathanKell:
"Our LEO heatshields currently do not lose any mass at all." <<< I thought we had a 1:1 volume conversion between Ablator and CharredAblator, and CharredAblator's density is lower than Ablator's, so mass is lost?

Also, while it's more work, it would be great if we made three distinctions here:
* Early LEO heat shields that operated in the Heat Sink approach (i.e. Discoverer/Corona film return buckets) -- this approach was also used for Mercury's suborbital heat shield. They required very steep reentries to keep heat loads low, but did not require any ablator.
* "Normal" LEO heatshields unlocked by human-rated EDL. What we're used to.
* Gemini's lifting-rated heat shield (barely capable of lunar return).

NathanKell:
What are the blockers for merging this down and shipping with the next RO? Tuning lunar? Checking spaceplanes?

siimav:
TODO list:
* Tune Lunar ablator to be more in line with Apollo. This means increasing the amount of energy each unit of ablator deletes.
* Figure out how to approach the Mercury-Gemini situation. Currently the LEO preset lies somewhere between the two. It's a bit too good for Mercury but cannot quite sustain the lifting re-entry of Gemini. 
Options are:
  * Balance the preset for Gemini
  * Split it into Mercury and Gemini presets. This would almost certainly require duplicating all the LEO HS parts though.

siimav:
There are separate Mercury and Gemini presets now. Lunar preset got buffed enough that Apollo would have at least 25% of the ablator left after a typical Lunar re-entry.
My suggestion is that all the generic HS parts in RO will use the Gemini preset instead of LEO. Capsule-specific parts can still be configured with Mercury and EarlyLunar presets as well. For RP-1 we would remove all the generic parts and have only capsule-specific parts or have the player use ROHeatshields mod for custom sizes. This would allow us to create proper tech progression without having an enormous amount of duplicate parts.

NathanKell:
That seems fine to me! I thought Apollo was only supposed to use about 30% of its ablator, not 75% though?

Does the Gemini preset work for a natural-decay Vostok orbit? (Or the Mercury preset but tons of ablator?)

I'm going to have a poke at a heat sink style heat shield.

siimav:
> That seems fine to me! I thought Apollo was only supposed to use about 30% of its ablator, not 75% though?

Oops, I meant to say that it should use up no more than 25% of the available ablator

PR #2370, opened 2020-10-17
Make rcs-fueled generic thrusters pressure-fed and instant-on

lpgagnon:
ROEngines thrusters used to be pressure-fed but no longer are due to side-effect of some cleanup; RO's own generics
never were.

ROEngines thrusters also stopped being instant-on at the same time.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RO_RCS_Config.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RO_RCS_Config.cfg
@@ -103,6 +103,18 @@
 	}
 }
 
+// Make all rcs-fueled Engines pressure-fed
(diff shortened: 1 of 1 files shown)


NathanKell:
Does this actually do anything? I don't recall there being a ModuleRCSRF like ModuleEnginesRF to support this in practice. Has that been added since?

lpgagnon:
These aren't RCS blocks, they're engines that use the same fuels as rcs blocks. They have a ModuleEnginesRF.

NathanKell:
I am a tired idiot. Sorry!

NathanKell:
Anyway, this seems fine to me. It'll get ignored by RCS and used by Engines.

lpgagnon:
feature creep! PR now also explicitly makes generic thrusters instant-on, instead of relying on the PART for it. (mostly because ROE's no longer take care of it)

PR #2372, opened 2020-10-29
Add a RSRM part based on stock's kickback

lpgagnon:
Because it's weird for RO to provide a "default" RSRMV but no RSRM.
Only tested with restock.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_NASA.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_NASA.cfg
@@ -74,6 +74,38 @@
 	}
 }
 
+// Shuttle SRM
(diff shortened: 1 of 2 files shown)


DRVeyl:
Any interest in sorting this out for VSR and stock-only models?

lpgagnon:
how about we pretend I didn't write the TODO, nobody noticed there might be extra work needed for non-restock installs, this gets merged, and _then_ somebody complains about whatever turns out to be broken and it gets fixed?

StonesmileGit:
Tested with VSR; pair of RSRM on the left, RSRMV on the right.
VSR to the left in each pair, reDIRECT to the right in each pair
image

lpgagnon:
"TODO try with ven's" removed, stonesmile's comparison looks close enough to me. Good to merge?

StonesmileGit:
LGTM, will hold until my current work on the RSRM is done

PR #2373, opened 2020-10-30
Fix Kerbal Scaling for the Mk2 Lander Can IVA

Glloyd:
As above.

Before: 

After:

There is also no IVA for the third kerbal in the can, but that would likely require a new IVA, which is beyond my abilities.


lpgagnon:
If you want to go the extra mile and try adding a third (invisible) seat, there's an example in this PR https://github.com/KSP-RO/RealismOverhaul/pull/2323

PR #2376, opened 2020-11-07
Fix up a bunch of issues.

vader111:
Fix up RCS Audio, Fixed Spelling Mistake In RO_Squad_Communication.cfg, Disable SLS Tank in RO_Squad_NASA.cfg if Restock is installed, Add a patch to remove Restock Features that don't work with RO.

Add patches to Remove the KerbNet module from all parts & Remove the test subject module from all parts.

### DIFF
--- a/GameData/RealismOverhaul/RO_Comms.cfg
+++ b/GameData/RealismOverhaul/RO_Comms.cfg
@@ -24,22 +24,3 @@
 		antennaCombinable = false
 	}	
 }
-
-//  ==================================================
-//  Cleanup.
-//  ==================================================
-
-@PART:HAS[#useAntennaConfig[*]]:FOR[zzzRealismOverhaul]
-{
-	!useAntennaConfig = DEL
-}
-
-//  ==================================================
-//  Remove All RemoteTech Related Configs.
-//  ==================================================
-
-@PART:[*]:FOR[zzzRealismOverhaul]
-{
-    !MODULE[ModuleRTAntenna*],{}
-    !MODULE[ModuleSPU*],*{}
-}
--- a/GameData/RealismOverhaul/RO_Comms.cfg
+++ b/GameData/RealismOverhaul/RO_Comms.cfg
@@ -33,3 +33,13 @@
 {
 	!useAntennaConfig = DEL
 }
+
+//  ==================================================
+//  Remove All RemoteTech Related Configs.
+//  ==================================================
+
+@PART:[*]:FOR[zzzRealismOverhaul]
+{
+    !MODULE[ModuleRTAntenna*],{}
+    !MODULE[ModuleSPU*],*{}
+}
--- a/GameData/RealismOverhaul/RO_Comms.cfg
+++ b/GameData/RealismOverhaul/RO_Comms.cfg
@@ -24,3 +24,12 @@
 		antennaCombinable = false
 	}	
 }
+
+//  ==================================================
+//  Cleanup.
+//  ==================================================
+
+@PART:HAS[#useAntennaConfig[*]]:FOR[zzzRealismOverhaul]
+{
+	!useAntennaConfig = DEL
+}
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCSFix.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCSFix.cfg
@@ -1,0 +1,4 @@
+@PART[*]:HAS[@MODULE[ModuleRestockDepthMask]]:NEEDS[ReStock]:FOR[zzzRealismOverhaul]
+{
+	!MODULE[ModuleRestockDepthMask],*{}
+}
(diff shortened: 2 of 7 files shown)


GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCSFix.cfg
```diff
@@ -0,0 +1,4 @@
+@PART[*]:HAS[@MODULE[ModuleRestockDepthMask]]:NEEDS[ReStock]:FOR[zzzRealismOverhaul]
+{
+	!MODULE[ModuleRestockDepthMask],*{}
```
Standecco, line 3: Is there a reason for this? It's a cool feature, why do we want to remove it?

vader111: because it spams null refs

Standecco: Could you post one of these null refs?

vader111: i would have to remove it and relaunch ksp, so give me a bit

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_NASA.cfg
```diff
@@ -237,7 +237,7 @@
 	@description = Engine core for the Space Launch System (SLS). Features four Rocketdyne RS-25D/E engines. [8.4 m]
 }
 
-@PART[Size3LargeTank]:FOR[RealismOverhaul]
+@PART[Size3LargeTank]:FOR[RealismOverhaul]:NEEDS[!ReStock]
```
DRVeyl, line 240: What is the intent here?  If ReStock is installed, then this tank will not be RO-configured at all.  If we don't want ReStock to touch Size3LargeTank because we have a better or exc,usive use for it, we should just black-list it for ReStock, instead of worrying about what they modify.

vader111: when you have restock installed this tank has invisible tanks and no collider to click to pick up.

DRVeyl: ReStock *also* employs the model-construction technique we use.  So `!MODEL {}` is insufficient to deal with all the MODELs added in <https://github.com/PorktoberRevolution/ReStocked/blob/master/Distribution/Restock/GameData/ReStock/Patches/FuelTank/restock-fueltanks-375.cfg#L128-L150>

Does ReStocked Size3LFB represent an SLS?  If so, we should allow what they do, apply appropriate rescaling, and then set our mass/etc in a separate patch.  If not, we should blacklist them touching this part with:
`@PART[Size3LargeTank]:NEEDS[ReStock]:FIRST { %RestockIgnore = True }`

vader111: "Does ReStocked Size3LFB represent an SLS?" Not sure.

vader111:
I have just added what you recommended until someone can sit down and look at it in detail

GameData/RealismOverhaul/RO_Comms.cfg
```diff
@@ -33,3 +33,13 @@
...
+
+//  ==================================================
+//  Remove All RemoteTech Related Configs.
+//  ==================================================
+
+@PART:[*]:FOR[zzzRealismOverhaul]
+{
+    !MODULE[ModuleRTAntenna*],{}
+    !MODULE[ModuleSPU*],*{}
+}
```
DRVeyl, line 45: Revert this, please.  While there does not appear much interest in maintaining RT patches in the current RO community, we should not delete support.

GameData/RealismOverhaul/RealismOverhaul_Global_Config.cfg
```diff
@@ -331,3 +331,17 @@
 		AllowSignalRelay = False
 	}
 }
+
+//  ==================================================
+//  Remove the KerbNet and Test Subject modules from all parts
+//  ==================================================
```
lpgagnon, line 337: why?
With scansat's RSS compatibility in a questionable state and ROKerbalism not providing scanners of its own, killing kerbnet doesn't seem like a good idea.

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_NASA.cfg
```diff
@@ -114,7 +114,7 @@
 	engineType = RSRMV
 }
 
-@PART[Size2LFB]:FOR[RealismOverhaul]:NEEDS[!ReStock]
+@PART[Size2LFB]:NEEDS[ReStock]:FIRST { %RestockIgnore = True }
```
DRVeyl, line 117: Why are we also changing this part?  The previous patch allowed ReStock to do its thing, or if you did not have ReStock then RO makes Size2LFB into something constructed

I understand changing Size3LargeTank because RO [apparently] wants to construct this model and needed to tell ReStock to stay away.  Possibly a mistake if what ReStock is making is also SLS-like and we should only rescale their work.  But I'll put that off as a problem for later.

vader111: @DRVeyl so keep the :FOR[RealismOverhaul]:NEEDS[!ReStock]

PR #2377, opened 2020-11-07
RL10 & BE-4 Updates

vader111:
Add BE-4 Config & Fix up RL10 Config.


vader111:
Also Needs https://github.com/KSP-RO/RP-0/pull/1406

DRVeyl (commented):
I'm somewhat concerned with destroying the RL10A-4-1-2 config altogether.  Any parts that were configured with that will become broken.  I think the config has to continue to exist, at least with some deprecation note.

GameData/RealismOverhaul/Engine_Configs/BE-4_Config.cfg
```diff
@@ -37,7 +38,7 @@
 //	* No TestFlight compatibility since this global engine config represents unflown hardware.
 //	==================================================
 
-@PART[*]:HAS[#engineType[BE4]]:FOR[RealismOverhaulEngines]
+@PART[*]:HAS[#engineType[BE-4]]:FOR[RealismOverhaulEngines]
```
DRVeyl, line 41: If you're going to rename the engine type with a -, you must find any other engines that used the old reference and update them.
Hint: RO_VNP_Engines.cfg has one.  So is a KK launcher.

DRVeyl (changes requested):

GameData/RealismOverhaul/Engine_Configs/BE-4_Config.cfg
```diff
@@ -70,38 +71,32 @@
 		CONFIG
 		{
 			name = BE-4
```
DRVeyl, line 73: Need to fix it here, too.

vader111: i never changed that

GameData/RealismOverhaul/Engine_Configs/BE-4_Config.cfg
```diff
@@ -111,7 +106,7 @@
 //	TestFlight compatibility.
 //	==================================================
 
-@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[BE-4]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
+@PART[*]:HAS[@MODULE[ModuleEngineConfigs]:HAS[@CONFIG[BE4]],!MODULE[TestFlightInterop]]:BEFORE[zTestFlight]
```
DRVeyl, line 109: Appears I meant here.  Was briefly confused.  But the `engineType` is BE4, and the Config is `BE-4`.  So this line should be reverted back to BE-4.

vader111: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/VSR/VNP/RO_VNP_Engines.cfg#L340
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/KK%20Launchers/RO_KK_Vulcan_Engines_ULA.cfg#L112

DRVeyl: Those set `engineType`.  You are writing TestFlight values for `CONFIG`.  One has a -, the other doesn't.

vader111: oh ok i get it

GameData/RealismOverhaul/Engine_Configs/BE-4_Config.cfg
```diff
@@ -8,15 +8,15 @@
 //	Vulcan, New Glenn
 //
 //	Dry Mass: 2250 Kg
-//	Thrust (SL): 2400 kN
-//	Thrust (Vac): 2647.5 kN
-//	ISP: 305 SL / 341 Vac
+//	Thrust (SL): 2446.55 kN
+//	Thrust (Vac): 2714.15 kN
+//	ISP: 350 SL / 355 Vac
```
DRVeyl, line 13: I've seen this pairing of numbers questioned in a few places.  It is very different from the previous values.  I'd like some additional comments on it [from you or others].

GameData/RealismOverhaul/Engine_Configs/RL10_Config.cfg
```diff
@@ -7,7 +7,7 @@
 //	RL10A-1
 //	Atlas-Centaur A/B/C
 //
-//	Dry Mass: 145 Kg
+//	Dry Mass: 131 Kg
```
DRVeyl, line 10: Some of these mass changes are fairly significant (10%).  Are our previous sources wrong, or were the values just guess?  I'd prefer not to flip between different sources without something to declare why one is more reliable.

DRVeyl (commented):
Almost done.  One other point of concern is we're actually reaching a limit where the MM for handling TF/TL only ever expected 10 configs.  Above that, the TF patches don't get processed.  We can fix that elsewhere, but raising it as an issue. (See the extensive duplication of blocks in https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg because there isn't the equivalent of while; the v2 configs in progress also have this issue).

GameData/RealismOverhaul/Engine_Configs/RL10_Config.cfg
Bornholio: I rarely trust net sources myself, NTRS and DTIC.mil however should be very concise.
We have very accurate numbers for an RL-10A-3-3,  That is 289.79lbs with 3 sigma of 290lbs that is to say (99.73%) of all RL10A-3-3 engines will weight between 289.58 and 290.00 lbs (~131kg)
Instrumentation and hydraulic brackets add 10.17lbs (~4.6kg, these should be avionics weights)

https://ntrs.nasa.gov/api/citations/19670005471/downloads/19670005471.pdf

vader111:
@DRVeyl should be good for a review

DRVeyl (approved):

PR #2378, opened 2020-11-08
Fix nodes on restock'ed commDish

lpgagnon:


DRVeyl:
Can you make this into an issue or PR into <https://github.com/PorktoberRevolution/ReStocked>?  Would rather fix at the source, and the ReStock folks are pretty active.

lpgagnon:
As far as restock is concerned, "this is as intended". I half-expected as much. https://github.com/PorktoberRevolution/ReStocked/issues/919#issuecomment-723878684

PR #2379, opened 2020-11-10
Remove ignitions from early RD-211 configs

Capkirk123:
Remove ignitions from early RD-211 configs (later versions allegedly used Tonka250 ignitor fluid, and so are theoretically capable of airlighting)


DRVeyl:
Does this require `LiteralZeroIgnitions` to be set?

DRVeyl:
Wrong level, literalZeroIgnitions flag needs to be at MEC level, not per-config.

Capkirk123:
Fixed

PR #2382, opened 2020-11-22
Kerbalism-Engine: add hydrolox-for-fuelcells config

lpgagnon:


GameData/RealismOverhaul/RO_RecommendedMods/RO_Kerbalism.cfg
```diff
@@ -46,6 +39,7 @@
 		CONFIG
```
DRVeyl, line 39: If you want to get clever and set up for multiple different engine configs:
```
@CONFIG,*
{
  %minThrust = 0.01
  %maxThrust = 0.01
  %heatProduction = 1
  %massMult = 1.0
  %ullage = False
  %pressureFed = False
  %atmosphereCurve
  {
	key = 0 0.01
	key = 1 0.01
  }
}
```
Then you don't need to keep setting these parts of the settings.

DRVeyl (approved):

PR #2383, opened 2020-11-22
Make tanks able to hold FLOXes

vevladdd:


Capkirk123:
Could you add Xenon to HP tanks while you're at it?

Standecco:
I'd also add Krypton and Argon along with Xenon, now that we're at it.

vevladdd:
Ok, done

PR #2387, opened 2020-12-02
Fix Apollo CSM Chutes

Capkirk123:
Fix Apollo CSM chutes not working at all.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
@@ -81,8 +81,8 @@
 	@MODULE[ModuleDockingNode]
 	{
 		@nodeType = Apollo
-        %gendered = True
-        %genderFemale = False
+		%gendered = True
+		%genderFemale = False
 		%acquireForce = 0.5 // 2
 		%acquireMinFwdDot = 0.8 // 0.7
 		%acquireminRollDot = -3.40282347E+38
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_ApolloCSM.cfg
```diff
@@ -137,29 +137,48 @@
...
+		mustGoDown = True
+		spareChutes = 0
+		cutSpeed = 0.5
+
+		PARACHUTE
+		{
+			parachuteName = Canopy
+			capName = Cap
+			preDeploymentAnimation = Apollo_Chute_Semi
+			deploymentAnimation = Apollo_Chute_Full
```
DRVeyl, line 170: This sort of thing, with references to animations, really seems like it needs to be in RO_FASA instead of RO core.

Capkirk123: I couldn't get it to work any other way. Something is weird about the config in FASA I think.

DRVeyl: I was under an apparently-incorrect assumption that RealChute compatibility in general was done either by RC or by the original part mod.  Neither, at least in the case of FASA, and there's a few other full RC configs in RO core.

https://github.com/KSP-RO/FASA-RO/blob/master/GameData/FASA/Apollo/ApolloCSM/FASA_Apollo_CM_parachutes.cfg is the original config.  I don't see some of the fields you adjust above in ModuleParachute there, so @ would do nothing unless something -else- is poking them first.  I'm -really- not sure that "2" and "3" are reasonable values for [semi]deployed drag, and they conflict with what RO set before.  There's likely not a lot of testing in RO *without* RealChute.

mkosmo:
Just a casual bystander that was annoyed that the FASA Apollo parachutes didn't work in RO that led me here... but I can certainly say that this update does exactly what I'm expecting (at least that my Kerbals don't die anymore) and I'd love to see some kind of fixed merged.  What kind of testing would need to be done to get this added so I don't have to manually patch it going forward?

DRVeyl:
Mostly, addressing my comments above.  Some of these fixes should just go straight into RO-FASA, preferrably.  Failing that, some of the fields set here for non-RealChute need to be corrected before merging.

Capkirk123:
It looks like all FASA chutes are broken, in some way, the FASA Apollo chutes are just the most broken. I'll fix them in FASA so they at least work properly without realchute. Their RealChute integration can be fixed later

Capkirk123:
Right, I forgot what a mess that was. This patch makes the RealChute module work, and doesn't touch anything else.

PR #2395, opened 2020-12-16
RN KH9 adds

raidernick:
-First pass of kh9 configs


GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_USRockets.cfg
```diff
@@ -5909,6 +5991,101 @@
...
+	@MODEL
+	{
+		@scale = 1.0, 1.0, 1.0
+	}
+	
+	%scale = 1.0
+	@rescaleFactor = 1.0
+	
+	@node_stack_top = 0.0, 7.43952, 0.0, 0.0, 1.0, 0.0, 3
+	@node_stack_bottom = 0.0, -7.0764, 0.0, 0.0, -1.0, 0.0, 3
```
DRVeyl, line 6006: Just rescaleFactor rather than having to mess with the nodes

GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_USRockets.cfg
```diff
@@ -5909,6 +5991,101 @@
...
```
DRVeyl, line 6054: Just rescaleFactor rather than having to mess with the nodes

GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_USRockets.cfg
```diff
@@ -7962,6 +8139,24 @@
...
+{
+	%RSSROConfig = True
+	@MODEL
+	{
+		@scale = 1.0, 1.0, 1.0
+	}
+	%scale = 1.0
+	
+	@node_stack_bottom = 0.0, -0.670061, 0.0, 0.0, -1.0, 0.0, 3
+	@node_stack_top = 0.0, -1.130061, 0.0, 0.0, 1.0, 0.0, 3
```
DRVeyl, line 8152: rescaleFactor

GameData/RealismOverhaul/RealPlume_Configs/RaiderNick/rn_us_probes.cfg
```diff
@@ -321,3 +321,84 @@
         }
     }
 }
+
+@PART[rn_kh9_ds]:FOR[RealPlume]:NEEDS[SmokeScreen]
+{
```
DRVeyl, line 326: Do not use `:FOR[AnotherMod]`.
s/b `:BEFORE[RealPlume]`

GameData/RealismOverhaul/RealPlume_Configs/RaiderNick/rn_us_probes.cfg
```diff
@@ -321,3 +321,84 @@
...
+        speed = 1                       //Adjust speed to fit resize, 
+                                        //generally close to 1:1 with scale.
+    }
+    @MODULE[ModuleEngines*]
+    {
+        @name = ModuleEnginesRF
+    }
+    @MODULE[ModuleEngineConfigs]
+    {
+        %type = ModuleEnginesRF
```
DRVeyl, line 344: These don't really belong in Plume patching.  If they're necessary, put them in the main part.

GameData/RealismOverhaul/RealPlume_Configs/RaiderNick/rn_us_probes.cfg
```diff
@@ -321,3 +321,84 @@
...
+        @CONFIG,*   //Add the effect to every engine config
+        {
+            %powerEffectName = Hypergolic-OMS-Red
+        }
+    }
+}
+
+@PART[rn_kh9_campod_srb]:FOR[RealPlume]:NEEDS[SmokeScreen]
+{
```
DRVeyl, line 353: `:BEFORE[RealPlume]`

GameData/RealismOverhaul/RealPlume_Configs/RaiderNick/rn_us_probes.cfg
```diff
@@ -321,3 +321,84 @@
...
+        speed = 1                       //Adjust speed to fit resize, 
+                                        //generally close to 1:1 with scale.
+    }
+    @MODULE[ModuleEngines*]
+    {
+        @name = ModuleEnginesRF
+    }
+    @MODULE[ModuleEngineConfigs]
+    {
+        %type = ModuleEnginesRF
```
DRVeyl, line 371: Doesn't belong in plume patches

GameData/RealismOverhaul/RealPlume_Configs/RaiderNick/rn_us_probes.cfg
```diff
@@ -321,3 +321,84 @@
...
+        @CONFIG,*   //Add the effect to every engine config
+        {
+            %powerEffectName = Solid-Vacuum
+        }
+    }
+}
+
+@PART[rn_kh9_campod_s_srb]:FOR[RealPlume]:NEEDS[SmokeScreen]
+{
```
DRVeyl, line 380: `BEFORE[RealPlume]`

GameData/RealismOverhaul/RealPlume_Configs/RaiderNick/rn_us_probes.cfg
```diff
@@ -321,3 +321,84 @@
...
+        speed = 1                       //Adjust speed to fit resize, 
+                                        //generally close to 1:1 with scale.
+    }
+    @MODULE[ModuleEngines*]
+    {
+        @name = ModuleEnginesRF
+    }
+    @MODULE[ModuleEngineConfigs]
+    {
+        %type = ModuleEnginesRF
```
DRVeyl, line 398: Move to main part or delete.  (I think there is already a general patch that changed ModuleEngines to ModuleEnginesRF.  Just set the type once in the proper MEC patch, if you're writing your own.)

DRVeyl (changes requested):

raidernick:
@DRVeyl I am NOT moving the nodes and no I won't use rescalefactor. I am changing the node SIZES from 0 to 1 or 2 to 3. rescalefactor does not do this, it will rescale the POSITION yes, but so does the way I'm doing it and I'm not going to change that for no reason. As for using the global chute configs, I won't use those either as they define an amount of ablator which will affect the part mass. If I'm going to need to write a patch to override the global anyway, there is no point in using it.

As for using final, I tried using after zzrealismoverhual after realismoverhaul and LAST realismoverhaul. Something was still trying to overwrite it. So unless you can find what that is, I'm keeping final.

For the plume patching, who says they don't go there? Is there some rule I'm not aware of? There are hundreds of plume patches where that is in the plume patch.

The only legitimate thing I see is the question of FOR vs BEFORE for realplume. I've always seen FOR used as so I've used FOR and never had any issues. I don't see any advantage to changing everything to BEFORE if it won't actually do anything.

As for the rest of these problems, they aren't actually problems but formatting which I use to help me organize my thoughts while doing patches and I've always done it the same way and don't plan to change it.

DRVeyl:
Firstly, I appreciate your edit and the caveat.

Re: Changing just the node sizes.  Ok.  You can do that more clearly with @node[6] = <size> instead of replacing the entire node.  If you want to change part.scale and MODEL.scale, ok.  I have fixed a number of parts in RO that inaccurately tried to move attachment nodes in the patch (effectively breaking the part), to use rescaleFactor and sometimes the MODEL.scale, so IMO it is more resilient, but you have a reason not to.

Re: the global heatshield configs, I'd ask you to reconsider.  First, they have a method to opt out of automatically setting the ablator.  <https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_Heatshields.cfg#L73>, don't set resetHeatShieldAblator and just set the Ablator resource yourself.  They're written to be configurable to create the result you want.  Second, manually specifying the parameters of the shield in innumerable places makes it effectively impossible to change those parameters or make them consistent.  They are already very inconsistent through RO, and this has all downside.  Heating and heatshield behaviour is hard enough to understand and deal with in KSP without myriad different parts playing at tweaking settings.  Spent too many hours on this particular rabbit-hole, which is why I wrote the tool.  I think this side-steps your argument against generic patches, precisely because you are opting-in and setting fewer parameters for the tool to generate things from.

Re: FINAL.  Eh.  Search your MM log for the specific part, and at least that will give all the patches that touched it.  Have to manually search all of those to see what is actually offending.  Paste that here or message it to me and I'll help look?

For the plume patching, well, the who is me.  What does the plume care what the ModuleEngines or MEC type is?  Those affect things that definitely aren't the plume, and if you're asking "what the heck is patching my part weirdly?" [an entirely valid question!] ... your plume changing your engine type is definitely out of expected territory.  Yes, there are hundreds of plume configs that do this, mostly built by copying processes already done, whether they still make sense or not.  Or there were -- because I cleaned a lot of them up in RO's Plume patches: https://github.com/KSP-RO/RealismOverhaul/commit/31f817ed13f24b7c944235f97dc3472ccf262ab4

FOR[x] vs BEFORE[x]:  If you aren't the mod that owns the pass x, you should never declare it with a FOR pass.  That tells MM, and thus every other mod that wants to use NEEDS, that x exists.  And it may not.  So use BEFORE or AFTER.  (Here's a great example of the hundreds of copied patch files that are simply doing A Bad Thing).  We've gotten away with :FOR[RealPlume] because who plays RO without RealPlume?  Not a good reason to continue.  I already went through the effort of changing this in RO core, and also ROEngines.

I offered only one comment I consider related to style [multi-line node deletions].  I accept your reason for keeping the structure you had.

raidernick:
OK so i'll look at the heatshield thing more and I'll try to implement that, at this on this part. I'm not going to retroactively look through all my stuff right now and change everything.

I'll try and find what's modifying that config for the FINAL patches

For changing the removing definitions from the plume configs, I'll have to think about that as I don't feel comfortable just changing that and leaving everything else and don't know if I want to change everything I've ever done in my mod, FASA and KK at this time.

I can agree with replacing the FOR. What is least likely to break things, using BEFORE or AFTER? I want to do I find and replace on every single config in FASA my mod and KK which I all maintain here and don't want anything to secretly break because of this.

DRVeyl:
Thanks.  Agree, no need to update everything you have -- if you like the tool, just switch to it when you're updating a different part for another reason.

When I looked at it, BEFORE made more sense than AFTER for what we're doing.  So that's what we've been doing in RO core and ROEngines for a while and the plumes are still good.

raidernick:
EDIT: Here is the mm patch list for the 2 chute parts:
https://pastebin.com/T3eMjvKb

DRVeyl:
maxTemp and skinMaxTemp are being changed by <https://github.com/Starwaster/DeadlyReentry/blob/master/DeadlyReentry/DeadlyReentry-RealChutes.cfg#L1-L6>.  (FINAL, here?  WTF?)

Are we looking for something that sets the mass, too?

raidernick:
Yes It's setting the temps, which I don't want and it's also changing the masses, which I don't want. Also why IS there a FINAL there. That would explain why I had to use FINAL to get the damn patch to work...

PR #2397, opened 2020-12-20
Add SpaceX Launch Tower compatiability

kyleander:
Not sure if RescaleFactor is needed since parts are real scale, however, adding _%RSSROConfig = True_ was all that was needed to add Realism Overhaul compatiability.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
@@ -4,7 +4,7 @@
         RSSROConfig = True
 }
 
-@PART[AM_MLP]:FOR[xxxRP0]
+@PART[AM_MLP]:FOR[RP-0]
 {
     %TechRequired = start
     %cost = 5
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
@@ -1,13 +1,5 @@
-@PART[AM_MLP]
+@PART[AM_MLP*]:FOR[RealismOverhaul]
 {
-        %rescaleFactor = 1.61
-        RSSROConfig = True
-}
-
-@PART[AM_MLP]:FOR[RP-0]
-{
-    %TechRequired = start
-    %cost = 5
-    %entryCost = 5
-    RP0conf = true
+    %rescaleFactor = 1.61
+    %RSSROConfig = True
 }
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
@@ -1,10 +1,10 @@
-@PART[AM_MLP*]
+@PART[AM_MLP]
 {
         %rescaleFactor = 1.61
         RSSROConfig = True
 }
 
-@PART[AM_MLP_*]:FOR[xxxRP0]
+@PART[AM_MLP]:FOR[xxxRP0]
 {
     %TechRequired = start
     %cost = 5
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
@@ -1,5 +1,13 @@
-@PART[AM_MLP*]:FOR[RealismOverhaul]
+@PART[AM_MLP*]
 {
-    %rescaleFactor = 1.61
-    %RSSROConfig = True
+        %rescaleFactor = 1.61
+        RSSROConfig = True
+}
+
+@PART[AM_MLP_*]:FOR[xxxRP0]
+{
+    %TechRequired = start
+    %cost = 5
+    %entryCost = 5
+    RP0conf = true
 }
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Science818/Science818.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Science818/Science818.cfg
@@ -1,55 +1,55 @@
-@PART[39A_pad]:LAST
+@PART[39A_pad]:LAST[REALISMOVERHAUL]
 {
     @rescaleFactor = 1.0
     %RSSROConfig = True
 
 }
-@PART[39Base]:LAST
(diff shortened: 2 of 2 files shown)


kyleander:
Also added RP-1 compatibility for AlphaMensae's Modular Launch Pads by "merging" RP-1 compat code from the Extras folder included in the Modular Launch Pads release

GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
```diff
@@ -1,5 +1,13 @@
...
+        RSSROConfig = True
+}
+
+@PART[AM_MLP]:FOR[RP-0]
+{
+    %TechRequired = start
+    %cost = 5
+    %entryCost = 5
+    RP0conf = true
 }
```
DRVeyl, line 13: Please PR the RP-1 configuration portion into RP-1, not into RO core.  We'll merge that bit there.
Actually, please revert the changes to this file.  Removing the pass specifier isn't helpful, not sure what all parts are included in MLP that the * on the PART should also be removed.  (The current patch matches what I took from MLP's repo history.)

GameData/RealismOverhaul/RO_SuggestedMods/Science818/Science818.cfg
```diff
@@ -0,0 +1,56 @@
...
+@PART[Vandy_F9]:FOR[RealismOverhaul]:AFTER[Science818]
+{
+	@rescaleFactor = 1.0
+    %RSSROConfig = True
+}
+@PART[Vandy_Heavy]:FOR[RealismOverhaul]:AFTER[Science818]
+{
+	@rescaleFactor = 1.0
+    %RSSROConfig = True
+}
```
DRVeyl, line 56: It is not legal syntax to have 2 pass specifiers (FOR and AFTER).  Just pick one.  Most likely FOR[RealismOverhaul] is correct unless Science818 is doing something weird (like patching its own part's rescaleFactor in its own :FOR[Science818] pass....)

Also please fix the whitespace/indenting.

DRVeyl (changes requested):

GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
kyleander: No problem.

GameData/RealismOverhaul/RO_SuggestedMods/Science818/Science818.cfg
kyleander: I actually have to add AFTER because without it, the _RSSROConfig = True_ flag doesn't get applied.

GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
kyleander: I've reverted the config to be the original now.

GameData/RealismOverhaul/RO_SuggestedMods/Science818/Science818.cfg
kyleander: I've fixed the indexing and replaced the incorrect syntax with _:LAST_, since using _:FOR[REALISMOVERHAUL] won't work somehow.

DRVeyl: I think `LAST` requires a pass specifier.  `FIRST` doesn't, but `LAST` did.  `LAST[RealismOverhaul]`.
Can you clarify what you mean by "won't work?"  Did rescaleFactor not get set?  Does it have the wrong value?  Does `%rescaleFactor = 1.0` (`%` instead of `@`) work, ie if the field was not declared in the part?

DRVeyl: If you also had the ReScale patches from the Spacedock download, those edit rescaleFactor during the default pass.  This happens very early.  They would happen after your patches *if you didn't have any pass specifier at all* (and maybe if you have multiple/invalid specifiers) but a single specifier `FOR[RealismOverhaul]` should work just fine here.

kyleander: > I think `LAST` requires a pass specifier. `FIRST` doesn't, but `LAST` did. `LAST[RealismOverhaul]`.
> Can you clarify what you mean by "won't work?" Did rescaleFactor not get set? Does it have the wrong value? ~Does `%rescaleFactor = 1.0` (`%` instead of `@`) work, ie if the field was not declared in the part?~ (I went and looked at the original part config, this isn't it.)

the _%RSSROConfig = True_ part of the RO config did not get applied and so the parts appeared as "**Non-RO**"


> If you also had the ReScale patches from the Spacedock download, those edit rescaleFactor during the default pass. This happens very early. They would happen after your patches _if you didn't have any pass specifier at all_ (and maybe if you have multiple/invalid specifiers) but a single specifier `FOR[RealismOverhaul]` should work just fine here.

I did not have the SpaceDock rescale patches installed and for some weird reason _:FOR[REALISMOVERHAUL]_ didn't work in my testing, which is why I used _AFTER(Later corrected to LAST)_  however I will add REALISMOVERHAUL pass specifier after the LAST to it so it becomes :LAST[REALISMOVERHAUL]

DRVeyl: Please try it again with `:FOR[RealismOverhaul]` instead of LAST?  I cannot see why the FOR is incorrect, as that is literally when and how RO sets things everywhere.  There are a lot of things with exactly this structure.  <https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/SquadExpansion/RO_SquadExpansion_Structural.cfg> is nothing but this.

>the %RSSROConfig = True part of the RO config did not get applied and so the parts appeared as "Non-RO"

This suggests to me the entire patch simply did not run, period.  It doesn't appear that MM is case-sensitive, but maybe it is.  In which case, REALISMOVERHAUL in caps is never declared, and BEFORE/AFTER of a pass that is not declared will get removed.  (And you will see a note in the MM log about a pass being deleted because it did not meet its NEEDS)

Ex:
```
@PART[39A_pad]:FOR[RealismOverhaul]
{
    @rescaleFactor = 1.0
    %RSSROConfig = True
}
```

kyleander: No problem will correct to _FOR[RealismOverhaul]_  now.

PR #2398, opened 2020-12-23
Let SM tanks capture boiloff products

lpgagnon:
The old ServiceModule tank definition was set this way,
but it got forgotten from SM-I and friends

Only tested with LOX.

Also added Ammonia to HP tanks while I was at it; it would have
been silly to define a boiloff product that can't be stored anywhere.
(Kerbalism processes will also benefit)


lpgagnon:
might not want to merge this until RealFuels gets a release that includes the kerbalism boiloff support (https://github.com/NathanKell/ModularFuelSystem/pull/246), or there'll be a lot of annoyed kerbalism users

PR #2401, opened 2021-01-21
Adjust Derwent Performance

Capkirk123:
Adjust Derwent V performance based on historical documents

The Derwent will not overheat at full throttle when stationary, although it will overheat by the time you reach 100 m/s, so players will have about 30 seconds to decrease throttle, if they are paying attention (this margin may be much smaller in warmer areas, but will be much larger at higher altitudes).

Although this sadly does not prevent supersonic Derwent-powered aircraft, it makes supersonic flight at sea level nearly impossible for them, and reduces their max speed at altitude by 20-50 m/s.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SXT/Jets.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SXT/Jets.cfg
@@ -14,20 +14,20 @@
 	{
 		model = VenStockRevamp/Squad/Parts/Propulsion/EngineCore-Medium
 		position = 0.0, 0.0, 0.0
-		scale = 0.7,0.9,0.7
+		scale = 0.7,1.1,0.7
 	}
 	MODEL:NEEDS[VenStockRevamp/Squad]
 	{
 		model = VenStockRevamp/Squad/Parts/Propulsion/EngineCore-Small
-		position = 0.0, 0.7, 0.0
-		scale = 1.3,0.3,1.3
+		position = 0.0, 0.8, 0.0
+		scale = 1.4,0.4,1.4
 	}
 	MODEL:NEEDS[VenStockRevamp/Squad]
 	{
 		model = VenStockRevamp/Squad/Parts/Propulsion/EngineCore-Small
-		position = 0.0, 0.7, 0.0
+		position = 0.0, 0.8, 0.0
 		rotation = 180, 0, 0
-		scale = 1.3,0.03,1.3
+		scale = 1.4,0.03,1.4
 	}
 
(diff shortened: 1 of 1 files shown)


siimav:
There's 2 problems with this nerf. First off, the engine will overheat and explode within 3 second after it's turned on and run at full throttle. Secondly the Derwent no longer gets anywhere near it's rated thrust of 16kN.

NathanKell:
Can probably use Nene data since the V was a downscaled Nene?

Capkirk123:
I did use Nene data. Unfortunately, AJE seems to make some very optimistic assumptions about supersonic compressor efficiency and such, and so doesn't seem to properly represent the issues an early centrifugal compressor would probably have at supersonic speed. At the moment, there is simply no good way to prevent Derwents from going supersonic.

NathanKell:
Well, guess we can push some AJE changes then? Issue being it would melt?

Capkirk123:
I'm not sure anyone has actually made a centrifugal jet go supersonic, so it's hard to say what it'd do. If I had to guess, I'd say the very complex series of turns the air has to pass through to enter the jet would choke due to shock formation, you probably can't easily utilize ram air like a axial jet. If you wanna add a "centrifugal" tag to AJE, I can try and find some data on what a centrifugal jet would do at those speeds.

NathanKell:
Sure. I don't know enough to do it _right_, but I can for sure hit it with a hammer.

Capkirk123:
Ok, after a bit of research on the subject, I'd reckon the issue lies with AJE's calculation of recovery pressure and temperature. AJE seems to assume you are using a pretty well optimized intake duct at supersonic speeds, giving fairly generous recovery pressures and temperatures. This means the Derwent and other early jets aren't faced with the heating they probably should be. This is especially egregious on centrifugal jets, since the 90 degree bend (at a minimum) required to feed the compressor makes it impossible for them to fully utilize ram air and compression from a supersonic intake.

I would recommend adding a tag for centrifugal (and maybe all early) jets that slightly decreases their recovery pressure (maybe 5%) and increases their recovery temperature (maybe up to 50%) due to ram air at supersonic speeds.

NathanKell:
Worth trying https://github.com/KSP-RO/AJE/commit/d9aff69817a27c52bbd69a90ddf409ffa4774635 -- I added some code to try to deal with centrifugal flow compressors.

NathanKell:
The default curve I'm using is:
key = 0.85 1.0 0 0
key = 1.0 0.7 0 0
key = 1.3 0.85 0 0

Those are total guesses.

NathanKell:
There were problems. I believe I have fixed them. To be clear, master of RO and AJE are needed for things to work.

NathanKell:
This seems to be good now. Close?

PR #2403, opened 2021-01-26
Add early jets

Capkirk123:
Add the J47, J48, and VK-1.

### DIFF
--- a/GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
+++ b/GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
@@ -49,7 +49,6 @@
 
 	!MODEL,1 {}
 
-	// ven's 1.x used these paths...
 	MODEL:NEEDS[VenStockRevamp/Squad]
 	{
 		model = VenStockRevamp/Squad/Parts/Propulsion/EngineCore-Medium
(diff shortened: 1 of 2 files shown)


GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
```diff
@@ -37,6 +37,201 @@
...
+		position = 0.0, 0.45, 0.0
+		scale = 1.56,0.6,1.56
+	}
+	MODEL:NEEDS[VenStockRevamp]
+	{
+		model = VenStockRevamp/Assets/Engine/EngineCore-Small
+		position = 0.0, 0.45, 0.0
+		rotation = 180, 0, 0
+		scale = 1.56,0.03,1.56
+	}
```
DRVeyl, line 94: Oof.  Since RO will not release for anything less than KSP 1.8 and the current version of VSR... please check VSR's current master branch and only use the correct path from there.

Capkirk123: I just copied what already existed without looking too closely. I'll clean it up.

DRVeyl:
Style comment:  If the EFFECTS blocks are all the same [since you're cloning the same base part, even] would it make sense to have the patch only once and apply it to all 3 parts at the end, rather than copying into each part?

GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
```diff
@@ -37,6 +37,177 @@
 	%title = Avon RB-146 Mk.302 Turbojet
 }
 
+// Clone aje_j57 to be J48
++PART[aje_j75]:FOR[RealismOverhaul]
```
DRVeyl, line 41: Typo, looks like should be `aje_j57`

GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
```diff
@@ -37,6 +37,177 @@
...
+				pitch = 2.0
+				loop = false
+			}
+		}
+	}
+}
+
 // Set RSSROConfig, set AvGas
 @PART[*]:HAS[@MODULE[ModuleEnginesAJEPropeller]]:FOR[RealismOverhaul]
 {
```
DRVeyl, line 213: This patch seems like it wants to be `AFTER[RealismOverhaul]` since it wants to modify any part that configured an AJE propeller.  I guess this is currently fine if AJE is the only mod providing that prop info, since your SXT patches are AJEJet instead of prop.

GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
DRVeyl: Both aje_j57 and aje_j75 exist in AJE.  Which is intended?

Capkirk123: they're nearly identical. In this case, I meant to clone aje_j57, but apparently cloned aje_j75 instead, and configured the part accordingly, so it should stay as j75

GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
Capkirk123: It was like that when I found it

siimav:
Some problem as in #2401 - the engines do not get anywhere near their rated thrust.

Capkirk123:
I have reverted the engines to the original Derwent levels of heating, making them capable of much greater speeds.

siimav:
VK-1 and J47 are fine. J48 on the other hand is unusable without the full VSR. It's probably missing the model at index 1.

PR #2404, opened 2021-01-26
Add RealPlume to NTR

vevladdd:
This adds big and beautiful plumes, like the BNTR has.


DRVeyl:
Why did you change the plume position for nuclearEngine?
These parts are not from Squad.  They're from BornholioNuclear.  (Well, that's not a separate mod, that's a set of direct contributions to RO from Bornholio.)

vevladdd:
Thank you for your attention! This plume position really doesn't look good with NERVA and Phoebus 2A, unlike other engines mentioned in config.

PR #2405, opened 2021-01-28
Added selectable control points to command parts

RCrockford:
Useful for rovers and similar situations where you want to move forwards.

### DIFF
--- a/GameData/RealismOverhaul/RealismOverhaul_Global_Config.cfg
+++ b/GameData/RealismOverhaul/RealismOverhaul_Global_Config.cfg
@@ -221,6 +221,30 @@
 	}
 }
 
+//  ==========================================================
+// Add forward and reverse control points to all command parts
+//  ==========================================================
+
+@PART[*]:HAS[@MODULE[ModuleCommand]:HAS[~defaultControlPointDisplayName]]:FOR[zzzRealismOverhaul]
+{
+    @MODULE[ModuleCommand]
+    {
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RealismOverhaul_Global_Config.cfg
```diff
@@ -221,6 +221,30 @@
 	}
 }
 
+//  ==========================================================
+// Add forward and reverse control points to all command parts
+//  ==========================================================
+
+@PART[*]:HAS[@MODULE[ModuleCommand]:HAS[~defaultControlPointDisplayName]]:FOR[zzzRealismOverhaul]
+{
+    @MODULE[ModuleCommand]
```
DRVeyl, line 230: Should this be `@PART:HAS[@MODULE[ModuleCommand]:HAS[!CONTROLPOINT]]:FOR[zzzRealismOverhaul]` instead?  Is `defaultControlPointDisplayName` a required parameter for anything to spec control points?  Trying to cover the case that a part already has control points defined.  Also, why are "forward" and "reversed" not on the same axis?

RCrockford: `defaultControlPointDisplayName` does appear to be a required parameter.
I have verified that the cupola does no have its control points overwritten, I'm not aware of any other parts in the current configs which have them.

The axes names are copied from stock. "Forward" is for driving, "reversed" is the reverse of "default". We could use different names I suppose.

PR #2407, opened 2021-02-08
Improve tank names

Capkirk123:
Improve tank names and descriptions in order to discourage the use of SM tanks by new players.


GameData/RealismOverhaul/RO_RecommendedMods/Procedurals/RO_ProceduralParts.cfg
```diff
@@ -379,7 +379,7 @@
 	%RSSROConfig = True
 
 	@title = Procedural Tank (Service Module)
```
leudaimon, line 381: Maybe change the title to:
@title = Procedural Service Module
Then lower  down in the description it could work better with:
@description = <b><color=red>Not intended to hold fuel</color></b>, these specialized modules hold batteries, payloads and life support tanks in a controlled environment. Insulation and high pressure make service modules more expensive, heavier and with a much lower useful volume than conventional tanks.

PR #2415, opened 2021-03-15
Add depth mask to Restock RCS quads

al2me6:
Without this patch, the nozzles appear solid/filled in.

Before:

screenshot1

After:

screenshot2

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCS.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCS.cfg
@@ -1,4 +1,4 @@
-@PART[RCSBlock]:FOR[RealismOverhaul]
+@PART[RCSBlock]:NEEDS[ReStock]:FOR[RealismOverhaul]
 {
     MODULE
     {
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCS.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ReStock/RO_Restock_RCS.cfg
@@ -1,0 +1,8 @@
+@PART[RCSBlock]:FOR[RealismOverhaul]
+{
+    MODULE
+    {
+        name = ModuleRestockDepthMask
+        maskTransform = 4Mask
+    }
+}


DRVeyl:
This is a stock part, that ReStock operates on?  Why are we adding a Restock PartModule to such a part -- as in, why hasn't ReStock?

al2me6:
I'm not sure if my understanding of the situation is entirely correct, but it appears to me that the following happened:

1. In KSP 1.7, Squad revamped the RCS block (the new one is `RCSBlock_v2`). However, RO is still using the old part (`RCSBlock`). ReStock thus deprecated their patch for `RCSBlock` in [version 0.1.4](https://github.com/PorktoberRevolution/ReStocked/releases/tag/0.1.4).
2. ReStock added depth mask support in [version 1.0](https://github.com/PorktoberRevolution/ReStocked/releases/tag/1.0.0), which was released for KSP 1.8.
3. It appears that ReStock never bothered to add a depth mask to the deprecated part, which is, IMO, entirely reasonable. (Compare [patch for deprecated part](https://github.com/PorktoberRevolution/ReStocked/blob/master/Distribution/Restock/GameData/ReStock/PatchesLegacy/Control/restock-rcs.cfg#L7.) vs [new part](https://github.com/PorktoberRevolution/ReStocked/blob/master/Distribution/Restock/GameData/ReStock/Patches/Control/restock-rcs.cfg#L10).)
4. Thus, RO's RCS quads show up solid on the inside (my understanding is that the depth mask module does special handling to hide the specified mesh, which is otherwise visible).

Inspection of the deprecated model reveals that while it is not patched in by ReStock, the depth mask mesh itself does exist. Thus, I am adding it here.

I suppose that a more proper fix would be to base RO's RCS blocks on the new `RCSBlock_v2`, but that is a much larger change.

DRVeyl:
Ok. Since the release will target 1.8+, the latest Restock release supporting KSP 1.8 has the PartModule and Restock has not updated their deprecated part, I'll merge this.  Thankyou for the detailed notes.

Yeah, it will be better if we can start patching "new" Squad parts.  Apparently we are still versions behind on that front.  Not sure how all our different RCS config patching will play with Restock's single-part variants.

PR #2422, opened 2021-03-25
Remove additional comma from RealismOverhaul.netkan

DasSkelett:
## Problem
The CKAN download counter fails to process RealismOverhaul since a few days, caused by [this change](https://github.com/KSP-RO/RealismOverhaul/commit/753645e4accc8c0930eaf3718acebaf9daf50ed5#diff-9667970026e55def385bb9a8a16032db2b9c1905b324b8e1bd708303705396b6) from https://github.com/KSP-RO/RealismOverhaul/pull/2414.

It leaves behind a comma after the last element of an array, which is not valid JSON by strict rules.
This makes the download counter that uses Python's `json` module unable to load the netkan file and thus to fetch the download counts:
> Failed decoding count for RealismOverhaul: Expecting value: line 44 column 5 (char 1628)
image

Inflation however still works fine, since the JSON library we use in the NetKAN code is more permissive.

## Changes
Remove that comma, should make download counts for RO reappear in CKAN.

---

I want to point to https://github.com/KSP-CKAN/xKAN-meta_testing on this occasion, which is the validation tool the CKAN team itself uses in the main CKAN-meta and NetKAN repositories to check PRs before merging them.
We've recently redone it to a GitHub Action to allow it to be used by other repositories hosting metanetkans themselves, too.
It can catch errors like this during the pull requests or when commits are pushed:
* https://github.com/DasSkelett/RealismOverhaul/runs/2195480100
* https://github.com/DasSkelett/RealismOverhaul/pull/1/files#diff-9667970026e55def385bb9a8a16032db2b9c1905b324b8e1bd708303705396b6
* https://github.com/DasSkelett/RealismOverhaul/pull/1
image

Would you be interested in using this in your repository (and maybe the other repos as well)?
If that's the case, we would polish the workflow file and do a PR to add it to this repository.

### DIFF
--- a/RealismOverhaul.netkan
+++ b/RealismOverhaul.netkan
@@ -40,7 +40,7 @@
             { "name" : "Kerbalism-Config-RO" },
             { "name" : "TACLS" }
         ] },
-        { "name" : "TexturesUnlimited" },
+        { "name" : "TexturesUnlimited" }
     ],
     "suggests"   : [
         { "name" : "RealismOverhaulCraftFiles" },


HebaruSan:
@DasSkelett, might want to tag somebody, GitHub pull request notifications :rofl: .

DRVeyl:
Hey, guys, thanks for the fix!  Sorry, saw this earlier [there's a notification service between the GH and a Discord channel] but didn't have opportunity to respond.

Your team [particularly @HebaruSan] have been especially responsive, helpful, and proactive in identifying issues, thoroughly describing them, and getting to resolution.  We very much appreciate it!

I think I'd prefer to rely on the tooling on your end, and _generally_ stay out of hosting our own .netkan here versus in your repo and updating via PR.  You'll note how long we went between releases, and how even more rarely we have updated the .netkan.  Might be speaking for myself, though, so I'll se if anyone chimes in otherwise.

DasSkelett:
That was a relatively fast merge compared to our usual waiting times for missing or misplaced commas :laughing: 

From our side you are very welcome to move the netkans back to the KSP-CKAN/NetKAN repo. Feel free to open PRs doing so, or just nudge us when you came to a decision within the team.
And if you decide to keep the metanetkans, the offer to help adding the validation workflow to your repos still stands too, of course.

PR #2427, opened 2021-03-31
Make J48 restock-compatible

lpgagnon:
Main problem was wiping out all models, then using a NEEDS[VSR/Squad]
(which means the restock-incompatible bits) to install replacement
models, including one that lives in Ven's New Parts (the
restock-compatible bits). End result was a part with no MODELs at all
in a vnp+restock install.

### DIFF
--- a/GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
+++ b/GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
@@ -43,15 +43,23 @@
 	%RSSROConfig = True
 	@name = RO-J48
 
-	!MODEL,* {}
-
-	MODEL:NEEDS[VenStockRevamp/Squad]
+	// AJE replaces the original model IF VenStockRevamp is detected.
+	// We like the replacement model, but need to resize it.
+	// TODO: change the NEEDS to reference Ven's New Parts, when that becomes possible
+	// AND AJE is changed accordingly. Right now, if Ven's Core exists but not VNP, AJE
+	// replaces a presumably-fine model by a non-existent one)
+	@MODEL,0:NEEDS[VenStockRevamp]
 	{
-		model = VenStockRevamp/Part Bin/NewParts/JetEngines/HighBypassJet
-		position = 0, -1.383335, 0
-		scale = 0.6217, 0.6217, 0.6217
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
```diff
@@ -43,15 +43,23 @@
...
+		@scale = 0.6217, 0.6217, 0.6217
 	}
+	// TODO: find out how the model should be resized if AJE does NOT replace it
+
+	// Remove the ven-added, AJE-resized compressor model, if any
+	!MODEL,1 {}
 
+	// Add our own multi-model compressor, if Ven's Stock Revamp is there
+	// to provide the models.
+	// If it isn't, the engine will be just a nozzle, but it will work.
```
DRVeyl, line 62: Relying on MODEL order is risky at best.  Yes, I understand there are a number of examples that do it.
Can we think of an appropriate string within the MODEL node to use in a :HAS block instead?

Similar comment for `MODEL,1` but I think this is actually just the negation of the :HAS for the other patch.  Really, that's what `!MODEL,*{}` wanted to do.
```
@MODEL:HAS[#model[*HighBypassJet]]:NEEDS[VenStockRevamp/PartBin]
{
    @scale = 0.6217, 0.6217, 0.6217
}
!MODEL:HAS[~model[*HighBypassJet]]:NEEDS[VenStockRevamp] {}
```

I prefer to be in a slightly-broken state now with specifying the VNP model even though the path doesn't exist currently.  The required changes are merged into VSR/VNP master.  We are waiting only on a release, and I much prefer not to revisit this again after a release happens.

lpgagnon: A HAS works, yes; it's what I started out with. I don't see relying on a specific model name as _less_ risky than relying on order; there's no getting away from the fact that stock, ven's, AJE and RO are all tightly coupled here. The indexes seemed like a cleaner admission of that reality.

I don't see the sense of a "fix" using PartBin that will leave the part comically large for every install, however; that doesn't seem to be an improvement over the current "unusable with restock, correct without" state.

so... feel free to tweak this, but I'll be keeping my version

lpgagnon:
no longer needed now that RO uses its own copy of the compressor models

PR #2430, opened 2021-04-02
Add a Waterfall patching system

al2me6:
This PR is the first of a series of PRs to add optional Waterfall support to Realism Overhaul and related mods.

Specifically, this PR adds the `ROWaterfall` ModuleManager patchset, which is used to configure effects in a uniform way. It does not contain any configurations for parts.

These patches have been developed by @vevladdd and myself over the past month at https://github.com/vevladdd/ROE-Waterfall and have received some testing on the RO Discord. They should be reasonably robust.

We have designed these patches to be interoperable with RealPlume. They are set up such that if installed, Waterfall will take precedence over RealPlume. Engines without Waterfall configurations will continue to use RealPlume (or stock) effects, even if Waterfall is installed. No changes are made if Waterfall Core is not installed.

`ROWaterfall` requires the newest version of [Waterfall Core](https://github.com/post-kerbin-mining-corporation/Waterfall), which is version 0.6.3 at the time of writing. There are no other dependencies.

In the dropdown below is an introduction to using `ROWaterfall`, which should be added to a wiki page once this PR is merged.

<details>

## `ROWaterfall` Technical Documentation

`ROWaterfall` is a set of ModuleManager patches used to easily apply Waterfall and audio effects to engines.

### Basic Usage

The minimal configuration for adding a Waterfall plume and sound effect is as follows:

```text
@PART[PartName]:BEFORE[ROWaterfall]:NEEDS[Waterfall]
{
    ROWaterfall
    {
        template = waterfall-kerolox-lower-4
        audio = pump-fed-medium-1
    }
}
```

**Note that the patch should be applied in the `:BEFORE[ROWaterfall]` pass.**

`ROWaterfall` will produce a `Module[ModuleWaterfallFX]` node and an `EFFECTS` node from this patch, removing existing effects (stock, RealPlume) as necessary. However, it is still possible to use RealPlume (or stock) effects in conjunction with `ROWaterfall` – see below.

**Only one `ROWaterfall` node will be processed for a single part.** If more complex plumes are needed, extra `ModuleWaterfallFX`es can be added manually as usual; they can coexists with `ROWaterfall`.

### Configuration Options

An `ROWaterfall` node can contain the following keys and nodes:

| <div style='min-width: 9em;'>Key or node name</div> | Required | <div style='min-width: 6em;'>Default value</div> | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
| :-------------------------------------------------- | :------- | :----------------------------------------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `template`                                          | Yes      | N/A                                              | The desired Waterfall template name. Default Waterfall templates can be found in `Waterfall/Templates`, and RO provides supplemental templates under `RealismOverhaul/Waterfall_Configs/_Templates`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
| `audio`                                             | No       | None                                             | The audio template to use, a selection of which are available in `RealismOverhaul/Waterfall_Configs/_Audio`. <br/> *While not strictly required, omitting this key will result in a silent engine.*                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
| `moduleID`                                          | No       | Part's `name`                                    | Override the `moduleID` of the generated `ModuleWaterfallFX`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
| `engineID`                                          | No       | `basicEngine`                                    | Set the `engineID` used for the throttle controller, if one is present.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| `transform`                                         | No       | `thrustTransform`                                | Set the value of `overrideParentTransform` on the main template. If the template is an RCS template, then the `thrusterTransformName` key of the RCS controller will also be set to this value.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
| `position`                                          | No       | `0,0,0`                                          | Set the position of the main template.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
| `rotation`                                          | No       | `0,0,0`                                          | Set the rotation of the main template.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
| `scale`                                             | No       | `1,1,1`                                          | Set the scale of the main template.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
| `glow`                                              | No       | None                                             | Add a nozzle glow that appears at the same position and has the same size as the base of the main template. This can be used to cover up the prominent "holes" in the plumes of engines that lack emissive textures. <br/> The value can be of two forms: `_<color>`, referencing one of the `waterfall-nozzle-glow-<color>-1` templates; or `ro-<name>`, referencing one of the `rowaterfall-glow-<name>` templates. *Setting this key to any other value will result in undefined behavior*. <br/> By default, the length of the generated glow will be twice the average of the x and y `scale`s of the main template. <br/> Glows that require more complicated configuration (ex. custom positioning) should be added manually using an `ExtraTemplate` node. |
| `glowStretch`                                       | No       | `1`                                              | This key applies a multiplier to the length of the generated glow. It is useful for very short nozzles, where the default length might result in unsightly clipping.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
| `ExtraTemplate {}`                                  | No       | None                                             | This node can be used to add extra templates (ex. for verniers) to the generated `ModuleWaterfallFX`. <br/> The `template` key is required. The `transform`, `position`, `rotation`, and `scale` keys are optional and will inherit from the parent `ROWaterfall` node if not specified. <br/> Multiple `ExtraTemplate` nodes may be used in an `ROWaterfall` node.                                                                                                                                                                                                                                                                                                                                                                                                |

Here is an example using some of the optional configuration options:

```text
@PART[ROE-RD107]:BEFORE[ROWaterfall]:NEEDS[Waterfall]
{
    ROWaterfall
    {
        template = waterfall-kerolox-lower-1
        audio = pump-fed-heavy-1
        transform = RD-107A-MainFXTransform
        position = 0,0,0.48
        rotation = 0,0,0
        scale = 1.3,1.3,1.3
        glow = _yellow  // Expanding to `waterfall-nozzle-glow-yellow-1`.

        ExtraTemplate
        {
            template = waterfall-kerolox-vernier-2
            transform = RD-107A-VernierFXTransform
            position = 0,0,0.29
            rotation = 0,0,0
            scale = 2.5,2.5,2
        }
    }
}
```

### Plume Templates

In addition to the templates provided by Waterfall, `ROWaterfall` also ships a number of supplemental templates.

#### The `rowaterfall-glow-` family

These effects are recolors of the glows shipped with Waterfall.

Currently provided colors:

* `hydrolox-blue`: Less brightly cyan blue glow for upper-stage hydrolox engines.
* `hydrolox-red-blue`: Transitions from red at sea-level to blue in vacuum, for sea-level hydrolox plumes.
* `hypergolic-az50`: Yellow-orangish, for use specifically with aerozine50-NTO plumes.
* `hypergolic-white`: White, without a yellow tinge, for use with generic hypergolic/monopropellant plumes.
* `methalox-blue`
* `methalox-purple`
* `ntr`

![glow gallery](https://i.imgur.com/Qxu5qon.png)

From left to right: `hydrolox-blue`, `hydrolox-red-blue` at sea level, `hydrolox-red-blue` in vacuum, `hypergolic-az50`, `hypergolic-white`, `methalox-blue`, `methalox-purple`, `ntr`.

#### `rowaterfall-hypergolic-superdraco`

Orange hypergolic plume with prominent shock diamonds, inspired by SpaceX SuperDraco thrusters.

![SuperDraco plume](https://i.imgur.com/qoL9MvI.jpeg)

#### Templates from Bluedog Design Bureau

* `BDB_HTP_vernier`
* `BDB_HTP_vernierVac`
* `BDB_nuclear_PBR_vac`

### Audio Templates

`ROWaterfall` will generate audio effects using the default node names: `running`, `engage`, `disengage`, and `flameout`.

The following generic templates are available:

* `pressure-fed-1`
* `pump-fed-very-light-1`
* `pump-fed-light-1`
* `pump-fed-medium-1`
* `pump-fed-heavy-1`
* `pump-fed-very-heavy-1`
* `rcs-jet-1`

The series of `pump-fed` effects are increasingly crackly, based on sound effects from Rocket Sound Enhancement (shipped as part of Waterfall).

There are also the following specialized effects for use:

* `pump-fed-f1-1`, which is extra crackly and energetic.
* `pump-fed-gamma-1`
* `pump-fed-lr87-1`, with the real "whoop" startup sound.
* `pump-fed-raptor-1`, with real sound effects from SpaceX test footage.
* `pump-fed-stentor-1`

### Using RealPlume or stock effects in conjunction with Waterfall

RealPlume effects can be added in the `:AFTER[ROWaterfall]` pass if they need to coexist with `ROWaterfall`. If they are added any earlier, they will be deleted by `ROWaterfall` in `:FOR[ROWaterfall]`.

Stock (including audio) effects can be added in `:AFTER[ROWaterfall]` *as long as they have a different name than [the ones used by `ROWaterfall`](#audio-templates)*. Otherwise, see [modifying effects generated by `ROWaterfall`](#modifying-effects-generated-by-rowaterfall).

### Modifying effects generated by `ROWaterfall`

**Warning:** This is inherently fragile; do so at your own risk.

Modifications to effects (`ModuleWaterfallFX` or `EFFECTS`) generated by `ROWaterfall` can be done in `:AFTER[zROWaterfall_99_Finalize]` or a later pass. Patches *should not* be added in passes between `zROWaterfall_00` and `zROWaterfall_99`. These are used internally by `ROWaterfall` and their behavior could change at any time.

### Adding templates to `ROWaterfall`

By convention, templates provided by `ROWaterfall` should have names beginning with `rowaterfall`.

All nozzle glows should have the same default position and size (that is, the visual size and position when `position`, `rotation`, and `scale` are all their default values) as the glows shipped with Waterfall. This is required for glow rescaling to work.

</details>

## Known Issues/Limitations

* ~~Only one plume can be configured for each engine, meaning that some engine configurations will not have matching plumes. Plume switching support will likely require work from RealFuels. At the moment, affected engines use the most common plume (ex. an AZ50/NTO plume for the LR-87, even though it has kerolox configurations).~~ This is being addressed with https://github.com/NathanKell/ModularFuelSystem/pull/259.
* ~~In some specific scenarios, engines using the `pump-fed-light-1` and `pump-fed-medium-1` audio effects may be unbearably loud. We were forced to set extremely high volumes (\~140) for these two templates because they were unacceptably quiet on our machines using normal volumes. This appears to work, but we have received one report of extremely _loud_ audio, presumably due to the sounds not being broken on that machine. We do not know what caused the sounds to be quiet to begin with.~~ Engines using the `pump-fed-light-1` and `pump-fed-medium-1` audio effects may be very quiet. We do not know why these two specific effects are broken.

~~Further PRs to RO, ROEngines, and ROCapsules to add configurations for specific engines will be coming soon.~~ Update: Config PRs have been opened.


DRVeyl:
> In some specific scenarios, engines using the pump-fed-light-1 and pump-fed-medium-1 audio effects may be unbearably loud. We were forced to set extremely high volumes (~140) for these two templates because they were unacceptably quiet on our machines using normal volumes.

Please don't do this.

al2me6:
> Please don't do this.

Fixed.

pap1723:
@DRVeyl and/or @blowfishpro can you please review the MM-fu that @al2me6 implemented here? They are asking for someone to peek at it.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/_early-initialize.cfg
```diff
@@ -0,0 +1,43 @@
...
+    @EnableLights = false
+}
+
+// Remove RealPlume templates. If a RealPlume template is needed in addition to Waterfall, add it
+// in `:AFTER[ROWaterfall]`.
+@PART[*]:HAS[@ROWaterfall]:FOR[ROWaterfall]:NEEDS[Waterfall]
+{
+    !PLUME,* {}
+    !PLUME_TEMPLATE,* {}
+}
```
DRVeyl, line 13: This seems a little hostile, to force another system to run in a late pass to accomodate this system.  Not sure when the ROWaterfall node is expected to be added to a part, but suggest allowing a tag in there to indicate if RealPlume configs should not be removed.

DRVeyl (changes requested):
Some changes required, some are just suggestions.
Please inspect items where removing `,*` should still work.  Leaving those in extraneously tends to create confusion.
I did not inspect the templates.  I have no idea how to evaluate a 4000+ line template.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/18_late-initialize.cfg
```diff
@@ -0,0 +1,54 @@
...
+            &rotation = #$../rotation$
+            &scale = #$../scale$
+        }
+    }
+}
+
+@PART[*]:HAS[@ROWaterfall:HAS[#glow]]:FOR[zROWaterfall_18_LateInitialize]:NEEDS[Waterfall]
+{
+    @ROWaterfall
+    {
```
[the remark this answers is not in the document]
al2me6: Done.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/39_glow-resize.cfg
```diff
@@ -0,0 +1,162 @@
...
+@PART[*]:HAS[@ROWaterfall:HAS[#glow]]:BEFORE[zROWaterfall_39_GlowResize]:NEEDS[Waterfall]
+{
+    @ROWaterfall
+    {
+        __glowScale = #$scale$
+
+        // Stretch in z, while accounting for any z-stretching in the main plume itself.
+        // Twice the average scale of x and y by default.
+        @__glowScale[2] = #$scale[0]$
+        @__glowScale[2] += #$scale[1]$
```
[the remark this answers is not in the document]
al2me6: Sure, I'll reword this. I intentionally avoided using "sum" though, because the logic behind it is "first take the average nozzle diameter, then double it because that's what rocket nozzles are generally shaped like."

GameData/RealismOverhaul/Waterfall_Configs/_Processor/40_glow.cfg
```diff
@@ -0,0 +1,34 @@
...
+    {
+        +TEMPLATE:HAS[#__rowaterfall[mainPlume]]
+        {
+            @__rowaterfall = glow
+
+            @templateName = #$/ROWaterfall/glow$
+            // Accept either a name of the form "_blue", corresponding to "waterfall-nozzle-glow-blue-1"...
+            @templateName ^= :^_(.+)$:waterfall-nozzle-glow-$1-1:
+            // ...or a name of the form "ro-ntr", corresponding to "rowaterfall-glow-ntr".
+            @templateName ^= :^ro-:rowaterfall-glow-:
```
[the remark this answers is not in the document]
al2me6: Yes -- tested using a regex tester and also used for basically every engine that has a Waterfall patch.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/18_late-initialize.cfg
```diff
@@ -0,0 +1,54 @@
...
+            &transform = #$../transform$
+            &position = #$../position$
+            &rotation = #$../rotation$
+            &scale = #$../scale$
+        }
+
+        @MainPlumeVariant,*
+        {
+            // Required key: `name`.
+            // Required key: `template`.
```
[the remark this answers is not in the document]
al2me6: I'm not sure how well "obviously incorrect value" would work here for `template`, because specifying a template that doesn't exist causes Waterfall to emit NREs at runtime. IMHO that's even more puzzling behavior.

Using a `MainPlumeVariant` requires you to manually specify which RFMEC configs use which `name`s. Thus, I don't think it's likely that people will accidentally omit this key.

FWIW, there's documentation in the body of the PR that specifies what keys are required.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/90_controllers.cfg
```diff
@@ -0,0 +1,72 @@
+// Add controllers to the generated module. Place template-specific controllers first; modules will
+// otherwise receive a default set of controllers.
+
+@PART[*]:HAS[@ROWaterfall]:FOR[zROWaterfall_90_Controllers]:NEEDS[Waterfall]
+{
+    @MODULE[ModuleWaterfallFX]:HAS[@TEMPLATE:HAS[#__rowaterfall[mainPlume],#templateName[waterfall-ntr-lh2-1]]],*
```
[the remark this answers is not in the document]
al2me6: Sometimes I prefer adding an extraneous `,*` for the reader, because I don't think it's immediately obvious to someone not familiar with ModuleManager that `:HAS` implies "apply to all."

However, in this case, this is a vestige of an older attempt at plume variant switching, so I'll remove them.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/18_late-initialize.cfg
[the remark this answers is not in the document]
DRVeyl: Provide that as a bug report to Waterfall: they shouldn't emit NREs and should instead detect the invalid config and message an appropriate error in the log.  Which would be much more clear when you see something like `ERR: ModuleWaterfallFX tried to load a template with name "ROWaterfall-SomeoneForgotToAssignATemplateName" but couldn't find one!`

Documentation's great.  Now if only it prevented people from writing broken patches, even folks who know what they're doing.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/_early-initialize.cfg
al2me6: It was my intention to force clear existing RealPlumes so that any RealPlume that coexists with Waterfall needs to have been written with Waterfall in mind. That said, in some rare situations (ex. a part with SRM ullage motors and RCS) this results in needless copy-pasting, so I'll add the tag in.

I feel like `:AFTER[ROWaterfall]` is not a particularly late pass to be adding things -- this patch runs in `:FOR[ROWaterfall]` instead of a `zROWaterfall_xx` pass for that specific use in mind, after all.

The documentation states that `ROWaterfall` nodes are added in `:BEFORE[ROWaterfall]`.

FWIW, RealPlume doesn't have any kind of facility for keeping stock plumes (see https://github.com/KSP-RO/RealPlume/blob/501fb305f619f4f80bcee25efd5e1825f4815a61/GameData/RealPlume/000_Generic_Plumes/000_EFFECTS_Remover.cfg#L63-L66).

GameData/RealismOverhaul/Waterfall_Configs/_Processor/10_initialize.cfg
```diff
@@ -14,5 +14,16 @@
...
     }
+
+    // Prevent infinite loop if the part has multiple `ROWaterfall` nodes defined for some reason,
+    // and the subsequent nodes contain `ExtraTemplate`s or `MainPlumeVariant`s.
+    @ROWaterfall:HAS[~__is_first],*
+    {
+        |_ = ROWaterfall_CannotProcess
+    }
+
+    @ROWaterfall,0 { !__is_first = delete }  // Cleanup.
```
blowfishpro, line 27: At this point the `,0` is redundant right?

blowfishpro:
I don't really understand what all the waterfall values mean but the MM syntax looks generally okay

GameData/RealismOverhaul/RO_RCS_Fixes.cfg
```diff
@@ -40,3 +32,11 @@
...
 	}
 }
+
+@PART[*]:HAS[@MODULE[ModuleRCS]]:AFTER[zzzRealismOverhaul]
+{
+	@MODULE[ModuleRCS]
+	{
+		@name = ModuleRCSFX
+	}
+}
```
DRVeyl, line 42: Prefer the other way around, just editing the HAS block in your patch.  You also want to operate on ModuleRCSFX, or at least ModuleRCS* and not just ModuleRCS.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/10_initialize.cfg
al2me6: Yes -- would you think it's better removed? I just wanted to be explicit.

GameData/RealismOverhaul/RO_RCS_Fixes.cfg
al2me6: Hm, I'm pretty sure the current patch preserves the semantics of the original patch (not that I understand what it does and why it does it the way it does). 

The original applied only to parts with a `ModuleRCS`, renaming the module and rewriting the effects.

The new patch does the same, but only applies the effect rewriting if the part doesn't already have a `ModuleWaterfallFX`. If we renamed first, then we don't know which of the parts containing `ModuleRCSFX` originally had a `ModuleRCS` anymore.

DRVeyl: Ok.  I think I wasn't understanding the intent of the original patch [specifically: only rewriting the effects on ModuleRCS].  I'm good with this.

GameData/RealismOverhaul/Waterfall_Configs/_Processor/10_initialize.cfg
DRVeyl: Yeah, probably should remove.  `,0` suggests I should look for the others.
Also fix the :HAS + ,* in line 23 (remove the ,*)

al2me6: Already removed the `:HAS` ,* 😉 

But sure, I'll remove all of them -- don't think there are many, if at all.

DRVeyl (approved):

NathanKell:
@blowfishpro I note you mentioned the MM looked fine, should that be taken as an approve?

PR #2431, opened 2021-04-02
Add Waterfall configs for some ReStock parts and a few mods

al2me6:
This PR depends on https://github.com/KSP-RO/RealismOverhaul/pull/2430.

This PR adds Waterfall configs for the following parts:

* Realism Overhaul
  * 1.1/1.78 kN Thruster
  * 2.2/3.6 kN Thruster
* Squad (ReStock models only for now)
  * NSTAR ion engine
  * RCS quads
  * NERVA I
  * Bornholio Nuclear
    * KIWI A24
    * KIWI B48
    * NERVA NRX 50
    * NERVA XE 100
    * Pewee 100
    * Phoebus One 50
    * Phoebus Two 100
    * RD-0410 NTR
    * SNTP-PFE 100
* ATK Propulsion Pack
  * Star motors
* Mandatory RCS Parts Pack
* Internal RCS
* ReStock Plus
  * RCS blocks
* Shuttle Orbiter Construction Kit
  * OV RCS
* SXT
  * LEM Ascent Module RCS
* Ven's New Parts
  * RCS blocks
* Real Scale Sea Dragon
  * First and second stage engines
  * TVC engine

cc @vevladdd


GameData/RealismOverhaul/Waterfall_Configs/MandatoryRCSPartPack/mandatory-rcs.cfg
```diff
@@ -0,0 +1,75 @@
+// N.B.: These blocks are nasty because they manipulate the `MODEL`'s scale directly instead of
```
DRVeyl, line 1: Where is this done?  Don't see it in the patches, which only appear to manipulate the plume scale?

GameData/RealismOverhaul/Waterfall_Configs/ReStock/BornholioNuclear/restock_KIWI-A24.cfg
```diff
@@ -0,0 +1,11 @@
+@PART[KIWIA24]:NEEDS[Waterfall,ReStock]:BEFORE[ROWaterfall]
+{
```
DRVeyl, line 2: Why do all the Bornholio Nuclear patches require Restock?  (Is there no waterfall plume for the stock nuclear engine?)

GameData/RealismOverhaul/Waterfall_Configs/ReStock/restock_nerva.cfg
```diff
@@ -0,0 +1,20 @@
...
+{
+    ROWaterfall
+    {
+        template = BDB_nuclear_PBR_vac
+        audio = pump-fed-light-1
+        position = 0,0,0.107
+        rotation = 0, 0, 0
+        scale = 1.23, 1.23, 1.23
+    }
+}
```
DRVeyl, line 20: This part is the basis for all of the Bornholio Nuclear parts.  Whatever model is applied to this part becomes for all the others, with various rescaling.  Suggest for sanity that these plume patches compress into a single file, rather than separate files for each, and this will allow you to highlight the patching difference between a Restock-provided model and a stock model.  What about VSR?

One example, you might write the base patch, then write a patch with NEEDS[ReStock] to edit, then write a patch with :NEEDS[VenStockRevamp/Squad] and do an alternate replacement.

GameData/RealismOverhaul/Waterfall_Configs/MandatoryRCSPartPack/mandatory-rcs.cfg
al2me6: It's done here: https://github.com/KSP-RO/RealismOverhaul/tree/master/GameData/RealismOverhaul/RO_SuggestedMods/MandatoryRCSPartPack

GameData/RealismOverhaul/Waterfall_Configs/ReStock/BornholioNuclear/restock_KIWI-A24.cfg
al2me6: Each model (stock, VSR) requires separate configuration due to the nature of Waterfall plumes. I'll make plumes for the VSR models at some point, but I myself won't be making them for stock parts because the artwork quality is not there.

But yes, currently, stock and VSR will use the existing plume.

GameData/RealismOverhaul/Waterfall_Configs/ReStock/restock_nerva.cfg
al2me6: Sure, I can reorganize these patches into one file.

However, I still think that each model (Restock/VSR) should be patched in separate files, because from the view of Waterfall, they may as well be separate parts.

GameData/RealismOverhaul/Waterfall_Configs/MandatoryRCSPartPack/mandatory-rcs.cfg
DRVeyl: I see.  If we ever get around to touching those files, may fix.  (I tend to convert things to use rescaleFactor when I'm making other modifications).  Is that relevant to these patches?

GameData/RealismOverhaul/Waterfall_Configs/ReStock/restock_nerva.cfg
DRVeyl: Personally, I would prefer that they went into the same file.  Reason being all of the effort that went into deconflicting the separate stock, VSR, and Restock patches for most stock parts, highlighting a vast amount of patch duplication.  If something about the plume changes, I would much prefer to avoid, or at least make it less likely, that the change doesn't propagate to the other files.

GameData/RealismOverhaul/Waterfall_Configs/MandatoryRCSPartPack/mandatory-rcs.cfg
al2me6: No, not relevant. I complained here because I didn't expect these blocks to manipulate `MODEL` and figured that I should explain why this patch uses a different strategy than the other blocks.

GameData/RealismOverhaul/Waterfall_Configs/ReStock/restock_nerva.cfg
NathanKell: As someone who is very pro-VSR (but also now very pro-Waterfall), I would love to be sure that VSR models are handled appropriately. I realize that's annoying given the close links between Waterfall and ReStock, but still. :)

al2me6: VSR will 100% happen (it's on the [todo list](https://github.com/vevladdd/ROE-Waterfall/issues/5)), it's just a question of when. I'm not a VSR user myself, but it's used widely enough to warrant support.

If you made me guess, I'd say it'll happen in June.

NathanKell: Great! Think the only other Ven's stuff in heavy use is the various stuff derived from the LV-1 model?

PR #2432, opened 2021-04-03
SOCK shuttle orbiter configs

StonesmileGit:
Configs made by Aviation on the discord server, managed for this PR by me

Things that need work:
- ~Engines get convection heating behind the shuttle, causing them to overheat~
  - This will be addressed through FAR
- ~Investigate if the aerodynamics are correct with FAR~
  - It's good enough

Some parts from the mod are still unconfigured (spacelab + other crew tubes for the payload bay)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SOCK/RO_ShuttleOV.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SOCK/RO_ShuttleOV.cfg
@@ -103,7 +103,9 @@
 	%skinMaxTemp = 2400
 	%heatConductivity = 0.01
 
-	@MODULE[DRAG_CUBE]
+	%rescaleCube = 2
+
+	@DRAG_CUBE
 	{
 		%rescaleX = 1.81
 		%rescaleY = 1.81
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SOCK/RO_ShuttleOV.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SOCK/RO_ShuttleOV.cfg
@@ -1,0 +1,522 @@
+// Shuttle Orbiter Construction Kit
+
+// Orbiter configs
+@PART[benjee10_shuttle_forwardFuselage]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = True
+	@rescaleFactor = 1.81
+	@mass = 11.56
+	@maxTemp = 1400
+	%skinMaxTemp = 2700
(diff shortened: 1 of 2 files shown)


sidseth26:
Hi, 
i tried with RO configs for this mod as well. the main issue for contention was aerodynamics and lateral stability with FAR. Do you experience this as well?

StonesmileGit:
These configs are mostly ready to be merged, except for how FAR is handled. Help regarding FAR configs for especially the rear rudder is needed.

NathanKell:
What exactly is the issue with FAR?

StonesmileGit:
Currently the wings are not using FAR configs, due to them generating different amounts of drag, causing the orbiter to roll. The tail is also unconfigured, since FAR apparently does not like parts with multiple control surfaces, the airbrales are different surfaces on the same part

NathanKell:
I don't understand what you mean about the wings generating different amounts of drag (when FAR-configured).

As for the tail, yeah, you probably can't do the brake-rudder as one part unless dkavolis writes some new code.

StonesmileGit:
When the same config was applied to the left and right delta-wing, the orbiter started to roll, as if they experienced different drag. This was done by Aviation over on Discord, so might have been something in the configs that was wrong.

I guess the issue with the tail can be left as is, since the shuttle can be controlled decently without it, or by attaching stock AIRBRAKES to the aft.

NathanKell:
Are the wings attached by node, or surface attached? FAR doesn't support
wings that aren't surface-attached.

StonesmileGit:
The wings are indeed node-attached (everything in SOCK is node attached)

bedve:
I believe this is an issue due to the fact the wing part is the same part for both the right and the left side. If you take the Shuttle into a flight, and look at the f12 display, you will see that the lift/drag vectors are correct for the left wing, but not the right wing. My guess is that when b9 switches the orientation of the wing, this doesn't switch the FAR calculations as to where the Mid Chord is on the right wing. 

In the DECQ shuttle, which has separate parts for each wing, a similar issue happens. Despite having a separate part for the right wing, the lift/drag vector is shifted sideways (very similar to SOCK when you look at the f12) on the right wing. This was corrected by applying a "rootMidChordOffsetFromOrig" on the right wing part, which is what the current RO configs do - [link](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/SpaceShuttleSystem/WingSurfaces.cfg).

So I don't think your FAR configs that are commented out are really that FAR off (ha-ha).

PR #2437, opened 2021-04-05
Give Derwent a non-VSR compressor

lpgagnon:
Fairly half-assed; just something to show the player how big
and awkward the thing really is.

One possible argument _against_ merging this is the potential to confuse the player into thinking the weird look is a bug


lpgagnon:
I'm debating whether it's worth extending to more engines; there's a few more with the same "wide-diameter front part missing without VSR" problem.

pap1723:
This is no longer necessary with the compressors being added to RO now with https://github.com/KSP-RO/RealismOverhaul/commit/271db6e70746e9ff0b9ce8fcc4dd1df5079f9d27.

PR #2441, opened 2021-04-13
Update RO_HabTech2.cfg

si2504:
Adds all Habtech2 parts configured for RO, in order to build ISS to scale. Does require a few parts in original Habtech2 mod to have duplicate 'rescaleFactor' instructions deleted from their files configurations.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
@@ -1,157 +1,804 @@
-//	Leonardo Permanent Multipurpose Module (PMM)
-//	Diameter: 4.57m
-//	Length: 6.6m
-//	Mass: 4.437944128t
-//	Source: https://en.wikipedia.org/wiki/Leonardo_(ISS_module)
 
-@PART[ht2_MPLM]:FOR[RealismOverhaul]
+@PART[ht2_handrail]:FOR[RealismOverhaul]
 {
-	%RSSROConfig = True
-	
-	@MODEL
-	{
-		%scale = 1.828, 1.9, 1.828
-	}
-	@scale = 1
-	@rescaleFactor = 1
-	
-	// stack, srfAttach, allowStack, allowSrfAttach, allowCollision
-    @attachRules = 1,0,1,1,0
-	
-	@title = Leonardo Permanent Multipurpose Module (PMM)
-	@manufacturer = Italian Space Agency
-	@description = Leonardo is primarily used for storage of spares, supplies and waste on the ISS, which was until then stored in many different places within the space station.
-	
-	@mass = 4.4379
-	
-	// fixme
-	// @CrewCapacity = 0
-	
-	@category = Utility
-	
-	@maxTemp = 773.15
-	@skinMaxTemp = 773.15
-	@crashTolerance = 8
-	
-	@MODULE[ModuleB9PartSwitch]
-	{
-		@SUBTYPE[Bulkhead]
-		{
-			@NODE[nodeTop]
-			{
-				@position = 0.0, 3.2775, 0.0
-			}
-		}
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@title = ISS EVA/Service handrail
+	@manufacturer = NASA
+	@description = Handrails serve as mobility and stability aids for astronauts during extravehicular activity. Handrails are installed to define a safe path and provide sturdy anchors for the astronauts to use while they are working outside the spacecraft.
+	@mass = 0.001
+}
+
+@PART[ht2_CBM]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 0.05 
+	@title = ISS CBM
+	@manufacturer = Boeing
+	@description = The common berthing mechanism (CBM) is a berthing mechanism used to connect all non-Russian pressurized modules of the International Space Station. It was developed by Boeing at the Marshall Space Flight Center (MSFC) in Huntsville, Alabama while under contract to the National Aeronautics and Space Administration (NASA). 
+	@maxTemp = 1073.15
+	@MODULE[ModuleDockingNode]
+	{ 
+		@nodeType = ht2_CBM
+		%acquireForce = 0.5
+		%acquireMinFwdDot = 0.8
+		%acquireminRollDot = -3.40282347E+38
+		%acquireRange = 0.25
+		%acquireTorque = 0.5
+		%captureMaxRvel = 0.1
+		%captureMinFwdDot = 0.998
+		%captureMinRollDot = -3.40282347E+38
+		%captureRange = 0.05
+		%minDistanceToReEngage = 0.25
+		%undockEjectionForce = 0.1 
 	}
 }
 
-// 	TACLS Compatibility
-//	20,000 lbs of Storage = 9.07184t
-@PART[ht2_MPLM]:FOR[RealismOverhaulTACLS]:NEEDS[TacLifeSupport]
+@PART[ht2_solarArray_duo]:FOR[RealismOverhaul]
 {
-	@description ^= :$: Can store Life Support inventory.
-	
-	!MODULE[ModuleFuelTanks] {}
-	
-	MODULE
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 15.865
+	@title = ISS P4-P6/S4-S6
+	@manufacturer = Boeing 
+	@description = The International Space Station's main source of energy is from three of the four large U.S.-made photovoltaic arrays currently on the station, sometimes referred to as the Solar Array Wings (SAW). The first pair of arrays are attached to the P6 truss segment, which was launched and installed on top of Z1 in late 2000 during STS-97. The P6 segment was relocated to its final position, bolted to the P5 truss segment, in November 2007 during STS-120. The second pair of arrays was launched and installed in September 2006 during STS-115, but they didn't provide electricity until STS-116 in December 2006 when the station got an electrical rewiring. A third pair of arrays was installed during STS-117 in June 2007. A final pair arrived mid March 2009 on STS-119. More solar power was to have been available via the Russian-built Science Power Platform, but it was canceled. 
+	@maxTemp = 1073.15 
+	@MODULE[ModuleDeployableSolarPanel]
+	{	
+		@chargeRate = 70
+	}
+}
+
+@PART[ht2_truss_S0]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 10.785
+	@title = ISS S0 Truss Segment 
+	@manufacturer = Boeing 
+	@description = The S0 truss, (also called the Center Integrated Truss Assembly Starboard 0 Truss) forms the center backbone of the Space Station. It was attached on the top of the Destiny Laboratory Module during STS-110 in April 2002. S0 is used to route power to the pressurized station modules and conduct heat away from the modules to the S1 and P1 Trusses. 
+	@maxTemp = 1073.15 
+	@RESOURCE[ElectricCharge]
+	{
+		@name = ElectricCharge
+		%amount = 3000
+		%maxAmount = 3000
+	}
+}
+
+@PART[ht2_truss_S1]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 12.4
+	@title = ISS S1/P1 Truss Segment 
+	@manufacturer = Boeing 
+	@description = The P1 and S1 trusses (also called the Port and Starboard Side Thermal Radiator Trusses) are attached to the S0 truss, and contain carts to transport the Canadarm2 and astronauts to worksites along the space station. They each flow 290 kg (637 lb) of anhydrous ammonia through three heat rejection radiators. The S1 truss was launched on STS-112 in October 2002 and the P1 truss was launched on STS-113 in November 2002. 
+	@maxTemp = 1073.15 
+	@RESOURCE[ElectricCharge]
+	{
+		@name = ElectricCharge
+		%amount = 2500
+		%maxAmount = 2500
+	}
+}
+
+@PART[ht2_radiatorTriple]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 3.4
+	@title = Triple HRS radiators
+	@manufacturer = Lockheed Martin 
+	@description = The three HRS radiators on S1 weigh approximately 7,500 pounds, which is almost 30 percent of the payload on the Space Shuttle Atlantis.
+	%breakingForce = 250
+	%breakingTorque = 250	
+	%fuelCrossFeed = False
+	%radiatorHeadroom = 0.29 // 0.2702 sets the limit to 17C, or 290K
+	@maxTemp = 1073.15
+
+	@MODULE[ModuleActiveRadiator]
+	{
+        	@maxEnergyTransfer = 1750
+        	@overcoolFactor = 0.0186367
+
+        @RESOURCE[ElectricCharge]
+        {
+            @rate = 1.545
+        } 
+}
+
+@PART[ht2_PMA]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 1.25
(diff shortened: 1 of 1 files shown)


pap1723:
@si2504 can you please clarify your comment about the `rescaleFactor`? Is there something in addition that needs to happen after this file is implemented?

si2504:
> @si2504 can you please clarify your comment about the `rescaleFactor`? Is there something in addition that needs to happen after this file is implemented?

In the config files of 2/3 Habtech2 parts, they have 2 rescaleFactor instructions within the same config. Therefore, as my RO config only addresses 1 "@rescaleFactor = xxx" for each part, the parts that have duplicate rescaleFactor instructions don't get configured correctly as MM still reads the remaining "rescaleFactor" config line.

If you know how to write the config so that it will delete all "rescaleFactor" instructions, i can alter the configuration so that nothing else is required to be altered in the original Habtech folder.

pap1723:
Ohh, so the original config file is broken?

si2504:
> Ohh, so the original config file is broken?

only 3 of the parts. They aren't broken as such, they just require one of the rescaleFactor lines to be deleted from the configuration file. would the line "!rescaleFactor{}" added to the affected parts RO config be sufficient to get rid of the duplicate instruction do you think? The other recaleFactor instruction is beneath the part Model path. If so I can amend the config to take care of this.

si2504:
issue of duplicate rescaleFactor instructions has now been addressed.

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
-//	Leonardo Permanent Multipurpose Module (PMM)
-//	Diameter: 4.57m
-//	Length: 6.6m
-//	Mass: 4.437944128t
-//	Source: https://en.wikipedia.org/wiki/Leonardo_(ISS_module)
 
-@PART[ht2_MPLM]:FOR[RealismOverhaul]
```
DRVeyl, line 7: Does this part `ht2_MPLM` no longer exist?

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
...
+	@manufacturer = NASA
+	@description = Handrails serve as mobility and stability aids for astronauts during extravehicular activity. Handrails are installed to define a safe path and provide sturdy anchors for the astronauts to use while they are working outside the spacecraft.
+	@mass = 0.001
+}
+
+@PART[ht2_CBM]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	!rescaleFactor{}
+	%rescaleFactor = 1.5625
```
DRVeyl, line 16: syntax here is questionable, !node{} is syntax for deleting a single node, not a key/value.
What you want is `!rescaleFactor,* = DEL` to delete all previous ones (hack for the original part maker did it wrong)

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
...
+@PART[ht2_moduleJEMlogistics]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 8.336
+	@title = ISS JEM ELM-PS
+	@manufacturer = IHI AEROSPACE
+	@description = The Japanese Experiment Logistics Module, Pressurized Section (ELM-PS), also called the JLP, is a pressurized addition to the PM. The module is a storage facility that provides storage space for experiment payloads, samples and spare items.
+	@maxTemp = 1073.15 
+        MODULE
```
DRVeyl, line 480: Indenting

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
...
+	@maxTemp = 1073.15  
+	%MODULE[ModuleCommand]
+	{
+		%RESOURCE[ElectricCharge]
+		{
+			%rate = 10.2
+		}
+	}
+
+         	MODULE
```
DRVeyl, line 550: Indenting

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
...
+@PART[ht2_MPLM]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	@rescaleFactor = 1.5625
+	@mass = 4.082
+	@title = ISS Leonardo PMM 
+	@manufacturer = ASI 
+	@description = The Leonardo Permanent Multipurpose Module (PMM) is a module of the International Space Station. It was flown into space aboard the Space Shuttle on STS-133 on 24 February 2011 and installed on 1 March. Leonardo is primarily used for storage of spares, supplies and waste on the ISS, which is currently stored in many different places within the space station. The Leonardo PMM was a Multi-Purpose Logistics Module (MPLM) before 2011, but was modified into its current configuration. It was formerly one of three MPLM used for bringing cargo to and from the ISS with the Space Shuttle. The module was named for Italian polymath Leonardo da Vinci.
+	@maxTemp = 1073.15 
+         	MODULE
```
DRVeyl, line 601: Indenting

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
...
+			amount = 10000
+			maxAmount = 10000
+		}
+	}
+}
+
+@PART[ht2_radialTrussPort]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	!rescaleFactor{}
```
DRVeyl, line 723: Syntax: `!rescaleFactor,* = DEL`

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
```diff
@@ -1,157 +1,807 @@
...
+			maxAmount = 100
 		}
 	}
 }
+
+@PART[ht2_trussPort]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = False
+	!rescaleFactor{}
+	%rescaleFactor = 1.5625
```
DRVeyl, line 801: Syntax: !rescaleFactor,* = DEL

DRVeyl (changes requested):

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
si2504: yeah I wasn't sure how to write it as the original files had 2 rescaleFactor instructions. In order for the config to work correctly, I needed to delete one of the instructions.

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
si2504: yes it does, I think its just removed it from that line as in my configs its on a different line.

DRVeyl: Ok, I see it.  For my sanity in reviewing, can you move it back up to the top?

GameData/RealismOverhaul/RO_SuggestedMods/HabTech2/RO_HabTech2.cfg
si2504: > Syntax: `!rescaleFactor,* = DEL`

I'm in the process of making the required changes, could you please just confirm the issue with duplicate rescaleFactor instructions in the original file. will !rescaleFactor,* = DEL get rid of ALL rescaleFactor instructions within the configuration file or just the first one it comes across?

either way, in the configs for the affected parts I need to have a '@rescaleFactor = 1.5625 if it just deletes one, or %rescaleFactor = 1.5625 if it removes them both.

DRVeyl: `!rescaleFactor,* = DEL` will remove all instances of `rescaleFactor`.  Deleting all, and then creating a new one with `%rescaleFactor` will provide what we want, and will work even after the original mod fixes its error, if/when that happens.

Note that `%` is "create or edit."  If one doesn't exist, it'll create.  If one exists, it'll edit.  If multiple exists, it'll edit the first one.

si2504: cool, that's why I wrote it like I did in the first instance because I was trying to avoid deleting all rescaleFactor instructions as I still needed one. However I will alter it to how you have suggested.

Aviation365:
I recommend you change the rescalefactor on all the modules to 1.81, There was a minor conflict of interest where I also wrote HT2 configs, however I did not finish those. Current rescale is too small and 1.81 lines up with the current Shuttle Orbiter Construction Kit PR.

si2504:
> I recommend you change the rescalefactor on all the modules to 1.81, There was a minor conflict of interest where I also wrote HT2 configs, however I did not finish those. Current rescale is too small and 1.81 lines up with the current Shuttle Orbiter Construction Kit PR.

I wrote them based on the size and scale of Space Shuttle System which is already supported by RO.

StonesmileGit:
Did you just copy the rescale factor from SSS? Not sure but it might be that the stock scale is different for the different mods, and a scale of 1.5625 does not recreate the real scale of the ISS modules

si2504:
> Did you just copy the rescale factor from SSS? Not sure but it might be that the stock scale is different for the different mods, and a scale of 1.5625 does not recreate the real scale of the ISS modules

HT2 and SSS as standard share the same model sizes. Therefore, as SSS already had a rescaleFactor of that size for realismOverhaul, I kept HT2 in proportion to SSS as it was originally.

Aviation365:
The scales are still wrong when put into the SSS cargo bay, I already had figured out the scales and found rescale of 1.81 to be the best. 
Other than that there is the issue of the modules not lining up with the real scale at all, but that's an issue related to the mod design and can't really be fixed without breaking the docking ports

Aviation365:
If possible are you able to discuss on the discord? I would like to discuss with you about the conflicting configs.

si2504:
> If possible are you able to discuss on the discord? I would like to discuss with you about the conflicting configs.

Sure my Discord is si2504

si2504:
> The scales are still wrong when put into the SSS cargo bay, I already had figured out the scales and found rescale of 1.81 to be the best.
> Other than that there is the issue of the modules not lining up with the real scale at all, but that's an issue related to the mod design and can't really be fixed without breaking the docking ports

surely there has to be a balance though between mods and their RO config compatibility? how many other mods use the 1.81 rescaleFactor or one that equates to that? having a rescaleFactor of 1.81 would just destroy cross compatibility with the shuttle without it being re-written as the size would be un proportionate, not to mention the issue of not being able to dock a shuttle to it. The masses of parts and their onboard supplies/life systems are there as they are in life, but I just think you need to strike a balance. I'm all for scaling it up even more if needs be, but at the end of it all, the parts need to be useable with true to life missions.

si2504:
> > The scales are still wrong when put into the SSS cargo bay, I already had figured out the scales and found rescale of 1.81 to be the best.
> > Other than that there is the issue of the modules not lining up with the real scale at all, but that's an issue related to the mod design and can't really be fixed without breaking the docking ports
> 
> surely there has to be a balance though between mods and their RO config compatibility? how many other mods use the 1.81 rescaleFactor or one that equates to that? having a rescaleFactor of 1.81 would just destroy cross compatibility with the shuttle without it being re-written as the size would be un proportionate, not to mention the issue of not being able to dock a shuttle to it. The masses of parts and their onboard supplies/life systems are there as they are in life, but I just think you need to strike a balance. I'm all for scaling it up even more if needs be, but at the end of it all, the parts need to be useable with true to life missions.

also at 1.81, the S0, S1/P1 trusses don't even fit in when the shuttle ODS and the payload truss are in the payload bay.

StonesmileGit:
Here is a large document of compiled data on some space station mods (including HT2), some data there might be able to help
https://docs.google.com/spreadsheets/d/16hq_t1JknUVKbE1XGSAe9Z5FmUYQfW6zfW0VlH70QzM/edit#gid=733450764

si2504:
> Did you just copy the rescale factor from SSS? Not sure but it might be that the stock scale is different for the different mods, and a scale of 1.5625 does not recreate the real scale of the ISS modules

So it turns out that SSS and SOCK are equal in size to each other with your SOCK 1.81 rescale. SSS must have a slightly bigger model as standard hence why its only rescaled to 1.5625. I have however recompiled my Habtech configs and changed the rescale to 1.81. Since you have SOCK at 1.81, it might be worth configuring Benjee's APAS v2 to 1.81 if you haven't already. I've had a play with it and seems to be just right and fits the Habtech PMA perfectly.

Noiredd:
I've spotted a few problems with Connected Living Space compatibility - not sure if this is the fault of HT2 or this patch, nonetheless I thought I'd let you know my findings:
* CBM seems to be completely unpassable (connecting any two parts with a pair of those creates two separate spaces)
* PMA the same case
* Columbus has both passable nodes - shouldn't only the top node be passable?
* Kibo (JEM-PM) has all nodes passable, including the top node (where the EF attaches) - clearly a mistake
* JEM ELM-PS also passable on both ends - seems that only the bottom node should be passable
* Quest Joint Airlock as well (only the bottom node should be passable)
* Leonardo (both the full and hals versions) is always passable from both ends - this shouldn't be the case if the bottom node is in the "end cap" configuration (right?)
* Cupola - the top node definitely shouldn't be passable

si2504:
> I've spotted a few problems with Connected Living Space compatibility - not sure if this is the fault of HT2 or this patch, nonetheless I thought I'd let you know my findings:
> 
> * CBM seems to be completely unpassable (connecting any two parts with a pair of those creates two separate spaces)
> * PMA the same case
> * Columbus has both passable nodes - shouldn't only the top node be passable?
> * Kibo (JEM-PM) has all nodes passable, including the top node (where the EF attaches) - clearly a mistake
> * JEM ELM-PS also passable on both ends - seems that only the bottom node should be passable
> * Quest Joint Airlock as well (only the bottom node should be passable)
> * Leonardo (both the full and hals versions) is always passable from both ends - this shouldn't be the case if the bottom node is in the "end cap" configuration (right?)
> * Cupola - the top node definitely shouldn't be passable

HT2 nor these configs have any Connected living space configurations embedded into them. My concentration was to first configure the modules to RL sizes and provide life support systems for them. Once this has been reviewed and passed, CLS can be looked at and implemented.

realjuliuspgkc:
I've tested your code in 1.10.1 and it only renders a few parts. as RO Compatible.

Scorpuu:
> I've tested your code in 1.10.1 and it only renders a few parts. as RO Compatible.

I've been using this without any significant issues, all parts referenced in the config show up in game. There are only inconsistencies in some of the parts mass/volume/power values. Like:

- Solar Arrays (_ht2_solarArray_duo_) produce half the power they are supposed to produce. With four Solar Array sets, right now it results in 120 kW total, when it should be 240 kW (maximum, assuming 100% sun exposure: [wiki link](https://en.wikipedia.org/wiki/Electrical_system_of_the_International_Space_Station#:~:text=Altogether%2C%20the%20eight%20solar%20array%20wings%5B3%5D%20can%20generate%20about%20240%20kilowatts%20in%20direct%20sunlight%2C%20or%20about%2084%20to%20120%20kilowatts%20average%20power%20(cycling%20between%20sunlight%20and%20shade).%5B4%5D)). Also the panels aren't using RO's SolarConfig, so they don't wear out over time.

- All model variants of the P4-P6/S4-S6 Truss (_ht2_solarArray_duo_) have the same mass of 15.9 tons. I assume that's a limitation of B9PartSwitch and not much can be done about this.

- Habitable modules seem so have too large ModuleFuelTanks volume. As if their volume was estimated using procedural tanks. In reality, not all of that volume would be available to store supplies.

- Quest Nitrogen/Oxygen pod, aka High-Pressure Gas Tank _(ht2_questPod)_ is too heavy. It's dry mass is 462.6 kg (1020 lb): [Description of linked image](https://commons.wikimedia.org/wiki/File:STS-134_High-Pressure_Gas_Tank_(cover_removed).jpg).

PR #2452, opened 2021-05-11
Add pressurant to pressurefed engines

Capkirk123:
Make pressure fed engines consume pressurant, to simulate pressurant entering the fuel tanks and becoming unusable for other systems.

Currently, the pressurant consumption is determined by (fuel flow rate*chamber pressure(Bar)*1.5)/fuel flow rate, which provides relatively accurate pressurant mass for the AJ10 according to https://forum.nasaspaceflight.com/index.php?action=dlattach;topic=16412.0;attach=276500;sess=0


NathanKell:
Not sure if it should go in this PR or not, but it'd be great to also do the combining of Aniline and Furfuryl Alcohol you mentioned.

Capkirk123:
Yeah, I can look into that. We'd probably just need to create a new fuel in CRP. Also, when we do revamp tanks, requiring HP tanks to keep some gas around rather than just eating it all would be nice, it'd make stuff like Able that used tank pressurant to feed RCS  easier.

NathanKell:
Yup! That can also lead to lower dry mass for those tanks if we're not baking the pressurant in? Or were you already not doing that?

And we don't need it for CRP at least for now--NavSatPayload isn't in CRP to my knowledge, and yet it exists. I mean, it'd be good if it made it to CRP eventually, but let's not be beholden.

Capkirk123:
Typically dry mass does not include pressurant, so HP tanks are probably a bit too light at the moment.

NathanKell:
Figure it makes sense to ship this in the update. But looks like there's a bunch of engines left to do? Do you need help going through 'em all?

Capkirk123:
This was just a proof of concept to show I could get reasonable masses. If you want to ship it, I can update all other engines to use it too.

NathanKell:
If we're breaking early career with the science changes, might as well get
this in too!

NathanKell:
Uh, this has all the reliability changes too?

Capkirk123:
Oops, I merged it with the dev branch for testing.

NathanKell:
Ah! Gotcha. Yeah, I plan to nuke dev after merging every PR down, because it has a crapton of merge commits and (on RP-1) dll rebuilds. Thanks!

NathanKell:
Aerobee has IgnoreForISP in all three places, should be ignoreForIsp

PR #2454, opened 2021-05-14
Add small radial sep motor

jwvanderbeck:
This patch duplicates the existing Medium radial separation motor (sepMotor1) and makes a smaller version of it, in the same fashion that a larger one is made.  This expands the Medium & Large range to Small, Medium, and Large.

Also slightly renames the three to "Radial Separation Motor (Size)" to avoid confusion with the existing "Separation Motor (Small)" which is a different model entirely and most often (though not exclusively) mounted inline with a tank (on the ends) rather than on the sides.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -977,6 +977,11 @@
 
 	@mass = 0.044
 
+	@MODULE[ModuleEngines*]
+	{
+		%exhaustDamage = False
+	}
+
 	MODULE
 	{
 		name = ModuleFuelTanks
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1041,6 +1041,7 @@
 
 //  ==================================================
 //  Radial Separation motor (small).
+//  Rescale and Volume dont align because volume was set to match snubotron for balance reasons and size for visuals
 //  ==================================================
 
 +PART[sepMotor1]:AFTER[RealismOverhaul]
@@ -1066,7 +1067,8 @@
 	}
 }
 
-@PART[sepMotor1|ROSmallSpinMotor|SnubOtron|sepMotorLarge|sepMotorSmall]:FOR[RealismOverhaul]
+// sepMotorSmall and sepMotorLarge get this through the copy not directly
+@PART[sepMotor1|ROSmallSpinMotor|SnubOtron]:FOR[RealismOverhaul]
 {
 	@manufacturer = Generic
 	@crashTolerance = 10
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -966,13 +966,13 @@
 }
 
 //  ==================================================
-//  Separation motor (medium).
+//  Radial Separation motor (medium).
 //  ==================================================
 
 @PART[sepMotor1]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-	@title = Separation Motor (Medium)
+	@title = Radial Separation Motor (Medium)
 	@description = Small solid motor use to help separate one stage from another or perform ullage. Best used with others.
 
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
Draft for now.  Not yet tested in game.  Will also be doing an RP-0 PR as well for the tech tree side.

jwvanderbeck:
Tested and works in game.  There is possibly a weird visual glitch, but not sure if it is wrong or not: Plume only shows for a brief moment when first ignited and then again when it cuts out.  Don't know what about my configs could possibly cause that though so it might be a bug in some other mod.

lpgagnon:
at least one of the others (medium, large) has a plume that does something like what you describe already

jwvanderbeck:
> at least one of the others (medium, large) has a plume that does something like what you describe already

Ah that makes sense then since I just coped the existing one. I won't worry about it :)

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
```diff
@@ -1047,7 +1047,39 @@
...
+	@title = Radial Separation Motor (Small)
+	@description = Small solid motor use to help separate one stage from another or perform ullage. Best used with others.
+
+	%rescaleFactor = 0.5
+	@mass = 0.008
+
+	@MODULE[ModuleEngines*]
+	{
+		@maxThrust *= 0.18
+	}
```
DRVeyl, line 1066: I think this is unnecessary since they have an MEC.  (And could be removed from the SepMotorLarge copy, and also from the SepMotor1 original patch.)  The advantage of MEC is we don't need to manipulate the ME/MERF directly.

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
```diff
@@ -1047,7 +1047,39 @@
...
+	{
+		@CONFIG[SolidFuel]
+		{
+			@maxThrust *= 0.18
+		}
+	}
+}
+
+@PART[sepMotor1|ROSmallSpinMotor|SnubOtron|sepMotorLarge|sepMotorSmall]:FOR[RealismOverhaul]
 {
```
DRVeyl, line 1083: If you duplicated in :AFTER[RealismOverhaul] then it wasn't available to run against in FOR[RealismOverhaul].
Which isn't to say your change is wrong, the original patch was.  But remove sepMotorLarge and sepMotorSmall from here, because they don't exist yet.

jwvanderbeck: should they be added somewhere else?

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
jwvanderbeck: Done in https://github.com/KSP-RO/RealismOverhaul/pull/2454/commits/da7dfdb705a7bf91b7ee135d81c968ae24c176bf

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
```diff
@@ -966,25 +966,17 @@
...
 	@description = Small solid motor use to help separate one stage from another or perform ullage. Best used with others.
 
 	@mass = 0.044
 
-	@MODULE[ModuleEngines*]
-	{
-		@name = ModuleEnginesRF
-		@maxThrust = 98
-		@heatProduction = 17.5
-		%exhaustDamage = False
```
DRVeyl, line 985: This one line you might still need?  Not sure if that specifically gets set by the MEC.

jwvanderbeck: The `%exhaustDamage`?

jwvanderbeck: Resolved in https://github.com/KSP-RO/RealismOverhaul/pull/2454/commits/274bf8c18023363735347e4114908b902535abbb

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
jwvanderbeck: https://github.com/KSP-RO/RealismOverhaul/pull/2454/commits/a8ebe6e95f3177993a8c629348f33143f84ea0c5

PR #2457, opened 2021-05-19
Preliminary configs for new TF features

Capkirk123:
Seems to work correctly, needs fixed TF from Agathorn

TF prototype here https://github.com/KSP-RO/TestFlight/pull/218

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_137_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_137_Config.cfg
@@ -133,7 +133,7 @@
 	TESTFLIGHT
 	{
 		name = AJ10-137
-		ratedBurnTime = 750
+		ratedBurnTime = 750		//appears to be ablative, no extra burn time
 		ignitionReliabilityStart = 0.987342
 		ignitionReliabilityEnd = 0.997468
 		cycleReliabilityStart = 0.944444
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_190_Config.cfg
@@ -174,7 +174,15 @@
 	TESTFLIGHT
 	{
 		name = AJ10-190-OMS
-		ratedBurnTime = 3600
+		ratedContinuousBurnTime = 1250
+		ratedBurnTime = 54000
+		//restartWindowPenalty		//2 seconds for purge
+		//{
+		//	key = 0 1 0 0		//immediate restart before purge has small penalty
+		//	key = 0.3 1 0 0		//purge valves open after 0.36 seconds
+		//	key = 0.36 0.25 0 0		//restart attempt during purge has serious penalty
+		//	key = 2.36 1 0 0
+		//}
 		ignitionReliabilityStart = 0.998758
 		ignitionReliabilityEnd = 0.999752
 		cycleReliabilityStart = 0.996283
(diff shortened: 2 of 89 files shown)


jwvanderbeck:
[TestFlight.zip](https://github.com/KSP-RO/RealismOverhaul/files/6511161/TestFlight.zip)
Development testing version

jwvanderbeck:
This doesn't appear to include the changes to the Generic Configger?

Capkirk123:
testflight_generic_engines.cfg is the last file in the list, just a little buried

jwvanderbeck:
I see it now.  Guess I was blind this morning.

jwvanderbeck:
So this still needs some work on the Generic Configs to handle the restart windows, right?

NathanKell:
What is remaining on this? Just swapping the curves around again, so we don't have rated (total) burntime < rated (continuous) burn time?

Capkirk123:
> What is remaining on this? Just swapping the curves around again, so we don't have rated (total) burntime < rated (continuous) burn time?

I can rename the variable in the config to make things less confusing, while keeping the output the same.

jwvanderbeck:
I still don't understand what you are actually doing here, and don't see how changing a variable in a config is going to change anything since the player in game doesn't see that anyway?

NathanKell:
I believe the actual (gameplay) purpose here is to make overburning penalties less severe on some engines (in effect keeping the "rated burn time" as a descriptive feature, but not _actually_ penalizing longer burn times until they go past a second, longer, stored value?)

jwvanderbeck:
But the way it is applied right now is done by making the continuous rated time longer than the cumulative time which is just outright confusing. Is the continuous is 10 minutes and the cumulative is 1 minute, and I have 10 restarts what does that actually mean for operation?  If can burn ONCE for say 2 minutes and that's it? Any restarted I'd be past the cumulative.

If you want to make it safe to burn an engine longer than a historical rating then why not just put a low modifier and a flatish curve on the continuous rated time of the historical stage time and then set the cumulative time to the longer time desired?

Capkirk123:
> I still don't understand what you are actually doing here, and don't see how changing a variable in a config is going to change anything since the player in game doesn't see that anyway?

This was supposed to be a way to get avoid arbitrary burn limits. Most engines, especially more modern ones, can burn for several times their "rated" time, and were tested up to those limits. The rated burn time as described ingame is simply set to the longest burn time the engine was used for.

jwvanderbeck:
> > I still don't understand what you are actually doing here, and don't see how changing a variable in a config is going to change anything since the player in game doesn't see that anyway?
> 
> This was supposed to be a way to get avoid arbitrary burn limits. Most engines, especially more modern ones, can burn for several times their "rated" time, and were tested up to those limits. The rated burn time as described ingame is simply set to the longest burn time the engine was used for.

Yeah I get that but again why not: 

> If you want to make it safe to burn an engine longer than a historical rating then why not just put a low modifier and a flatish > curve on the continuous rated time of the historical stage time and then set the cumulative time to the longer time desired?

Capkirk123:
> > > I still don't understand what you are actually doing here, and don't see how changing a variable in a config is going to change anything since the player in game doesn't see that anyway?
> > 
> > 
> > This was supposed to be a way to get avoid arbitrary burn limits. Most engines, especially more modern ones, can burn for several times their "rated" time, and were tested up to those limits. The rated burn time as described ingame is simply set to the longest burn time the engine was used for.
> 
> Yeah I get that but again why not:
> 
> > If you want to make it safe to burn an engine longer than a historical rating then why not just put a low modifier and a flatish > curve on the continuous rated time of the historical stage time and then set the cumulative time to the longer time desired?
> 
> Unless it has only 1 ignition of course, but the majority of the ones you did this to do have restarts.
If I did that then it would artificially increase the reliability of a lot of engines. The cumulative burn time is used to calculate the MTBF/failure chance, but all of our data is based on the "rated" burn time.

Also oops, you're right. That does let engines with multiple ignitions burn as long as they want, doesn't it.

jwvanderbeck:
> If I did that then it would artificially increase the reliability of a lot of engines. The cumulative burn time is used to calculate the MTBF/failure chance, but all of our data is based on the "rated" burn time.

ehhh not quite.  I mean it's complicated.  Engine Cycle (IE the running of the engine) doesn't change the engine's MTBF.  The base reliability remains what it is regardless of the engine cycle.  The curve in the engine cycle however applied a multiplier to the MTBF based on time to generate a "momentary" failure rate for any given second of time.  Both the cumulative and continuous curves can do this.

No matter what you do the momentary failure rate is going to change.  That's how the system works.  That's how it essentially forces the engine to fail at some point by progressively increasing the multiplier until at a given moment the failure rate is so high it just can't keep going.

Capkirk123:
Yeah, I should probably take another look and see if there's a better way to do this. I was mostly just afraid of touching the MM-fu that makes the curves in the first place.

NathanKell:
Gonna merge this down to master unless there are objections, now that TF is ready to go.

PR #2477, opened 2021-06-28
Scale du by rated continuous time not cumulative

jwvanderbeck:
Not 100% sure this will work though so need an MM expert to weigh in.

### DIFF
--- a/GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg
+++ b/GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg
@@ -73,7 +73,7 @@
 	@TESTFLIGHT:HAS[~explicitDataRate[?rue]]
 	{
 		@reliabilityDataRateMultiplier *= 640 // normalized to rate=4 at 160s burntime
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg
```diff
@@ -73,7 +73,7 @@
 	@TESTFLIGHT:HAS[~explicitDataRate[?rue]]
 	{
 		@reliabilityDataRateMultiplier *= 640 // normalized to rate=4 at 160s burntime
-		@reliabilityDataRateMultiplier /= #$ratedBurnTime$
+		@reliabilityDataRateMultiplier /= #$ratedContinuousBurnTime$
```
al2me6, line 76: This is gonna explode because that value isn’t guaranteed to exist until line 303.

al2me6 (changes requested):

jwvanderbeck:
Yeah was afraid of something like that. 

I'm actually going to close this as was pointed out there needs to be more nuance into this anyway.

I will instead just open an issue for discussion.

PR #2480, opened 2021-06-29
TF patches fixes

NathanKell:
Redo the configurer to use ratedBurnTime, ratedContinuousBurnTime, and testedBurnTime. In normal use, leave ratedBurnTime as the stage burn time, and use testedBurnTime of the longer burn time for which the engine was tested. You must use safeOverburn=true in that case. Only for cases where an engine is used as a taxi engine with many relights during a single life cycle should the actual ratedBurnTime and ratedContinuousBurnTime keys in the TESTFLIGHT node be used.

Internally, what will be created are two curves, as before.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_Early_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_Early_Config.cfg
@@ -345,8 +345,8 @@
 	TESTFLIGHT
 	{
 			name = AJ10-37
-			ratedBurnTime = 150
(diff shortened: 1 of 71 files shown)


NathanKell:
Note @jwvanderbeck this requires a TestFlight change so that TF reports the 100% fail time accurately (currently it only reads the `cycle` curve, not the product of the two curves. This exposes the problem because for some engines, only a ratedBurnTime exists, so the dummy cycle curve is used. I could also fix this by renaming the continuousCycle node to cycle maybe?)

PR #2482, opened 2021-07-03
Rework TF cycle curve generation again

al2me6:
Now, there are three cases:

1. Only `ratedBurnTime` is specified. This is the pre-TF v2.1 behavior (only a `cycle` curve, 100 at 2.5x `ratedBurnTime`).
2. `ratedBurnTime` and `ratedContinuousBurnTime` are specified. This is for 'space taxi' engines whose full burn time is split across multiple small burns. In this case, create a `cycle` curve as with (a) but also a `continuousCycle` curve with the same rules (100 at 2.5x `ratedContinuousBurnTime`).
3. `ratedBurnTime` and `testedBurnTime` are specified. This is for engines that have been tested for a longer duration on the test stand than what they have been 'rated' for. In this case, Create only a `cycle` curve, but use more keys. After the (`ratedBurnTime` + 5, 1) point, the next point is (`testedBurnTime`, `overburnPenalty`). The in and out tangents (which are equal) on this point are set to provide a smooth transition in failure chance from `ratedBurnTime` to `testedBurnTime`, and then to complete failure at (2.5x `testedBurnTime`, 100).

Example of `cycle` curve generated for case (3) (this is the RL-10A-1):

image

### DIFF
--- a/GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg
+++ b/GameData/RealismOverhaul/TestFlight_Generic_Engines.cfg
@@ -233,7 +233,7 @@
 	//     engines whose full burn time is split across multiple small burns. In this case,
 	//     create a `cycle` curve as with (a) but also a `continuousCycle` curve with the
 	//     same rules (100 at 2.5x `ratedContinuousBurnTime`).
-	// (c) ``ratedBurnTime` and `testedBurnTime` are specified. This is for engines that have
+	// (c) `ratedBurnTime` and `testedBurnTime` are specified. This is for engines that have
 	//     been tested for a longer duration on the test stand than what they have been 'rated'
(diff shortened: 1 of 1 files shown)


NathanKell:
Had a quick read and looks neat! Will try this evening. There is one corner case that I'm not sure we can deal with in MM that I thought of: it is possible that the failure chance under case (a), at a given time, is lower than under case (b). I mention this because of the first few models of LR105, which has rated time of 330s and tested time of 350s. While I have not done the math, I'm pretty sure that a failure chance of 2 at 350s is worse than just using the original case-a curve would yield. I think for now we should just remove testedBurnTime from those two configs, but it is something to keep in mind in case there is a way to solve this with MM.

PR #2490, opened 2021-07-24
Adjust life support resources for Raidernick Soyuz 7K-OK for 20 day/2…

dgeastman:
… crew Soyuz 9 record, 10day/3 crew nominal mission (with 1 day margin)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Salyut_Soyuz.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_Salyut_Soyuz.cfg
@@ -1194,7 +1194,7 @@
 	MODULE
 	{
 		name = ModuleFuelTanks
-		volume = 1454.00001414
+		volume = 1250
 		type = ServiceModule
 		basemass = -1
 		TANK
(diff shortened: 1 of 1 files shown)


NathanKell:
I'm confused why the volume went down despite adding more resources

dgeastman:
The max values for food and water both went down. They were set closer to 14 days than the 11 I have here. The starting values went up however, as they were only like 3 days, not the 10 of a nominal mission.

raidernick:
@NathanKell do not merge this into a RO release without letting me know before you do it. This will require changes to all of the craft files that contain these parts.

@dgeastman Do not make changes to the craft files yourself and PR them, I will test this and do it if/when it needs to be done.

NathanKell:
Yup, that's why I've left it as is--as I said on discord when Squirrel asked, I know you tend to handle your own stuff and leave you to look over such PRs. :)

dgeastman:
To be clear, the patches here are just because as I was balancing out numbers and checking out real mission profiles, this is what I came up with for real mission duration and intent, and it's what I tested against. But if you want to stick with the original values here, I don't really have a problem with that. It's the PR in ROKerbalism to add nitrogen, PotassiumSuperoxide, and LithiumHydroxide that's important.

raidernick:
These fixes were incorporated into this PR: https://github.com/KSP-RO/RealismOverhaul/pull/2791

Thanks for the help here, this will now be closed.

Issue #2505, opened 2021-08-19
Reliability issues


NathanKell:
Think we probably need to police reliability data. Issues I'm skeptical of / worried about:
- [ ] Are we double-dipping on failure chance? 80% cycle reliability and 80% ignition chance is 64% _actual_ success chance, e.g. (It's actually slightly lower due to the 5s of increased failure chance)
- [ ] If possible we should tweak the configger to take the increased failure chance in the first 5s into account. Right now it's a uniform (but inversely proportional to burn time) increase in failure chance.
- [ ] Some configs just don't make sense. The Altair starts at like 99.8% reliability but the Antares starts very low, e.g.
- [ ] The LR79 has a serious problem where S-3D is cheaper but has better specs, *and* better reliability than the config that unlocks one node later (LR79-NA-9).
- [ ] In general, IMO we should not simply have exactly the same values for cycle reliability and ignition chance. Pressure-fed hypergolics, e.g., should have rather higher ignition chance and lower cycle reliability, and the reverse for pump-fed engines I think.
- [ ] Another issue with ignitions: subsequent relights should probably have a higher success chance. That might need a TF change however.
- [ ] We might want to include some level of tech transfer between branches. E.g. XLR41 data -> XLR43 -> LR79/89/105, but also LR87 tree should probably inherit from XLR43 since it does for the ECM. Antares should inherit a bit from Altair and GCRC (and probably Sergeant too). Not much, like 5% rather than 50%, but some. With TF's branch support, it will work fine, e.g. LR79-NA-13 gets 50% of previous LR79 data _or_ 10% of LR87 data, _or_ 20% of XLR43 data, etc.

Anything else?

leudaimon:
We could define a consistent way to deal with engines that have too few data for a good estimation of reliabilities, I have the impression this is all over the place

jwvanderbeck:
> Are we double-dipping on failure chance? 80% cycle reliability and 80% ignition chance is 64% actual success chance, e.g. (It's actually slightly lower due to the 5s of increased failure chance)

You have a 20% chance to fail one check and a 20% chance to fail the other.  They aren't cumulative because one isn't dependent on the other. (one doesn't influence the other)

StonesmileGit:
If the probabilities are completely independent, that means you can mutiply them, P(igniting) x P(burn for t seconds)

StonesmileGit:
Two more things:

[...]

PR #2513, opened 2021-09-05
Merge & Deprecate XLR43

Capkirk123:
Based on new information, deprecate the XLR43. The XLR43-NA-1 will be moved into the NAA-75-110 config, and the XLR43-NA-3 will be moved to the LR89 config.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/LR89_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/LR89_Config.cfg
@@ -4,6 +4,21 @@
 //	Manufacturer: Rocketdyne
 //
 //	=================================================================================
+//	XLR43-NA-3
+//	Navaho Phase IV
+//
+//	Dry Mass: 489 Kg	//Without gimbal. Use half of XLR71 mass (567 kg) to estimate gimbal mass
+//	Thrust (SL): ??? kN
+//	Thrust (Vac): 617.4 kN
+//	ISP: 230 SL / 265 Vac
+//	Burn Time: 65
+//	Chamber Pressure: ??? MPa
+//	Propellant: LOX / Ethanol90
+//	Prop Ratio: 1.375
+//	Throttle: N/A
+//	Nozzle Ratio: ???
+//	Ignitions: 1
+//	=================================================================================
 //	LR43-NA-3
 //	MA-1, or MA-0 without the sustainer
 //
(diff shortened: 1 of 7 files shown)


GameData/RealismOverhaul/Engine_Configs/LR89_Config.cfg
```diff
@@ -487,6 +539,25 @@
...
+		testedBurnTime = 200		//switched to spaghetti tube design, can probably safely make longer burns
+		ratedBurnTime = 65
+		safeOverburn = true
+		ignitionReliabilityStart = 0.666667		//Very troublesome to start, < 3 attempts required for most launches
+		ignitionReliabilityEnd = 0.933333
+		cycleReliabilityStart = 0.818182
+		cycleReliabilityEnd = 0.963636
+		techTransfer = XLR41-NA-1,XLR43-NA-1,A-6,A-7,A-6H:50
+		}
+	@MODULE[ModuleEngineConfigs] { @CONFIG[XLR43-NA-3] { %ratedBurnTime = #$/TESTFLIGHT[XLR43-NA-3]/ratedBurnTime$ } }
```
siimav, line 558: tabs vs spaces

GameData/RealismOverhaul/RO_SuggestedMods/ROMods/ROEngines.cfg
```diff
@@ -182,7 +182,7 @@
 @PART[ROE-XLR11]:FOR[RealismOverhaul] { %RSSROConfig = true }
 @PART[ROE-XLR25]:FOR[RealismOverhaul] { %RSSROConfig = true }
 @PART[ROE-XLR41]:FOR[RealismOverhaul] { %RSSROConfig = true }
-@PART[ROE-XLR43]:FOR[RealismOverhaul] { %RSSROConfig = true }
+//@PART[ROE-XLR43]:FOR[RealismOverhaul] { %RSSROConfig = true } // deprecated
```
siimav, line 185: Just remove the line

GameData/RealismOverhaul/Engine_Configs/RD108_RD118_Config.cfg
```diff
@@ -776,7 +775,7 @@
 		ignitionReliabilityEnd = 0.981818
 		cycleReliabilityStart = 0.909091
 		cycleReliabilityEnd = 0.981818
-		techTransfer = RD-200:25&RD-108-8D75:50
```
siimav, line 779: Should mention in the PR that it also fixes a bunch of bad tech transfer configs on other engines. Ideally these would be a separate PR though.

siimav (changes requested):

GameData/RealismOverhaul/Engine_Configs/NAA75_110_Config.cfg
```diff
@@ -270,7 +350,7 @@
 		ignitionReliabilityEnd = 0.97
 		cycleReliabilityStart = 0.77 // a bit higher than XLR43-NA-1
 		cycleReliabilityEnd = 0.94
-		techTransfer = XLR41-NA-1:30&XLR43-NA-1:50			// A-4/V-2 derivative. Modified XLR43-NA-1
+		techTransfer = XLR41-NA-1,XLR43-NA-1:50			// A-4/V-2 derivative. Modified XLR43-NA-1
```
NathanKell, line 353: Why increase the transfer from the XLR41 here? (That opens up a larger question I guess, which I've just added to #2505)

GameData/RealismOverhaul/Engine_Configs/RD107_RD117_Config.cfg
```diff
@@ -744,7 +744,7 @@
 		ignitionReliabilityEnd = 0.983077
 		cycleReliabilityStart = 0.915385
 		cycleReliabilityEnd = 0.983077
-		techTransfer = RD-200:25
+		techTransfer = RD-200:50
```
NathanKell, line 747: Same question: Why give a full 50% transfer between the RD200 series and the RD107? They're not the same engine at all...

NathanKell:
In the main this looks great, although I'm not sure about why you would remove the use of TF tech transfer branches and give a full 50% transfer even between separate engines. Can you go into the reasoning there?

Capkirk123:
Do multiple tech transfers work properly now? I might have to fix it in that case.

siimav:
Finally loaded up the RO and RP-1 PRs ingame.
* NAA-75 starts up with A-6 config applied
image
* XLR43-NA-1 is shown twice in TF configs section
* For the old XLR43 part should probably explicitly add the word Deprecated in the title and possibly also mention in the description that configs were moved to other parts.

NathanKell:
> Do multiple tech transfers work properly now? I might have to fix it in that case.

oh ugh thought I responded. Yes, they do. At first use of an engine, TF selects the branch with the highest transferred du (so they don't stack) but IMO that's correct. Format is (comma-separated list of configs):transferamount&next list:amount&....

Capkirk123:
I will fix this as soon as I get the time, and move the TF changes into a different PR when I redo reliability with a better grasp of statistics.

Capkirk123:
Ok, changed tech transfer to use multiple tech transfer sources correctly.
-Direct Navaho derivatives (LR79, LR89, LR105) get 20% from the XLR43-NA-1/NAA-75 line, and 50% from the XLR43-NA-3
-NAA-75 gets 25% from the XLR41
-LR87 now gets 25% from the XLR43-NA-3 (Aerojet wasn't involved in Navaho, but it still had significant influence)

NathanKell (approved):

PR #2515, opened 2021-09-06
Add RO configs for New Tantares Mir station parts revamp.

D0m1nu2:
Theese RO configs are for Mir Tantares Parts that are currantly being updated you cna get them there (https://github.com/Tantares/Tantares).
Theese parts are completed and will be officially release soon after the rest of revamp is done.
After revamp is done I'm most likely going to configure the rest of station parts for Salyut and Almaz stations but since a lot of people asked for this made at least theese.


D0m1nu2:
So, revamp of Mir parts has been officially released! You can get it here: https://github.com/Tantares/Tantares/releases/tag/v25.0
I tested it and everything should work fine, if you find some glitch of part config missing (Only Mir parts from revamp aka Aquilla parts) tell me here on on RO Discord, my name is D0m1nu2#7206.

al2me6:
Is there a reason why all the parts are duplicated instead of patched directly?

StonesmileGit:
AIUI the same stock part is used for multiple RO parts

al2me6:
Huh, I didn't notice any but I might not have looked close enough.

D0m1nu2:
> Huh, I didn't notice any but I might not have looked close enough.

Some of the parts like base of TKS based station is similar but I decided to duplicate it so it's more simple to build stations.

StonesmileGit:
Would it be possible to reuse the original part the first time it's used instead of leaving the original part unused? The current setup where the original part isn't used at all isn't common in RO configs.

If not possible, then at least a patch to remove the original parts would be good

PR #2522, opened 2021-09-26
Updated Space Shuttle System Configs

giuliodondi:
-Adjusted max temperatures of various parts and OMS engines.
-Added helium tanks since new OMS configs use it
-Added Helium consumption to RCS
-Increased SRB decoupler force
-Adjusted cargo bay CG offset to bring CG where it used to be after the FAR wing mass overrides were removed
-Fixed right wing FAR offset (resulted in roll moment)
-Fixed MAC of FAR wing configs (was too short)
- Adjusted value of cargo bay lifting coefficient (part of an attempt to increase L/D and crossrange)

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/SpaceShuttleSystem/WingSurfaces.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/SpaceShuttleSystem/WingSurfaces.cfg
@@ -4,7 +4,7 @@
 	
 	@mass = 5.3
 	
-	@maxTemp = 1600
+	@maxTemp = 2200
 	%skinMaxTemp = 2700
 	
 	%rescaleFactor = 1.5625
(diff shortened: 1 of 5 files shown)


StonesmileGit:
Would some of these changes (mainly aero) be useable for SOCK too?

giuliodondi:
The MAC change probably yes, that value is taken straight from the Orbiter aero data (I left the link to the technical paper in a comment in the configs). It should increase L/D slightly as the wing area is larger now.

The right wing FAR offset most likely no. This change is meant to roll back the last change done when the SOCK configs were also updated (I believe when you figured out how to disable FAR wing mass override). I presume things are fine with SOCK, for whatever reason the DECQ shuttle became unstable.

The cargo bay CG most likely needs adjusting for SOCK too but I wouldn't take this value straight. SOCK has a different number of parts and they probably have their origin in different locations, so the CG is probably somewhere else to begin with.

GameData/RealismOverhaul/RO_SuggestedMods/SpaceShuttleSystem/WingSurfaces.cfg
```diff
@@ -19,21 +19,22 @@
...
+		%rootMidChordOffsetFromOrig = 0, 0.5, 0
 		%massOverride = 0
 	}
 }
 @PART[WR_SHUTTLE|ShuttleWingR]:FOR[RealismOverhaul]
 {
 	@MODULE[FARWingAerodynamicModel]
 	{
-		%rootMidChordOffsetFromOrig[0] = 6.91
+		%rootMidChordOffsetFromOrig = 6.91, 0.5, 0
```
StonesmileGit, line 37: The previous line here changed the first value only. Is there a need for setting the whole vector when only the first value changes?

StonesmileGit: Huh, let's leave it as it is now then!

PR #2523, opened 2021-10-03
Bump up RD-215-8D513 reliability

lpgagnon:
Because giving a 10% failure chance to an engine with a perfect record
is excessive, even if the sample is small (28/28); especially considering
how reliable the successor configs proved. The RD-217 fared worse, but
also seems to have been "more different" than RD-215->RD-215M.

Picked 95% arbitrarily; if we suppose the next launch had failed, we'd
get 28/29=96.5%

This works out to a 10% failure rate for a dual-rd215 launch, so still
on the high side. (and worse if we count ignition failures)


lpgagnon:
superseded by #2539 (which is even more generous)

PR #2527, opened 2021-10-10
Support for new ROEngines parts

al2me6:
Support for https://github.com/KSP-RO/ROEngines/pull/146, also adding the RL-10A-4N and RL-10A-4-1N extendable nozzle engine configs.


Capkirk123 (approved):
Everything looks good from here!

PR #2539, opened 2021-10-31
Revise reliability data for engines

Capkirk123:
Revise reliability data for engines. Changes include

- Add 1 success and 0.75 failures to all data in order to imitate a binomial confidence interval.
- Distinguish between ignition and cycle failures for all air-lit engines, to avoid charging the same reliability roll twice (ground lit engines use the same data for both, since ignition failures on the ground are typically not recorded, and they only force a rollback to repair, rather than mission failure)


GameData/RealismOverhaul/Engine_Configs/AJ10_137_Config.cfg
```diff
@@ -134,18 +134,20 @@
...
 		name = AJ10-137
 		ratedBurnTime = 750		//appears to be ablative, no extra burn time - [A] describes it as 12.5min "max"
-		ignitionReliabilityStart = 0.987342
-		ignitionReliabilityEnd = 0.997468
-		cycleReliabilityStart = 0.944444
-		cycleReliabilityEnd = 0.998889
+		ignitionReliabilityStart = 0.990506
+		ignitionReliabilityEnd = 0.998101
+		cycleReliabilityStart = 0.958333
+		cycleReliabilityEnd = 0.991667
```
NathanKell, line 150: A minimum failure chance of 1% per run seems very, very high for SPS, given the amount of work that went into making it as reliable as possible. EDIT: in particular, it has a lower ceiling than the AJ10-138 on Transtage, which *actually suffered failures*. I would be inclined to give it at least four nines for reliability end in both ignition and cycle, though start values being low-ish seem fine.

GameData/RealismOverhaul/Engine_Configs/AJ10_Adv_Config.cfg
```diff
@@ -301,40 +301,51 @@
...
 		name = AJ10-118F
 		ratedBurnTime = 450
-		ignitionReliabilityStart = 0.923077
-		ignitionReliabilityEnd = 0.984615
-		cycleReliabilityStart = 0.875000	// 1 failure out of 8 flights
-		cycleReliabilityEnd = 0.975000
+		ignitionReliabilityStart = 0.953125
+		ignitionReliabilityEnd = 0.990625
+		cycleReliabilityStart = 0.906250
+		cycleReliabilityEnd = 0.981250
```
NathanKell, line 348: I am very doubtful that the ending reliability for 118F should be *lower* than for the -138. I really think we have to use judgment here, not just run the numbers, because otherwise we unfairly penalize rarely-flown variants.

NathanKell:
Also this is somewhat off-topic, but is on-topic to the extent it deals with reliability: what the heck do we do with the S-3D vs the LR79-NA-9? The latter is more expensive, but also both a worse performer *and* has lower reliability.

GameData/RealismOverhaul/Engine_Configs/AJ10_Adv_Config.cfg
Capkirk123: Oops, I tried to use other data for engines with lower sample sizes. I must have forgotten this one

NathanKell: :+1:

Capkirk123:
The S-3/S-3D was developed for Jupiter, and only got used for Thor later. The LR79-NA-9 was MB-I, for production Thors. Conceivably, some compromises were made to reduce cost for mass production of Thors, that also led to reduced reliability. I could do the same thing I did with the later RD-107/108 variants having worse performance than the start, and reduce their cost to represent mass production models sacrificing some reliability for easier production.

GameData/RealismOverhaul/Engine_Configs/AJ10_137_Config.cfg
Capkirk123: Same here, forgot to give the SPS special treatment

GameData/RealismOverhaul/Engine_Configs/Aestus_Config.cfg
```diff
@@ -239,17 +239,21 @@
 //	TestFlight compatibility.
 //	==================================================
 
-// 31 Flights, 1 failure.
+// 31 Flights, 1 failure (performance loss).
+// Mostly used for GEO, assume 2 ignitions per flight
+// 31 ignitions, 0 failures
+// 31 cycles, 1 failures
+// 30 restarts, 30 failures
```
NathanKell, line 246: 30 restarts, 30 failures!?

NathanKell:
That, but also IIRC the issue with the early LR79 was a known turbopump issue and some of those failures were intentionally rolling the dice (because Schriever didn't want to take missiles/LVs out of service to repair and eat the delay)--or at least that's what I recall from the Pioneer report. So while I think it's quite reasonable to *start* the MB-I with low reliability, it seems reasonable to offer a higher reliability cap.

On another note, what do you think about tech transfer? I am coming to the belief that for later models of engines, the starting reliability should be set such that the reliability *post transfer* matches the computed starting reliability. Example: Agena-D had very few failures. But part of why is because of the inherited knowledge from A and B. That's represented ingame by starting at DU=2000, rather than DU=0. So I think the reliability *at 2000 DU* should be your computed starting value of 0.989, rather than as it is now/in this PR where the starting reliability at 2000DU is 75% of the way from 0.989 to 0.9978, i.e. 0.9956.

Capkirk123:
Sure, I can adjust the reliability of MB-I so end reliability is a bit better at least.

As for starting reliability, that also probably makes some sense. At the moment, starting reliability is a bit worse than IRL, and ending is better, so you should experience failure rates similar to real life.

NathanKell:
Yeah, I think that makes sense with MB-I.

To be clear, the point I was making about starting reliability was a general point for all engine configs that have tech transfer, though, not purely about MB-I. It's particularly noticeable for late-model highly-reliable configs like Agena-D, AJ10-137, H-1-205K, J-2-230K, etc.

Capkirk123:
Yeah I get that. It shouldn't be too hard to do

Capkirk123:
Ok, pretty much everything with data transfer has had it's starting reliability decreased.

ec429:
@Capkirk123 if this gets merged could you take a look at updating the corresponding SPE configs with the same numbers?  (They mostly state in the `TESTFLIGHT` node which engine they're based on.)

siimav:
Has conflicts now

Capkirk123:
Should be good to go

PR #2546, opened 2021-11-13
Fix VSR filepath for 1.12

ZombieZilla:


GameData/RealismOverhaul/RO_DependentMods/RO_AJE.cfg
```diff
@@ -115,7 +115,7 @@
 
 	MODEL
 	{
-		model = VenStockRevamp/Part Bin/NewParts/JetEngines/HighBypassJet
+		model = VenStockRevamp/PartBin/NewParts/JetEngines/HighBypassJet
```
lpgagnon, line 118: Maybe support both old and new path using NEEDS? RO already does this in a few places https://github.com/KSP-RO/RealismOverhaul/search?q=partbin

(the path switch was actually to help us, because NEEDS doesn't work with spaces)

DRVeyl: Prefer not to carry forward supporting multiple versions of a mod in general, especially for a change like this.  Get newest, resolve problem.  Looks like I missed one reference when I did the earlier conversion.  (Or maybe something got added in after.  No big deal.)

lpgagnon:
> get newest, resolve problem

except that
> [siimav] Ugh, CKAN says that the release is 1.12 but on the releases page in the repo it says 1.10-1.12

https://discord.com/channels/319857228905447436/620690446540341261/909058994071556167

DRVeyl:
Easily solved in the CKAN meta, and apparently an error in the .version file.  The VSR path changes have been running & tested on 1.10+ for a long time now.

PR #2548, opened 2021-11-14
Add J57-P-8, RD-9B; adjust size and CoM offset of jets

Capkirk123:
Add the J57P8 and RD-9B so the 1951 node actually has some supersonic capable engines.
Adjust size and CoM offset of jets.


siimav:
The new J57 has so low thermal limits that it explodes during takeoff

Capkirk123:
> The new J57 has so low thermal limits that it explodes during takeoff. edit: is [1f64fc7](https://github.com/KSP-RO/RealismOverhaul/commit/1f64fc75751d86ae7d79674a2d07d88815010cfa) the fix for that?
Yes, the J57 thermal limit has been raised to be a bit more manageable

siimav (approved):

PR #2550, opened 2021-11-16
MIGHT BREAK SSPX CRAFT - Fixed typos in existing configs, added new configs for new sspx parts

mike-gravley:
Fixed various typos in the existing configs and finished up some unfinished configs. Also added new configs for all the new parts that were recently added to SSPX, including the new 1.875m station parts, the new 5m station parts, and the new 5m geodesic ground domes. Scaled the 5m parts by 1.66 to result in 8.3m, and scaled the 1.875m parts by 2 to result in 3.75m. This was done to keep them between the original 1.25m and 2.5m parts, which were scaled by different factors (2.44 vs 1.66). With the addition of the 1.875m parts, I personally think all parts from SSPX should now be scaled by 1.66. I understand why the original 1.25m parts were scaled up to 3.05m initially, but with the new parts added at 1.875m, those could now cover 3.1125m parts if scaled by 1.66. If everything was scaled by 1.66, this would result in 2.075m, 3.1125m, 4.15m, 6.225m, and 8.3m parts, and all the adapters would then properly work between sizes. I had not done that yet, as that would definitely break any crafts that currently use the 3.05m scaled parts. Let me know your thoughts on making the scaling consistent!

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_375m_Parts.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_375m_Parts.cfg
@@ -1,0 +1,272 @@
+//	=================================================================
+//	3.75m Modules - STOCKALIKE STATION PARTS EXPANSION REDUX
+//
+//	sspx-attach-1875-1  	PMA-AR Radial Attachment Module
+//	sspx-core-1875-1		PMA-2 'Quay' Station Core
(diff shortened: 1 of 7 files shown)


pap1723:
This shouldn't be needed anymore now that the other SSPX patch was merged.

pap1723:
> This shouldn't be needed anymore now that the other SSPX patch was merged.

The other PR was chosen to merge instead because it contained a PR for RP-1 as well as RO-Kerbalism.

PR #2561, opened 2021-12-20
Sspx update

Tagir-1:
**Might break existing crafts or change their proportions!**
- New, unified scaling system. Every part is scaled to 1.8 from stock. SPPX-remodeled stock parts (Hitchhiker, Lab, Cupola) are affected too.
- Masses and tank volumes are rescales accordingly
- Latest SSPX parts, 1.875m (3.38m in RO) and 5m (9m in RO) are now supported
- Extendable habitats are now supported.
- Also, see Kerbalism and RP-1 changes

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_225m_Parts.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_225m_Parts.cfg
@@ -32,7 +32,7 @@
 	@mass = 0.5
 }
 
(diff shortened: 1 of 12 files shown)


JohnMcLane4:
..../RO_SSPX_900m_Parts.cfg
line 223, there is a colon at the end. I think it doesnt belong there....

TheDestroyerKSP:
The 9m logistics modules appear to be suspiciously light for a high pressure tank - here's my tests:

6.75m (long) storage module full of argon: 26t dry, 72t wet

Equivalent tank, AliLi-HP: 22t dry, 72t wet

(seems correct)

9m logistics module (long) full of argon: 20t dry, 340t wet

Equivalent tank is twice the size (appears to have more volume than it should) and is 100 tons dry, 340t wet.

PR #2580, opened 2022-01-14
Correct X-405, Stentor sea-level thrust

### LINKED ISSUE #2579: Vacuum thrust set as listed sea level thrust for X-405, Stentor booster
Per the outlined statistics in the commented portion of the config file for the X-405, its vacuum thrust is `135.5` kN (~30,460 lbf). In the `CONFIG` block for the engine, its `maxThrust` is given as `123.6` kN (~27,785 lbf) with a comment of 27,800 lbf at sea level - however, the `maxThrust` value applies to vacuum thrust, not sea level. Current sea level thrust is `112.9` kN. Assuming precise Isp values, a vacuum thrust of `135.28` kN (~30,410 lbf) would return the same `123.6` at sea level. Prior to the most recent commit to the engine (#2506), it had the listed `135.5` kN vacuum thrust (yielding `123.8` kN sea level, ~27,830 lbf), so it seems to have got lost in the mix? Currently the TWR of an in-game vanguard is only around 1.14, which is quite a bit lower than the ~1.25 it should have.

The Stentor booster suggests sea level thrust of `110` kN (~24,730 lbf), but delivers `100` kN (~22,480 lbf).  Assuming that is the intended sea level thrust and the specific impulse (200/220) is correct (the commented portion does not give a sea level Isp, and the two listed sources - Wikipedia & A Vertical Empire, the former of which cites the latter - do not specify whether given Isp is vacuum or sea-level) a thrust of `121` kN would yield `110` kN sea level. 
At a quick glance, it appears unclear as to what the thrust "should" be for the Stentor -  A Vertical Empire gives it generic thrust between 24,000 lbf and 25,000 lbf, stating 25,500 at 45,000 feet (at "around 220" Isp), and gives the derived PR-27 between 22,500 and 25,000 lbf per chamber, with no altitude remarks whatsoever. Purely according to what is contained immediately within the outline at the top of the file, however, a vacuum thrust of `121` kN would deliver the outlined performance.

ilmcamam:
Fixes #2579.
X-405 values should be in-line with outlined performance (27,800 lbf SL, 254/278 Isp). Stentor lists 110 kN thrust SL but sets vac as 110, this corrects for that, regardless of the accuracy of 110 kN SL for the stentor.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Stentor_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Stentor_Config.cfg
@@ -64,8 +64,8 @@
 		{
 			name = Stentor
 			description = Booster for the Blue Steel missile
-			minThrust = 110
-			maxThrust = 110
+			minThrust = 121	// 110 kN (~25klbf) at sea level
+			maxThrust = 121
 			heatProduction = 100
 			massMult = 1.0
 
--- a/GameData/RealismOverhaul/Engine_Configs/X405_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/X405_Config.cfg
@@ -8,8 +8,8 @@
 //	Vangaurd
 //
 //	Dry Mass: 192 Kg
-//	Thrust (SL): ??? kN
-//	Thrust (Vac): 135.5 kN
+//	Thrust (SL): 123.6 kN		// 27,785 lbf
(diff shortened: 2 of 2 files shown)


Capkirk123:
Although theses values appear to disagree with G. Sutton, they do match up better with the X-405H numbers. These are probably more correct.

PR #2587, opened 2022-01-24
Engine config for Filament Wound Casing RSRM

StonesmileGit:


Capkirk123:
If the mass savings are per-booster, the mass ratio is still worse than other large composite motors like SRMU. Seems reasonable enough.

PR #2588, opened 2022-01-28
Update Resources, add new Hydrocarbons

Capkirk123:
- Update the cost of resources based on the DoD fluids pricing guide
- Add RP-1, RG-1, and update the properties of kerosene to better match JP-4/TS-1
- Add subcooled versions of kerosene, RP-1, RG-1, Syntin, and Liquid Oxygen
- Increase the boiling point of liquids when stored under pressure

These changes are probably save-breaking


Capkirk123:
Running tests to see what kind of difference RG-1 vs normal kerosene makes, there are about 0.6% gains in mass limited scenarios, or 0.9% gains in volume limited scenarios, with Al2 tankage. This drops to about 0.4% and 0.8% as tankage improves.

NathanKell:
This would also be the time to do the pressurant change you talked about before I think--where you were adding high-pressure resources and the engines would convert from high to low? Since that is also a save-breaking resource change.

Capkirk123:
Sure, but that would require a lot more than some RO config tweaks. That needs new functionality from RF to achieve.

> This would also be the time to do the pressurant change you talked about before I think--where you were adding high-pressure resources and the engines would convert from high to low? Since that is also a save-breaking resource change.

NathanKell:
> Sure, but that would require a lot more than some RO config tweaks. That needs new functionality from RF to achieve.
Isn't the only thing it needs adding the tank of appropriate size (but empty) when a propellant has negative ratio? Or am I forgetting something else?

Capkirk123:
> > Sure, but that would require a lot more than some RO config tweaks. That needs new functionality from RF to achieve.
> 
> Isn't the only thing it needs adding the tank of appropriate size (but empty) when a propellant has negative ratio? Or am I forgetting something else?

Yes, but that might be more complex than it seems. Handling the refilling of an empty tank, for example, might prove difficult.

Capkirk123:
> @Capkirk123 what do you mean? Are you talking about launch clamps autofilling?

I was thinking more about mid flight refueling, like from internal cross feed on the shuttle. You need some way to remove the low pressure gas from the tank and replace it with propellants. Determining the behavior of a partially filled tank is also an issue (you’d want any empty tank volume to be backfilled with gas, but also remove the gas if you try to fill the tank the rest of the way).

NathanKell:
Replace? I'm not following. We were, I thought, talking about a setup like this:
MMH (full)
NTO (full)
HeliumHP (full)
HeliumLP (empty)

with a positive ratio on HeliumHP and a negative ratio on HeliumLP, such that burning the engine removed HeliumHP and created HeliumLP.

The only issue would be refueling, yes, where you would need to dump the HeliumLP prior to refueling. But that can be done with ShipManifest, so I wouldn't hold off on making savebreaking changes on that account.

NathanKell:
Leaving a comment here to remind me to rename Tank-Int to Tank-Iso when we merge this and break saves.

NathanKell:
Making a new PLC branch to stick this and the X-405 one (and anything else save-breaking you got) in. Can you fix the conflicts?

NathanKell:
Ooof there's a lot here. I'm noping out for now, sorry.

Capkirk123:
Yeah this is gonna need to be transferred over to everything I’ve added in the past year or so, plus I’ve discovered a lot of other resources have just plain wrong densities set in CRP. I’ll probably redo the whole thing with everything else I’ve discovered in the past year.

NathanKell:
Ok, that works. Gonna close this then 👍

PR #2591, opened 2022-02-04
Add tagging system for ModuleCargoPart

siimav:
Every part that doesn't have ROCargoConfig tag will be stripped of it's ModuleCargoPart PM.


GameData/RealismOverhaul/RO_Cargo.cfg
```diff
@@ -0,0 +1,11 @@
+// Unless the part has ROCargoConfig tag, removes the partmodule
+// that is responsible for the part being usable in the stock inventory system.
+@PART:HAS[~ROCargoConfig,@MODULE[ModuleCargoPart]]:LAST[RealismOverhaul]
```
al2me6, line 3: Hm do we want `~ROCargoConfig[?rue]` for more flexibility?

PR #2595, opened 2022-02-18
Add MLI PARTUPGRADES

Capkirk123:
Use the RF Part upgrades system to allow for MLI technology upgrades over time, rather than tying MLI to tank type.

RP-1 PR: https://github.com/KSP-RO/RP-0/pull/1632

### DIFF
--- a/GameData/RealismOverhaul/RO_RealFuels.cfg
+++ b/GameData/RealismOverhaul/RO_RealFuels.cfg
@@ -14,6 +14,7 @@
 //	the new tooling mechanic and to create a proper mass ratio progression.
 //	===============================================================================
 
+//maxMLILayers will be defined by PARTUPGRADEs rather than by tank
 TANK_DEFINITION
 {
 	name = Tank-Sep-Steel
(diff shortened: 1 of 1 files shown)


DRVeyl:
You don't want to remove all of the MLI settings configs in RO and then require folks to use RP-1 to actually get them configured.  PARTUPGRADEs can be applied in sandbox -- and anyone using ProcParts or ProcFairings already needs to use that setting.  Which means RO should also make use of the PARTUPGRADEs if you wish to change the configurability of the MLI values.

IMO most of the PR into RP-1 should move into RO, and RP-1 should just be concerned with setting the tech node and entry cost for the existing PARTUPGRADE.

Capkirk123:
RF already sets part upgrades. If you use this without RP-1, it'll just use the stock part upgrades instead.

DRVeyl:
I see.  I was confused because I saw some of the values also change in here.  Just delete the lines rather than commenting them out and make sure the merge commit message includes "these values will be set by the upgrades and don't need to be configured here anymore"

PR #2599, opened 2022-02-25
Add SolarConfig support for multiple identical solar modules

StonesmileGit:
Also adds support to tag a module with `realismOverhaulSolarIgnore` to make the patch tool ignore the module

### DIFF
--- a/GameData/RealismOverhaul/RO_SolarConfig.cfg
+++ b/GameData/RealismOverhaul/RO_SolarConfig.cfg
@@ -102,7 +102,7 @@
 	{
 		@kgm2 *= #$massMult$
 	}
-	
+
 	@SolarConfig:HAS[#type[hinged]]
 	{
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SolarConfig.cfg
```diff
@@ -145,6 +145,10 @@
 	!SolarConfig {}
 	!solarUseMass = DEL
 	!useSolarConfig = DEL
+	@MODULE[ModuleDeployableSolarPanel]
```
al2me6, line 148: Need `,*`

al2me6 (changes requested):

GameData/RealismOverhaul/RO_SolarConfig.cfg
StonesmileGit: Fixed

PR #2600, opened 2022-03-06
Add more resource boiling points, move resource prices to RP-1

Capkirk123:
Adjust resource prices based on the DoD fluid pricing guide. Raise boiling points of pressurized fluids, and add boiling points to water and ethanol.

### DIFF
--- a/GameData/RealismOverhaul/RO_Resources.cfg
+++ b/GameData/RealismOverhaul/RO_Resources.cfg
@@ -18,6 +18,7 @@ RESOURCE_DEFINITION
 {
   name = Aniline22
   density = 0.001042
+  unitCost = 0.0001 //placeholder
   hsp = 2163 // specific heat capacity (kJ/tonne-K as units)
   flowMode = STACK_PRIORITY_SEARCH
   transfer = PUMP
@@ -29,6 +30,7 @@ RESOURCE_DEFINITION
 {
   name = Aniline37
   density = 0.0010585
+  unitCost = 0.0001 //placeholder
   hsp = 2150 // specific heat capacity (kJ/tonne-K as units)
   flowMode = STACK_PRIORITY_SEARCH
   transfer = PUMP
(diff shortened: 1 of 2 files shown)


GameData/RealismOverhaul/RO_Resources.cfg
```diff
@@ -38,54 +38,81 @@ RESOURCE_DEFINITION
...
+// Source:
+// https://www.dla.mil/Portals/104/Documents/Energy/Standard%20Prices/Aerospace%20Prices/E_2016Oct1AerospaceStandardPrices_160921.pdf
+// https://www.dla.mil/Portals/104/Documents/Energy/Standard%20Prices/Aerospace%20Prices/E_2017Oct1AerospaceStandardPrices_170913.pdf?ver=2017-09-13-145335-477
 @RESOURCE_DEFINITION[LqdOxygen]:FOR[RealismOverhaul]
 {
-  @unitCost = 0.00004564
+  //$22.38/US ton, 1965$
+  //@density = 0.001141
+  @unitCost = 0.00002815
 }
```
DRVeyl, line 49: RO setting unit costs seems a little weird.  Well, I guess it has to be defined *somewhere*.
About the comments "High because of enviromental regulations, divide by 10" did we inflate the cost by 10x its market value?  That sounds like something more in the domain of RP-1 to do than RO.

Comment applies throughout.

Capkirk123: Due to various reasons, including environmental regulations and the production lines being shut down, propellant grade hypergolics are now incredibly expensive. Since what limited info I can find shows they were much cheaper in the 60s, and we’re mostly focused on the 60s, I decided to divide by 10, which gave a cost similar to 60s costs.

I defined the cost in RO because RO was already setting costs.

Capkirk123: Ok, all cost settings have been moved to RP-0.

DRVeyl (approved):

PR #2605, opened 2022-03-13
Add SAF config tool

StonesmileGit:
Currently rescales sizes, but not mass of SAF fairings

### DIFF
--- a/GameData/RealismOverhaul/RO_SAFConfig.cfg
+++ b/GameData/RealismOverhaul/RO_SAFConfig.cfg
@@ -4,7 +4,7 @@
 /// 	%ROSAFRescale = 1.81 //Make everything 1.81 times bigger
 /// }
 
-@PART[*]:HAS[@MODULE[ModuleSimpleAdjustableFairing]:HAS[#ROSAFRescale]]:AFTER[RealismOverhaul]
+@PART:HAS[@MODULE[ModuleSimpleAdjustableFairing],#ROSAFRescale]:AFTER[RealismOverhaul]
 {
 	@MODULE[ModuleSimpleAdjustableFairing],*
 	{
@@ -78,7 +78,7 @@
 }
 
 // Delete the rescale tag
-@PART[*]:HAS[@MODULE[ModuleSimpleAdjustableFairing]]:FOR[zzzTagCleanup]
+@PART[*]:HAS[#ROSAFRescale]:FOR[zzzTagCleanup]
 {
 	!ROSAFRescale = null
 }
--- a/GameData/RealismOverhaul/RO_SAFConfig.cfg
+++ b/GameData/RealismOverhaul/RO_SAFConfig.cfg
@@ -1,49 +1,46 @@
 /// Example config - Replace 'MyPartWithSAFModule' with your part name
 /// @PART[MyPartWithSAFModule]:FOR[RealismOverhaul]
 /// {
-/// 	@MODULE[ModuleSimpleAdjustableFairing]
-/// 	{
-/// 		ROSAFRescale = 1.81 //Make everything 1.81 times bigger
-/// 	}
+/// 	%ROSAFRescale = 1.81 //Make everything 1.81 times bigger
 /// }
 
 @PART[*]:HAS[@MODULE[ModuleSimpleAdjustableFairing]:HAS[#ROSAFRescale]]:AFTER[RealismOverhaul]
 {
-	@MODULE[ModuleSimpleAdjustableFairing]
+	@MODULE[ModuleSimpleAdjustableFairing],*
 	{
 		&scale = 1
-		@scale *= #$ROSAFRescale$
-		@segmentLength *= #$ROSAFRescale$
-		@shieldingCenter[*] *= #$ROSAFRescale$
-		@shieldingBaseRadius *= #$ROSAFRescale$
-		@editorOpenOffset[*] *= #$ROSAFRescale$
+		@scale *= #$/ROSAFRescale$
+		@segmentLength *= #$/ROSAFRescale$
+		@shieldingCenter[*] *= #$/ROSAFRescale$
+		@shieldingBaseRadius *= #$/ROSAFRescale$
+		@editorOpenOffset[*] *= #$/ROSAFRescale$
 
 		@WALL_BASE
 		{
-			@CoM[*] *= #$../ROSAFRescale$
-			@rootOffset[*] *= #$../ROSAFRescale$
+			@CoM[*] *= #$/ROSAFRescale$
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SAFConfig.cfg
```diff
@@ -0,0 +1,90 @@
+/// Example config - Replace 'MyPartWithSAFModule' with your part name
+/// @PART[MyPartWithSAFModule]:FOR[RealismOverhaul]
+/// {
+/// 	@MODULE[ModuleSimpleAdjustableFairing]
+/// 	{
+/// 		ROSAFRescale = 1.81 //Make everything 1.81 times bigger
+/// 	}
+/// }
+
+@PART[*]:HAS[@MODULE[ModuleSimpleAdjustableFairing]:HAS[#ROSAFRescale]]:AFTER[RealismOverhaul]
```
al2me6, line 10: Doesn't RO normally put these things in the part itself?

GameData/RealismOverhaul/RO_SAFConfig.cfg
```diff
@@ -0,0 +1,90 @@
...
+		{
+			@CoM[*] *= #$../ROSAFRescale$
+			@rootOffset[*] *= #$../ROSAFRescale$
+		}
+
+		@CAP
+		{
+			@CoM[*] *= #$../ROSAFRescale$
+			@rootOffset[*] *= #$../ROSAFRescale$
+		}
```
al2me6, line 43: We're sure that all of these things exist? (Though if we aren't it gets a lot more annoying to patch...)

GameData/RealismOverhaul/RO_SAFConfig.cfg
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	@MODULE[ModuleB9PartSwitch]:HAS[#moduleID[fairingSwitch]]
```
al2me6, line 46: This `HAS` is fragile. It might work for most BDB things, but not always. I guess you can match on having subtypes with SAF switches inside.

GameData/RealismOverhaul/RO_SAFConfig.cfg
```diff
@@ -0,0 +1,90 @@
...
+	{
+		@SUBTYPE,*
+		{
+			@MODULE:HAS[@IDENTIFIER[ModuleSimpleAdjustableFairing]]
+			{
+				@DATA
+				{
+					@WALL_BASE
+					{
+						@CoM[*] *= #$../../../../../MODULE[ModuleSimpleAdjustableFairing]/ROSAFRescale$
```
al2me6, line 56: `/MODULE[...]/ROSAFRescale`

(`/` starts from the root of the part, `@` starts from the top-level. No need for `../` spam.)

al2me6 (changes requested):

GameData/RealismOverhaul/RO_SAFConfig.cfg
al2me6: ...hmm that might catch the wrong MODULE if there are multiple. Not sure if `:HAS` is allowed in a path off the top of my head.

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: If they do not exist, scaling them would mean nothing

al2me6: If they don't exist, using `@` would result in errors.

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: Supporting multiple modules seems difficult, might leave that for when it's needed.

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: Moved tag to part level

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: `@` tag fails silently, according to DRVeyl

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: Removed `HAS`

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: Made the patch apply on all SAF modules and all B9PS modules

GameData/RealismOverhaul/RO_SAFConfig.cfg
```diff
@@ -0,0 +1,84 @@
+/// Example config - Replace 'MyPartWithSAFModule' with your part name
+/// @PART[MyPartWithSAFModule]:FOR[RealismOverhaul]
+/// {
+/// 	%ROSAFRescale = 1.81 //Make everything 1.81 times bigger
+/// }
+
+@PART[*]:HAS[@MODULE[ModuleSimpleAdjustableFairing]:HAS[#ROSAFRescale]]:AFTER[RealismOverhaul]
```
DRVeyl, line 7: This test does not match the comment.
Comment puts the tag at the PART level, but this patch looks for it inside the PartModule.
You want `@PART:HAS[@MODULE[ModuleSimpleAdjustableFairing],#ROSAFRescale];AFTER[RealismOverhaul] {...}`

GameData/RealismOverhaul/RO_SAFConfig.cfg
```diff
@@ -0,0 +1,84 @@
...
+			}
+		}
+	}
+}
+
+// Delete the rescale tag
+@PART[*]:HAS[@MODULE[ModuleSimpleAdjustableFairing]]:FOR[zzzTagCleanup]
+{
+	!ROSAFRescale = null
+}
```
DRVeyl, line 84: `:HAS[#ROSAFRescale]` not the `@MODULE`.
Also add newline before EOF.

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: Fixed with latest commit, forgot to change that line when I moved the tag to the part level

GameData/RealismOverhaul/RO_SAFConfig.cfg
StonesmileGit: Fixed

PR #2612, opened 2022-03-28
Fix cfg syntax errors

HebaruSan:
Hi, KSP-RO team!

I'm using RealismOverhaul to test some cfg parser code for KSP-CKAN/CKAN#3525, and I found a few minor syntax errors.

This pull request fixes them.

Cheers!


DRVeyl:
That's pretty amazing.
I'll look over them in the next day or so.  Thanks!

NathanKell:
Echoing DRVeyl, this is awesome, thank you so much!! :)

GameData/RealismOverhaul/REWORK/RO_CZ3B.cfg
```diff
@@ -440,9 +440,6 @@
 	%title = YF-75 (2x) [3 m]
 	%manufacturer = CALT
 	%description = Chinese gas generator cryogenic upper stage engine, used on the Long March series of rockets in a double configuration. Can restart up to two times.
-	{
```
NathanKell, line 443: I think this is missing a @MODEL, rather than the node needing to be deleted.

GameData/RealismOverhaul/REWORK/RO_KOSMOS.cfg
```diff
@@ -1000,7 +1000,6 @@
 		name = ModuleEngineConfigs
 		type = ModuleRCS
 		thrustRating = resourceFlowMode = STACK_PRIORITY_SEARCH
```
NathanKell, line 1002: the correct form here is `thrustRating = thrusterPower` and then the `resourceFlowMode` line.

GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_RealPlume/BDB_Atlas_Classic.cfg
```diff
@@ -81,7 +81,6 @@
             fixedEmissions = false
             sizeClamp = 50
             randomInitalVelocityOffsetMaxRadius = 0
-            randConeEmit
```
NathanKell, line 84: Someone who understands plume code (@al2me6 ?) needs to look at these, I'm not sure it's correct to nuke the value rather than do something with it

GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_RealPlume/BDB_Saturn_Liquid_new.cfg
```diff
@@ -359,7 +359,6 @@
             fixedEmissions = false
             sizeClamp = 50
             randomInitalVelocityOffsetMaxRadius = 0
-            randConeEmit
```
NathanKell, line 362: ditto

GameData/RealismOverhaul/RO_SuggestedMods/RO_B9.cfg
```diff
@@ -675,7 +675,7 @@
 	{
 		@name = AJEInlet
 		Area = 1.2
-		#@AJE_TPR_CURVE_DEFAULTS/FixedCone/TPRCurve {}
+		//@AJE_TPR_CURVE_DEFAULTS/FixedCone/TPRCurve {}
```
NathanKell, line 678: I think this was maybe trying to copy a node rather than be a comment. @DRVeyl or @al2me6 will know the syntax for sure.

NathanKell (changes requested):

NathanKell:
Oh maaaan some of these are real howlers. Again, thanks so much for catching them! (And we probably should add your checker to our actions so we can check PRs in future!!)

I found a few things that needed to be changed (and a few others that I had to tag @DRVeyl and @al2me6 on since they will know where I don't!) but everything else LGTM.

GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_338m_Parts.cfg
```diff
@@ -65,8 +65,7 @@
 	
 	@MODULE[ModuleProbeControlPoint] {}
 	@MODULE[ModuleCommand] {}
-	@MODULE[ModuleDataTransmitter] {}
```
StonesmileGit, line 68: These rows all look weird to begin with, not sure if this is the intended behaviour

HebaruSan:
> And we probably should add your checker to our actions so we can check PRs in future!!

I like that idea a lot, but unfortunately the parser is not quite ready for prime time yet. The error messages that it produces are still very poor quality; it only tells you the first line of the top level config node containing the problem, and then a human has to inspect that node and all of its values and child nodes, potentially dozens of lines, to find the problem by eye. I need to do some more research into monadic parsers to figure out how to make it pinpoint problems more precisely, but if I do, I'll look into making it available for a GitHub Action (the current code is buried deep within some CKAN routines that aren't easily applied to other purposes).

GameData/RealismOverhaul/REWORK/RO_CZ3B.cfg
HebaruSan: OK, I'll try it that way instead. Is this what you had in mind?

```
	@MODEL
	{
		%scale = 1, 1, 1
	}
```

GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_338m_Parts.cfg
DRVeyl: I don't know what this is trying/intending to do, either.  Should those be `%MODULE` to guarantee they exist?  Or are they just NOPs to be deleted?
Why are we NOPing a ModuleCommand and then patching an MDT to be a ModuleCommand?
Are there supposed to be *2* ModuleFuelTanks on here?  Or just 1?  (This has a weird pattern of creating a node and then immediately patching it).
@pap1723 ?

HebaruSan:
While I've got all you wonderful experts gathered around this cfg syntax campfire, maybe this would be a good time to ask about this `*@` syntax:

https://github.com/KSP-RO/RealismOverhaul/blob/e7fc1512acc0e2ef412ddddfc0e95d0730754365/GameData/RealismOverhaul/RO_Physics.cfg#L91-L94

https://github.com/KSP-RO/RealismOverhaul/blob/e7fc1512acc0e2ef412ddddfc0e95d0730754365/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_NK_M55.cfg#L30

My parser doesn't handle this at all, and it looks too intentional to be a typo, but I haven't been able to find anything about it in my ModuleManager syntax references:

- https://github.com/sarbian/ModuleManager/wiki/Module-Manager-Syntax
- https://github.com/sarbian/ModuleManager/wiki/Module-Manager-Handbook
- https://github.com/sarbian/ModuleManager/wiki/Patch-Ordering

Can anyone share any links explaining how this `*@` thing works? I can make some guesses based on how it's used here, but I'd rather go by explicit documentation if any is available.

al2me6:
@HebaruSan AFAIK this is the only official reference for the feature that exists: https://forum.kerbalspaceprogram.com/index.php?/topic/50533-18x-112x-module-manager-421-august-1st-2021-locked-inside-edition/page/136/#comment-2413546

GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_RealPlume/BDB_Atlas_Classic.cfg
al2me6: Hm I didn't get notified for some reason. Should be fine to just nuke.

al2me6:
> Huh. Any idea how MM tells the difference between this use of `#` and the common use of it in translation tokens?

Node name vs key name.

GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_AtlasV.cfg
```diff
@@ -14,7 +14,7 @@
 	%thermalMassModifier = 1.0
 	%skinMassPerArea = 4
 
-	!RESOURCE, * {}
```
al2me6, line 17: I'm not entirely sure that this is invalid. Regardless it's good to normalize.

GameData/RealismOverhaul/RO_SuggestedMods/RO_B9.cfg
HebaruSan: Thanks, I have just now learned about the "paste" operator. :+1:

GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_AtlasV.cfg
HebaruSan: That was my thought as well. MM is pretty permissive in what it handles, so I'm not certain that it wouldn't handle `, *`, but I do know for sure that `,*` **is** valid.

GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_338m_Parts.cfg
HebaruSan: Maybe the intention was to delete `ModuleCommand` (rather than edit) and then create a new one (i.e., add `MODULE` before the `{`)?

HebaruSan: Semi-confirmed, that's close to what the `225m` part does:

https://github.com/KSP-RO/RealismOverhaul/blob/743d917814d5bbbad015404a2bcc013776ad1f6d/GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_225m_Parts.cfg#L53-L61

DRVeyl:
Two `ModuleCommand` is wrong, functionally.  But fixing that isn't really in scope for this PR.  You found something where we were doing something syntactically broken that happened to be right next to where we were doing something logically dumb.  I'll go through the SSPX patches [that mistake is part of a pattern there] separately after merging this PR.  So I'm not worried about the state of the SSPX patches in this PR.

GameData/RealismOverhaul/RO_SuggestedMods/StationPartsExpansion/RO_SSPX_338m_Parts.cfg
DRVeyl: Thanks.  TODO for @DRVeyl : Fix the SSPX configs with a more appropriate pattern, they're making this mistake everywhere in addition to only a couple of the files having invalid syntax.

DRVeyl (approved):
Ok, looked over everything.  Modulo identifying a new issue regarding SSPX config pattern, I think everything in here has been addressed.

NathanKell (approved):

HebaruSan:
> (And we probably should add your checker to our actions so we can check PRs in future!!)

Preliminary support for this now exists, although we are still working on several things, and it may break without warning:

<https://github.com/KSP-CKAN/KSPMMCfgParser>

PR #2619, opened 2022-03-31
Split X-405H proposals

### LINKED ISSUE #1819: X-405-H - adjustments and upgrades

Capkirk123:
Split X-405H proposals into one config without restart, and one with restart. Also adjust mass a bit based on Convair documents.

After rereading the Convair final report, it seems that while restart was eliminated from Atlas-Vega due to the presence of the third stage, the engine still possessed the capability, and there was still interest in a two-stage Atlas-Vega with restart capability. I split the X-405H into two configs, the basic X-405H without restart as was to be flown on the first Atlas-Vega launches, and the X-405H-2 with restart, with mass guesstimated based on mass breakdowns of the 3-stage and 2-stage proposals.

Fixes #1819


DRVeyl (approved):
Looks good to me.  Is slightly breaking in that things using X-405H expecting the re-ignitions now don't have it.

DRVeyl:
Should this close <https://github.com/KSP-RO/RealismOverhaul/issues/1819>?

Capkirk123:
Assuming the Convair Final Report is accurate (which it appears to be), then yes. The question of ignitions and engine performance are answered there.

Capkirk123:
Closing this, since I broke something with git

PR #2622, opened 2022-04-03
BDB Updates

Vieju17:
This request aims to update some of the old RO configs for Bluedog Design Bureau. The changes are:

1. Updating the part name of the Centaur D configs so the rescaling and properties for RO are applied. 

2. Changing the rescale factor of the Able parts from 1.3 to 1.6 so they match and can be used with the other launch vehicles. 

3. Adding RO configs to all SAF fairings, including the ROSAFRescale factor made by @StonesmileGit. 

I have tested everything with my install and it seems to work fine. Nevertheless, I'm new to all of this so there's a high possibility that I have missed something. Apologies in advance if that's the case.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_Avionics.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/RO_BDB_Avionics.cfg
@@ -359,7 +359,7 @@
 @PART[bluedog_ThorAble_Guidance]:NEEDS[RealismOverhaul]
 {
 	%RSSROConfig = True
-	%rescaleFactor = 1.3
+	%rescaleFactor = 1.6
 	
 
 	@title = Able Avionics
@@ -399,7 +399,7 @@
 
 // CENTAUR
 
-@PART[bluedog_Centaur_Avionics]:NEEDS[RealismOverhaul]
+@PART[bluedog_CentaurD_Avionics]:NEEDS[RealismOverhaul]
 {
 	%RSSROConfig = True
 	%rescaleFactor = 1.6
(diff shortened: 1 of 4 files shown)


DRVeyl:
Neat.

One thing, the re-ordering in the SAF file makes the changes very hard to read.  (The change tracking ends up basically showing almost everything as rewritten, because it interprets DELTA parts as having been replaced by ABLE, and then ABLE parts as having been replaced by DELTA).  Some time later I can probably just look at the file overall: the new SAF tool should make this a lot easier.

Question: are there parts that used to have patches, but no longer do after this change?
Do those parts still exist in recent (1.12-supporting) BDB releases?

Looking over this, I see that apparently our BDB configs are old enough that they didn't already have any pass specifier (ie `:FOR[RealismOverhaul]`) and ran in default.  Ick.

I know there are a couple efforts to update our BDB configs.  BDB is such a large [and active] mod to track, I think there is some focus on splitting the patches out to follow individual lines of BDB parts [ie a set of patches for Able parts, then a set of patches for Delta, all in different folders].  Can anyone else shed some light on where that effort is?

Vieju17:
Regarding the re-ordering, I've noticed that. Is it critical? If so it can be changed. 

As for your question, I'm 99% sure it doesn't affect parts that had configs previously because before applying the change I had no Centaur D appearing in my game with RO configs. It must have been a matter of the creator of BDB changing the name of the part. Also, yes, I'm working with the latest BDB version. 

Shall I add the identifier? 

No idea about such efforts, it'd be nice if someone commented about it.

However, I think I'm going to close this request since some checks have failed and I suspect what the issue is. I'll keep it open if you want to respond to this and once I open the new request I'll close this one.

StonesmileGit:
> some checks have failed

This is most likely not your fault, and caused by the build script in the repo only working for PRs originating from within the repo

DRVeyl:
This is fine to leave open, and you can always commit any fixups atop this branch, or force-push to yours.  We usually squash-merge PRs if the commit history is a bit ugly, so that's also fine.

Mostly I was trying to ask... are there parts that the file was patching before that aren't patched now, and do those parts still exist or did they all go away/get renamed?  (This is related to the comment about the change tracking: it'd be a lot easier to see if parts went away or got added if they weren't moved throughout the file.  So the re-ordering isn't critical, it just will make any future PRs you do easier to evaluate.)

For the main file that you rewrote, I'd say please add `:FOR[RealismOverhaul]` to the patches.  But going through and adding it to other files isn't on you to do.

[What Stonesmile said about the build checks.]

PR #2624, opened 2022-04-04
Fix :FOR[RealPlume] in RealismOverhaul

HebaruSan:
Hi KSP-RO team,

RO has a few hundred `:FOR[RealPlume]` clauses in it at the moment. Module Manager interprets this to mean that RealPlume is installed; other mods' patches with `:NEEDS[RealPlume]` would be activated even if RealPlume is absent. RO depends on RealPlume, so there should be no impact as long as the user's install is valid, but it's still better to avoid this:

https://github.com/sarbian/ModuleManager/wiki/Patch-Ordering#the-beforemodname-formodname-and-aftermodname-directives

> It is not recommended to use the :FOR directive to refer to other mods than the one you are writing.

In case these clauses were being used to control the patch orering, they are now changed to `:BEFORE[RealPlume]`, which should be (mostly) equivalent except for not telling MM that RO is RealPlume.

Cheers!

### DIFF
--- a/GameData/RealismOverhaul/RealPlume_Configs/AIES/ROKestrel.cfg
+++ b/GameData/RealismOverhaul/RealPlume_Configs/AIES/ROKestrel.cfg
@@ -2,7 +2,7 @@
 //  Kestrel plume configuration.
 //  ==================================================
 
-@PART[RO-Kestrel]:FOR[RealPlume]:NEEDS[SmokeScreen]
+@PART[RO-Kestrel]:BEFORE[RealPlume]:NEEDS[SmokeScreen]
 {
     PLUME
     {
--- a/GameData/RealismOverhaul/RealPlume_Configs/AIES/VR1vulcan.cfg
+++ b/GameData/RealismOverhaul/RealPlume_Configs/AIES/VR1vulcan.cfg
@@ -2,7 +2,7 @@
 //  RS-68 plume configuration.
 //  ==================================================
 
-@PART[VR1vulcan]:FOR[RealPlume]:NEEDS[SmokeScreen]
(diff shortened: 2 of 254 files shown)


DRVeyl:
We did one mass patch of this, but maybe that was only for RaiderNick's parts.  Then enforce it on any new submissions, but clearly never got around to replacing everything that exists.

BTW, the pattern we adopted elsewhere when we have spot-fixed this is `BEFORE` RealPlume rather than `AFTER`.

Since RealPlume depends on Smokescreen, the NEEDS is also superfluous [modulo a few arguments about the tiny patch time to configure it but never evaluate].

HebaruSan:
OK, I'll switch these to `:BEFORE`. There are 13 files already using `:AFTER`; should I update those as well?

DRVeyl:
Skimming 254 files to make sure they're all doing what I expect with the pass was not a pleasant experience.

Turns out they are, so that at least is good.
So is making a list of a couple bizarre things they're still doing that we don't do anymore [namely, plume configs are NOT where something should change the engine module name to ModuleEnginesRF, or worse, NOT MERF.]

As for AFTER, depends.  IF they are doing the same things these do [declare PLUME node, point some things at it] then yes, for consistency those can become BEFORE.

HebaruSan:
Heh, sorry for the hassle. There are a _lot_ of files in this mod. :grimacing: 

> As for AFTER, depends. IF they are doing the same things these do [declare PLUME node, point some things at it] then yes, for consistency those can become BEFORE.

As far as I can tell, that is the case. Added a new commit with just those files.

DRVeyl:
Please back out that last commit.  The plume template system is correct and meant to run in AFTER.

HebaruSan:
OK, it's gone.

PR #2626, opened 2022-04-05
Tune composite/advanced composite tank mass

Capkirk123:
Based on 
https://hydrogen-central.com/hypoint-zero-emission-hydrogen-flight-range-ultralight-liquid-hydrogen-fuel-tanks/ 

improve the mass ratio of advanced composite tanks to ~10 when loaded with LH2. (The example tank achieved a mass ratio of nearly 13.5, but it's just a tank. Adding mass of stage structure and other systems, and assuming improvements continue, a mass ratio of 10 seems fairly reasonable for near future tankage. The example tank, when completed as a dewar, has a much less impressive mass ratio, but still very good for a dewar. SM-V material?)

Also make composite tanks slightly better based on RL electron performance info.


DRVeyl:
To save me some sanity, I'd like to stall this until after <https://github.com/KSP-RO/RealismOverhaul/pull/2609>.  Merging that will definitely create a conflict here, tho.

DRVeyl:
2609 merged, so adjust this to the new structure and it can get merged.

Capkirk123:
Conflict resolved. Should now be compatible with the new RF configs

PR #2627, opened 2022-04-05
Update Cryo Engines Support

vevladdd:
A small update that brings beautiful models of the Raptors (finally lol)
New: 
- Raptor SL;
- Raptor SL (9x);
- Raptor Vacuum;
- RS-18 (methalox LMAE);
- BE-4;
- STBE-3;
- Waterfall support;
Doesn't have support (yet) for:
- BE-4U (Vulture);
- ESA Prometheus (Compsognatus);
- Avio M-10 (Buzzard);
- Landspace TQ11/12 (Harrier);
These ones require engine configs in RO.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/CryoEngines/RO_CryoEngines_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/CryoEngines/RO_CryoEngines_Engines.cfg
@@ -1,4 +1,18 @@
 //=============
+//	Common patch
+//=============
+
+@PART[cryoengine-deinonychus-1|cryoengine-eagle-1|cryoengine-stromboli-1|cryoengine-tharsis-1|cryoengine-pavonis-1|cryoengine-etna-1|cryoengine-ulysses-1|cryoengine-fuji-1|cryoengine-hecate-1|cryoengine-vesuvius-1|cryoengine-tyrannosaur-1|cryoengine-iguanodon-1|cryoengine-hawk-1|cryoengine-allosaur-1|cryoengine-erebus-1]:FOR[RealismOverhaul]
+{
+	%crashTolerance = 10
+	%maxTemp = 673.15
+	%skinMaxTemp = 773.15
+	%fuelCrossFeed = true
+	%breakingForce = 10000
+	%breakingTorque = 10000
+}
+
+//=============
 //	Vulcain 2
 //=============
 
@@ -7,12 +21,6 @@
 {
 	%RSSROConfig = True
 	%rescaleFactor = 2
-	
-	%maxTemp = 1700
-	%crashTolerance = 12
-	%breakingForce = 250
-	%breakingTorque = 250
-	
 	%engineType = Vulcain
 }
 
@@ -36,14 +44,7 @@
 @PART[cryoengine-hecate-1]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
-
 	%rescaleFactor = 2
-	
-	%maxTemp = 1700
-	%crashTolerance = 12
-	%breakingForce = 250
-	%breakingTorque = 250
-
 	%engineType = RL10
 }
 
@@ -74,13 +75,7 @@
 {
 	%RSSROConfig = True
 	// No rescale needed, model nozzle exit diameter matches LE-9 spec
-
 	%mass = 1.72
-	%maxTemp = 1700
-	%crashTolerance = 12
-	%breakingForce = 250
-	%breakingTorque = 250
-
 	%engineType = LE7
 }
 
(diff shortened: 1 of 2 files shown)


GameData/RealismOverhaul/Waterfall_Configs/CryoEngines/CryoEnginesWF.cfg
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	//just rescale the plumes
+	@MODULE[ModuleWaterfallFX]
+	{
+		@TEMPLATE
+		{
+			@scale[*] *= #$/rescaleFactor$
+		}
+	}
```
DRVeyl, line 22: This gets to be *a lot* of repetition in the file.  Consider instead just one generic patch for the parts that use it?
```
@PART[a,b,c,d,e]:BEFORE[ROWaterfall]
{
	@MODULE[ModuleWaterfallFX] { @TEMPLATE { @scale[*] *= #$/rescaleFactor$ } }
}
```

GameData/RealismOverhaul/RO_SuggestedMods/CryoEngines/RO_CryoEngines_Engines.cfg
```diff
@@ -217,4 +217,160 @@
...
+
+@PART[cryoengine-deinonychus-1]:FOR[RealismOverhaul]
+{
+	%RSSROConfig = True
+	@rescaleFactor = 1.557
+	
+	%maxTemp = 1700
+	%crashTolerance = 12
+	%breakingForce = 250
+	%breakingTorque = 250
```
DRVeyl, line 234: Please break these out into a common patch:
```
@PART[a,b,c,d,e]:FOR[RealismOverhaul]
{
	%maxTemp = 1700
	%crashTolerance = 12
	%breakingForce = 250
	%breakingTorque = 250
}
```
Also suggest using the defaults configured in ROE [and which perhaps we should make the default settings within RO?]  <https://github.com/KSP-RO/ROEngines/blob/master/GameData/ROEngines/ROEEngineDefaults.cfg>
```
	%crashTolerance = 10
	%maxTemp = 673.15
	%skinMaxTemp = 773.15
	%fuelCrossFeed = true
	%breakingForce = 10000
	%breakingTorque = 10000
```

DRVeyl (changes requested):

GameData/RealismOverhaul/Waterfall_Configs/CryoEngines/CryoEnginesWF.cfg
vevladdd: I have changed that. I hope it's correct now)

GameData/RealismOverhaul/RO_SuggestedMods/CryoEngines/RO_CryoEngines_Engines.cfg
vevladdd: Same with this.

DRVeyl (approved):
Looks good to me.  @Capkirk123 ?

StonesmileGit:
This might need a change in ROE to not remove the engines here that are present in both mods

Capkirk123:
The engines work fine. However, the attach nodes are incorrectly positioned on many of them.

DRVeyl:
> This might need a change in ROE to not remove the engines here that are present in both mods

I think that's the role of the patch mananger.  I don't think we want to get into "RO-supported non-stock engine gets special patch treatment internal to ROE."  Or if we do, we need a much more holistic approach than supporting just-this-mod, plus whatever custom bit ended up making it for ROC's engines.

> The engines work fine. However, the attach nodes are incorrectly positioned on many of them.

Ugh.  Is CryoEngines maintained?  Since this is only touching rescaling, this should be fixed in the original mod.  (Are they doing something weird with MODEL.scale or their own rescaleFactor != 1?  Or the stock rescaleFactor that is != 1?)

Capkirk123:
The engines have switchable mounts with B9PS. If B9PS tries to move the attach node location, then that will cause issues when the part is rescaled.

DRVeyl:
Is there a way to PR a fix into B9PS to correctly deal with rescaleFactor?
I don't -think- that would get rejected because of everything else trying to use it, since I've seen this particular issue has been reported a few times.
The uglier way is we end up patching the node values in the B9PS info... which will then break when this eventually does get fixed at B9PS?

Capkirk123:
> Is there a way to PR a fix into B9PS to correctly deal with rescaleFactor?
> I don't -think- that would get rejected because of everything else trying to use it, since I've seen this particular issue has been reported a few times.
> The uglier way is we end up patching the node values in the B9PS info... which will then break when this eventually does get fixed at B9PS?

Fixing B9PS would certainly be the proper way to do it. I can’t say I’m confident in my ability to fix it myself, but I could certainly open an issue.

al2me6:
Do note that there are already patches that assume the erroneous B9PS behavior. Notably, https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg and a good number of the engines that I've configured.

Issue #197, opened 2015-03-27
Reaction wheels have incorrect stats


NathanKell:
@pjf did calculations, and they're way off my original guesstimate values. See https://github.com/KSP-RO/RP-0/issues/103 for details.

stratochief66:
I believe these have all been removed, with maybe 1 or 2 being worked on by winged7 to have more realistic stats and prices, which effectively makes them only useful in very specialized circumstances for RO anyway.

@NathanKell Close?

NathanKell:
Ah, my understanding is that they were not yet fixed and are still somewhat OP.

stratochief66:
Next To-Do:

Check if there are in fact any reaction wheels in RO or RP-0.
If found, they should be very weak, and also fairly late in the tech-tree.

Report work or updates back here please.

Capkirk123:
Is this resolved? Reaction wheels have been heavily nerfed and are fairly power hungry now.

DRVeyl:
Simple counts:
`ModuleReactionWheel`: 493 instances in the RO repo excluding craft files.
`!MODULE[ModuleReactionWheel]`: 392 instances, so we get rid of a lot of them by hand.
`@MODULE[ModuleReactionWheel]` 91 results.  So still a lot to go around.
`%MODULE[ModuleReactionWheel]` 6 results.
Not sure what the other 4 results are, but doesn't matter.

So I doubt this is resolved, without inspecting those 97 cases of configured reaction wheels.
If we still care, maybe a different patching strategy would be good.  Say, mark modules that we've configured and then blanket delete anything not marked in LAST[RealismOverhaul].  At least then there'd only be 100 patches instead of 500, plus who knows how many that are not configured by RO but sneak in from the source mod's stock config.

siimav:
Looks like DRVeyl implemented a config system that ensures all reaction wheels have consistent values.

PR #2629, opened 2022-04-09
Reaction wheel config system

DRVeyl:


DRVeyl:
Addresses <https://github.com/KSP-RO/RealismOverhaul/issues/197>
Note there is nothing in here regarding part mass

PR #2630, opened 2022-04-11
Update Top-Level RO Configs

DRVeyl:
See individual commits, recommend merge rather than squash for tracking

Notable is [Aggressively Set Tolerane, GForce, Pressure, Breaking Force, Max Temps](https://github.com/KSP-RO/RealismOverhaul/commit/8711beb425a642f5a2ffc2565403e9531e41f18f).  I didn't get any comments in the dev channel on why we had the strategy we did [where we appear to have been setting a limit on the original/"stock" part config rather than enforcing a default] and the commit history & blame don't seem to align with the actual changes, which I now believe was an MM error rather than intentional.  So I have removed the HAS checks and am enforcing defaults in BEFORE[RealismOverhaul].  Anything that wants to escape the default should do so properly here in a FOR[RealismOverhaul] patch or else later.


NathanKell:
SGTM.

PR #2632, opened 2022-04-11
Add MR-1 Meteo engines

Capkirk123:
Add the U-1250 and R-103, sustainer and booster for the Soviet MR-1 Meteo. This was a sounding rocket, roughly equivalent to the American Aerobee.

Engine configs mostly written by @scorpuu, I just did some more research to find the manufacturer's designations.

Co-Authored-By: Scorpuu <29466395+Scorpuu@users.noreply.github.com>

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
@@ -36,7 +36,9 @@
 	%title = R-103
 	%manufacturer = Kostin Bureau
 	%description = Post-war Russian version of German Taifun anti-aircraft barrage rocket. Developed and tested in 1947-1951 but abandoned in favor of the R-110. A cluster of 8 of these rockets formed the first stage of the MR-1 sounding rocket. Length without payload - 1.81 meters. 
-	
+
+	%specLevel = operational
+
 	@MODULE[ModuleEngines*]
 	{
 		%EngineType = SolidBooster
@@ -97,10 +99,6 @@
 			}
 			
 			curveResource = NGNC
-			
-			chamberNominalTemp	= 1500
-			maxEngineTemp = 2040
-			
 			thrustCurve
 			{
 				key = 1.00 0.90
--- a/GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
@@ -1,0 +1,155 @@
+//	==================================================
+//	R-103
+//
+//	Manufacturer: Kostin Bureau
+//
+//	=================================================================================
+//	R-103 (2.07 x 0.1 m)
+//	MR-1 Booster (cluster of 8)
+//
+//	Dry Mass: 8 kg					//literally a steel tube filled with solid fuel (24 kg full - 1 kg warhead - 11.6 kg fuel)
+//	Dimensions: 1.81 x 0.1 m 		(without warhead, tank lenght 1.44m - 69.5% of total 2.07 m usable)
+//	Thrust (SL): 10.175 kN
+//	Thrust (Vac): 11.192 kN
+//	ISP: 200 SL / 220 Vac
+//	Burn Time: 2.92 s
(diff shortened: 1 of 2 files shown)


GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
```diff
@@ -0,0 +1,155 @@
...
+			name = NGNC
+			amount = 6.41
+			maxAmount = 6.41
+		}
+	}
+	
+	MODULE
+	{
+		name = ModuleEngineConfigs
+		type = ModuleEngines
```
DRVeyl, line 70: Is it just "ModuleEngines" in the template or should this be ModuleEnginesRF?
Note there's a patch in RO's global config that will change all ME to MERF and all references in the MEC to MERF.  So the end result is MERF, and this question is more about consistency.

GameData/RealismOverhaul/Engine_Configs/U1250_Config.cfg
```diff
@@ -0,0 +1,286 @@
...
+				name = Kerosene
+				ratio = 0.32714
+				DrawGauge = True
+			}
+
+			PROPELLANT
+			{
+				ratio = 0.67286
+				name = AK20
+			}
```
DRVeyl, line 113: This isn't wrong, but it's weird ordering to *not* put the name key first (repeated through the configs)

GameData/RealismOverhaul/Engine_Configs/U1250_Config.cfg
```diff
@@ -0,0 +1,286 @@
...
+			pressureFed = True
+		}
+	}
+
+	RESOURCE
+	{
+		name = Tonka250
+		amount = 1
+		maxAmount = 1
+	}
```
DRVeyl, line 239: Another question out of curiosity, why do we apply an ignition resource *and* set ignitions to 1?  This seems redundant, is there a feature we get from this setup?  (I can see limited ignitor resource and large ignitions count.)

DRVeyl (commented):
Looks legal, so it can be merged as-is, but would like some info re: the comments.

GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
Capkirk123: Every single engine config is set up this way. It's probably meant to be that way.

GameData/RealismOverhaul/Engine_Configs/R103_Config.cfg
```diff
@@ -0,0 +1,155 @@
...
+			atmosphereCurve
+			{
+				key = 0 220	 // Guess
+				key = 1 200
+			}
+			
+			curveResource = NGNC
+			
+			chamberNominalTemp	= 1500
+			maxEngineTemp = 2040
```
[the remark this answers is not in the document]
Capkirk123: That sets the resource the thrust curve follows. I think it needs to be defined for the thrust curve to work even with one propellant.

GameData/RealismOverhaul/Engine_Configs/U1250_Config.cfg
Capkirk123: The ignition resource is set because the documents said that's what it used to ignite. It doesn't do anything else, since the ignition limit in MEC is set to one.

GameData/RealismOverhaul/Engine_Configs/U1250_Config.cfg
```diff
@@ -0,0 +1,286 @@
...
+
+			ullage = True
+			ignitions = 1
+			pressureFed = True
+		}
+		CONFIG
+		{
+			name = U-2000
+			description = Sustainer engine for the R-112 SAM. Tested, but never entered service.
+			specLevel = operational
```
DRVeyl, line 189: Last thing, I think.
Is this "operational" or "prototype" based on the comment?

Capkirk123: They appear to have launched a few of them off, so operational, if only briefly.

PR #2644, opened 2022-04-30
Add missing Source

Capkirk123:
Add missing source to U-1250, before I lose it again.


HebaruSan:
Sorry about the bogus config validation error, should be fixed now (see KSP-CKAN/KSPMMCfgParser#7). I will be monitoring the KSP-RO repos for a while, but if you see this sort of thing again, bug reports would be very welcome:

- https://github.com/KSP-CKAN/KSPMMCfgParser/issues/new

PR #2646, opened 2022-05-01
Add KTDU-5A, KRD-61

Capkirk123:
Add engine configs for the KTDU-5A and KRD-61. Also modify stock parts to create a KTDU-5A part, and add waterfall configs for the stock KDU-414, KTDU-5A, and KTDU-417.


NathanKell:
@Capkirk123 this conflicts now...

PR #2647, opened 2022-05-05
Fix non-restock support for new stock rcs blocks

lpgagnon:
- Without stock parts, RCSblock_01_small uses the same model
  as RCSBlock_v2; that means it should use the same rescale
  factor. Current rescaleFactor was only valid for restock's model

- Likewise the RCSLinearSmall rescaleFactor was good for restock
  but too big for stock. Eyeballed this one; the factor from
  the old Linear Thrusters isn't right either

- Give the _small parts a different title than the _v2 parts,
  because 2 different parts with the same title is confusing
  (better names welcome)

- And finally *hide* the RCSblock_01_small parts when restock
  isn't installed, because they end up exactly identical to the
  _v2 parts, making them completely redundant. We still want to
  create them (and scale them right) to allow creating a vessel
  with restock and loading it without restock.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_RCSnew.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_RCSnew.cfg
@@ -556,17 +556,19 @@
     %RSSROConfig = True
     @PhysicsSignificance = 0
     !MODULE[TweakScale] {}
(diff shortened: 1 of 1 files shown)


NathanKell:
Sorry, merged in wrong order I guess? @lpgagnon can you resolve?

lpgagnon:
with the other one merged, this one can just go away

PR #2649, opened 2022-05-06
Update RN Soyuz, Salyut and TKS configs

Capkirk123:
Update RN Soyuz, Salyut and TKS configs. Many changes, including:

- Add RealAntenna support patches
- Remove stock fuel cells/life support modules when Kerbalism is present
- Remove crew requirement from command modules and stations. Soviet spacecraft were extensively automated, and should be able to perform basic maneuvers with no crew present.
- Update Soyuz DPO/KTDU thruster performance according to "Soyuz: A Universal Spacecraft" and http://lpre.de/kbhm/index.htm. This includes dropping HTP RCS ISP numbers to sane values, dropping RCS thruster power to 10 kgf for all Soyuz 7K variants, adding HTP consumption to the KTDU-35 to drive turbopumps, and switching Soyuz-T RCS to 13.5 kgf biprop thrusters.
- Double power output of Salyut, Almaz and TKS solar panels. The given values seem to have been for average power output, resulting in stations being severely underpowered, and solar panels with unusually low power outputs for their mass and size compared to contemporaries

Resolves https://github.com/KSP-RO/SalyutStations/issues/8#issuecomment-1094538364

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_LK_LOK.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_LK_LOK.cfg
@@ -40,11 +40,11 @@
 	{
 	}
 	
-    !MODULE[ModuleCommand] {}
+	!MODULE[ModuleCommand] {}
 	MODULE
 	{
 		name = ModuleCommand
-		minimumCrew = 1
+		minimumCrew = 0
 		
 		RESOURCE
 		{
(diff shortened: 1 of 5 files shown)


raidernick:
@Capkirk123 I noticed this was merged with only the Soyuz-T craft updated. Did this change not affect RO RN Progress 7K-TG.craft, RO RN Soyuz 7K-OK-A.craft, RO RN Soyuz 7K-OK-P.craft, RO RN Soyuz 7K-OKS.craft, RO RN Soyuz 7K-T.craft, RO RN Soyuz 7K-T-AF.craft, RO RN Soyuz 7K-TM.craft and RO RN Zond 7K-L1.craft which all either use the same service module or some variation of it?

Capkirk123:
I assumed only the Soyuz-T used the Soyuz-T service module. The rest should use the Soyuz 7K-T service module, which was unaffected.

raidernick:
@Capkirk123 these parts were some of the first things I ever did and I took shortcuts and was lazy when doing them so things that should not have shared parts sometimes have shared parts. Doing a quick search inside all of the files I have it appears the RO RN Soyuz 7K-TM.craft and RO RN Soyuz 7K-OKS.craft also use the same service module because it looks the same visually and I probably saw no reason to make an exact duplicate with speculative configs anyway at the time.

This can probably easily be fixed by simply duplicating the service module again using +PART config and naming it t.pao3 instead of t.pao2 like it is for the Soyuz-T. Then the craft file just needs to be edited to replace the 2 with a 3 to fix it without redoing anything.

PR #2650, opened 2022-05-08
Larch prototype configs for Gamma

DeltaV1ms:
Adds prototype Larch configs to all the Gamma engines and corrects stats on Gamma-201/301


Capkirk123 (approved):
Everything seems reasonable here (and the performance is so low it probably won't significantly change balance anyway).

PR #2673, opened 2022-06-16
Fix SXTAntenna

Capkirk123:
RO adds a second ModuleDataTransmitter to the PS-1 antenna instead of editing the existing one. This is not only messy, but also breaks RA compatibility. Change patches to edit existing antenna module instead.


Capkirk123:
Those are probably also broken. I can look more closely at them.

PR #2677, opened 2022-06-22
Updated rescale factors for non-generic MLP launch pads.

hallucinogender:
Patched unique rescale factors for non-generic MLP launch pads. Every MLP part associated with a specific vehicle (with the current exception of the N1 pad) now has a unique rescale factor that should be correct for the real vehicle scale, unless I have erroneously missed something. Apologies if I have made some error with submitting this pull request, as this is my first time doing so.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/ModularLaunchPads/MLP.cfg
@@ -1,7 +1,7 @@
 @PART[AM_MLP*|MountPlatform_MLP]:FOR[RealismOverhaul]
 {
-    %rescaleFactor = 1.61
-    %RSSROConfig = True
+	%rescaleFactor = 1.61
+	%RSSROConfig = True
 	@MODULE[ModuleB9PartSwitch],*
 	{
 		@SUBTYPE,*
@@ -14,12 +14,7 @@
 			}
 		}
 	}
-	!MODULE[ModuleGenerator] {}
-	
-	@MODULE[ModuleAnimatedDecoupler],*
-    {
-        @name = ModuleStagedAnimation
-    }
+	!MODULE[ModuleGenerator] {}	
 }
 @PART[AM_MLP*|MountPlatform_MLP]:HAS[@MODULE[LaunchClamp]]:FOR[RealismOverhaul]
 {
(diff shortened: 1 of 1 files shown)


hallucinogender:
N1 pad now has a custom rescale factor. Every part in MLP that is associated with a real rocket launch pad is now properly rescaled to fit that rocket as closely as possible. Everything else is covered by the general patch for the mod, which I have left unchanged.

NathanKell:
This may have been true of the generic patch which you copied from as well, but I notice there's an alignment issue here (and in all the other patches): https://github.com/KSP-RO/RealismOverhaul/pull/2677/files#diff-ff02558e17296e89902a4f39f3a5db16fa4a32e8fceaeb816451c57353b7048dR38 the rescaleFactor and the MODULE patches are offset differently. Is one using spaces and the other tabs?

Also the switch from AnimatedDecoupler to StagedAnimation is no longer needed since MLP does that itself now. Can also be fixed in the base patch.

Other than that--thank you for doing this, it's really neat!

hallucinogender:
I'll go through and fix the alignment issue; it's inherited from the generic patch, since I didn't realize there was a need to check that before copying over the original generic patch for the dedicated rescale factors. I'll also remove the obsolete module switch. Thanks for letting me know about that.

hallucinogender:
Obsolete code is gone. Fairly sure the spacing errors should all be fixed now too.

arrowmaster:
This PR has caused the nodes on the bases to be in the wrong location. I've personally checked the SaturnV base and I believe someone else said the Shuttle base was also wrong

AlphaMensae:
I see what's going on now, all the nodes are being "fixed" by dividing them by the rescale factor. However, the B9PS node error only happens when B9PS *moves* the nodes in a height or length subtype. The prime examples of this are the modular general service tower sections, which have multiple height variants. As each height variant is selected, B9PS moves the top node to match the new height, except it is doing it by the rescaleFactor squared. In this case dividing the node positions by the rescaleFactor would correct the error.  But for nodes that are now moved, they are being moved *down* instead, which for nodes on the deck of a launch base will make them move down below the deck and thus become invisible.

Most of the historic launch stands do not have tower nodes on the deck or ring itself, so there won't be anything amiss to notice.  An exception is the Redstone Launch Stand, there are size 0 nodes for additional umbilical poles on the ring. The Saturn V and Shuttle launch bases do have nodes on the deck (mainly the tower node for the Saturn V base), so they get moved down below the deck.

All the corrections for stationary nodes need to be removed from this patch.

PR #2685, opened 2022-07-11
Fix XLR35 testedBurnTime

lpgagnon:
2 testedBurnTime values provided, first one wins, second one is "correct"

Confirmed fix in MM cache: a key=3600 becomes key=900


NathanKell:
Thanks!

NathanKell:
lol thought I'd merged this already

PR #2689, opened 2022-07-15
Updating RD-119 config according to new data

VeselyiKot:
According to new data source, RD-119 should have 362 s isp and 100,5 kn of thrust (instead of current 352 and 105,5), as well as 5% throttle range.
Taking into account the bibliographic list and the authority of the paper ('Н.К. МАТВЕЕВ, А.А. СЕМЁНОВ, УСТРОЙСТВО ДВИГАТЕЛЯ РД-119'), these data can be considered more reliable than the available ones (current config contains links to aggregators of lower reliability and foreign sources).


Capkirk123:
This source appears to be accurate, these changes are good to merge

PR #2690, opened 2022-07-16
Increase RD-0122 testedBurnTime to 6000s

VeselyiKot:
This engine was planned to be modified to ensure its tenfold reuse (source: B.I.Gubanov, "THE TRIUMPH AND TRAGEDY OF ENERGIA", chapter 41).
If it had been created, there is every reason to believe that these target parameters would have been achieved.


Capkirk123:
This is a little dubious, considering the issues the SSME had achieving it's advertised burn time.

VeselyiKot:
Source states this: "Block A uses engines with a thrust of up to 850 tons in the void, running on standard fuel components - liquid oxygen, hydrocarbon fuel, in the second stage - engines with a thrust of 230 tons in the void, running on fuel, the components of which are liquid oxygen and liquid hydrogen. The engines were borrowed from the Energia launch vehicle, they are subject to modifications in terms of ensuring their reusable use and some increase in characteristics. At the same time, it was assumed that the modernization of these engines was supposed to create a reserve in increasing the mass of the payload. It was supposed to start the project with the existing engines without modification to accelerate development."

So, given that 0122 is the intended engine for Energia II *upgraded* configuration, it should get reusability capability, given that it was one of such modernization main tasks. Especially given that 0122 was intended only for tenfold reuse, not 55 times like SSME.

PR #2692, opened 2022-07-24
Add some early tech transfer to RD-100, RD-200, and RD-107/108

arrowmaster:
Submitting this for review to get more eyes on it and input. DO NOT MERGE YET.

RD-100 series: Keep 50% transfer from previous config, add 25% transfer from other earlier configs
RD-200: Change 25% transfer from the RD-100 to the RD-102 simply because ECMs have the RD-200 as a direct descendant of the RD-102. Add 10% transfer from pre RD-102 configs.
RD-107/108: Add 10% transfer from the RD-102,RD-103,RD-103M. I did not add transfer from the 100 or 101 as I modeled this partially on how the US line does DU transfer. This decision was more for game balance than historical accuracy.

The value of 10% was just arbitrarily chosen when I made the changes as a value less than 25%.

The RD-211 and beyond all have ECM links to the RD-200 but haven't been changed yet.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD100_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD100_Config.cfg
@@ -425,7 +425,7 @@
 		ignitionDynPresFailMultiplier = 2.0
 		cycleReliabilityStart = 0.8
 		cycleReliabilityEnd = 0.92
-		techTransfer = RD-101:50
+		techTransfer = RD-100:25&RD-101:50
 		reliabilityDataRateMultiplier = 1
 	}
 	@MODULE[ModuleEngineConfigs] { @CONFIG[RD-102] { %ratedBurnTime = #$/TESTFLIGHT[RD-102]/ratedBurnTime$ } }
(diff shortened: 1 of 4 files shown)


arrowmaster:
With the 10% value this is probably still weak compared to US engines

leudaimon:
I'm not sure your syntax there is correct, using &... but I haven't looked at these in quite a while so I'll leave it to Kirk or someone else to review that. 
That said, I would treat the RD-102 as a side-grade instead of a step between RD-101 and RD-103. Thus, following your logic, the RD-103 gets the same du from the RD-101 and RD-102 (and less from the RD-100), and the RD-107/8 get also a bit from the RD-101.

Capkirk123:
Yes, the formatting is wrong. The most recent engine should be first on the list, with the most distant engine last.

The RD-102 is mostly a side-grade. I would say the ancestor the the RD-107/108 is the RD-103 (turbopump) and RD-200 (combustion chamber).

arrowmaster:
I think I have fixed the formatting to have the engines in the correct order now. I've also changed the RD-101 and RD-102 to have the same worth to the RD-103 and RD-200. The RD-103 and RD-200 have the same worth to the RD-107/108.

This creates a point where the RD-102 progression diverges into two branches the RD-103 and RD-200 which then merge back together for the RD-107/RD-108 matching Capkirk's comments. The RD-103M is left as a deadend which makes sense given its unlocked in the same tech tree node as the RD-107/108. This branch and merge could also be applied to ECMs to give the RD-103 more value.

Capkirk123:
That all seems correct. Looks good to me!

PR #2696, opened 2022-08-10
Revise engine burn time and add throttle curves

Capkirk123:
Revise engine overburn limits with better sources and more consistency. Also add throttle curves for engines with throttling capability.


NathanKell:
I still think it's worth setting *rated* burn time for XLR11 and XLR25 to the aircrafts' burn times (as we do for rated burn time wrt stage burn time), and use tested burn time for the static fires?

Capkirk123:
4-5 minutes is the quoted burn time on an X-1E flight, and on the final and fastest X-2 flight, a 160 second burn was performed. As far as I can tell, 5 minutes and 175 seconds is a fairly accurate guess for the average burn time of the X-1 and X-2.

NathanKell:
Was it throttled down on the X-1E? 4-5min sounds like an *awful* high prop load for that a/c.

NathanKell:
Nope, does look like the X-1E with ~3.5t load made about 4-5min. Dang.

Capkirk123:
> 

Yup, just ran the math myself.

PR #2704, opened 2022-08-19
General RO Configs

SovietOnion1608:
A few RO configs for other mods, ones that don't really fall under a very specific category or too small to have their own branch. Work probably needed on NS pod IVA. Values for both New Shepard and James Webb were a bit hard to come by, so a few guesses were made.


GameData/RealismOverhaul/RO_SuggestedMods/NewShepard/NS-ROConfig.cfg
```diff
@@ -0,0 +1,209 @@
...
+	!MODULE[ModuleDataTransmitter],* {}
+	MODULE
+	{
+		name = ModuleRealAntenna
+		referenceGain = 1.5 //FIXME guess
+		RFBand = UHF
+	}
+}
+
+@PART[NewShepard_Pod]:NEEDS[FeatureHabitat]:AFTER[zzzKerbalism]
```
StonesmileGit, line 153: The bottom two blocks here should probably move to ROKerbalism

PR #2705, opened 2022-08-19
AWSI RO Configs

SovietOnion1608:
Configs for EStreet's Kistler K-1 mod. Fix necessary for antennae and Nitrogen amount stored in tanks.


GameData/RealismOverhaul/RO_SuggestedMods/Asbury/RO_Config_Engines.cfg
```diff
@@ -0,0 +1,341 @@
+@PART[asbury_aj-26-58]:FOR[RealismOverhaul]
```
StonesmileGit, line 1: is there a reason for not placing engine configs in the [Engine_Configs](https://github.com/KSP-RO/RealismOverhaul/tree/master/GameData/RealismOverhaul/Engine_Configs) folder and using the common patching system?

PR #2707, opened 2022-08-19
BDB Delta RO Configs

SovietOnion1608:
Mostly correct, I think. Might need some work regarding node positions though.


raidernick:
This PR is superfluous now due to this one which has been merged: https://github.com/KSP-RO/RealismOverhaul/pull/3155

PR #2709, opened 2022-08-20
BDB Titan RO Configs

SovietOnion1608:
Has most variants, but not all. A few fixes might be necessary at some places.


raidernick:
This PR is superfluous now due to this one which has been merged: https://github.com/KSP-RO/RealismOverhaul/pull/3155

PR #2714, opened 2022-08-27
Add optional restart kits to Agena & J2

Capkirk123:
Add restart kits to Agena and J2. 
Agena has optional start tanks as developed for GATV, which could be installed on Agena at the customer's request.

J-2 itself supported restarting in all configurations, but engine ullage, conditioning, and pressurization systems were only installed on SIVB-500s.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Agena_XLR81_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Agena_XLR81_Config.cfg
@@ -377,30 +377,30 @@
 
 		CONFIG
 		{
-			name			= XLR81-BA-11
-			description		= Model 8096, Agena D. Cannot restart between 15 minutes and 3 hours after shutdown.
-			specLevel		= operational
-			maxThrust		= 71.17
-			minThrust		= 71.17
-			heatProduction	= 100
-			ullage			= False //(1) p 1-4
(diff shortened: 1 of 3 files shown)


NathanKell:
Is the Agena restart kit replacing XLR81-BA-13 ?

GameData/RealismOverhaul/Engine_Configs/Agena_XLR81_Config.cfg
```diff
@@ -377,8 +384,8 @@
 			ullage			= False //(1) p 1-4
```
al2me6, line 384: Let's avoid `tab`-alignment IMHO. Clutters the diff and can be wrong depending on tab size.

Capkirk123:
> Is the Agena restart kit replacing XLR81-BA-13 ?

Yes, BA-11 and BA-13 were identical other than the installation of the start tanks.

NathanKell:
Now that KSPCF 1.22.1 is released (and I updated the upgrade pipeline for RO) we can setup an upgrade to deal with these changes and then the PR can be merged. That goes in UpgradeData.cfg  - holler if the comments don't make sense! (You'll want to make a new subnode and target 14.18.0, then put the old and new configs in there per comment).

NathanKell:
I made the change.

PR #2715, opened 2022-09-02
Add RL10A-3-7

Capkirk123:
Add RL10A-3-7, the actual throttlable RL10 lunar lander engine developed by P&W. 

Replace the now redundant speculative Lunex config? The speculative Lunex config is based off of era documentation I found of throttling RL10 tests, the culmination of which was the RL10A-3-7 presumably. Since both engines fill the same role and have mostly the same performance, the imaginary config should be removed in favor of the actually built prototype. 

Unfortunately, removing engine configs tends to be save-breaking. Just deprecate the config for now? Marking as draft for awaiting further discussion


Capkirk123:
It's probably fine. Lunex config can be removed at a later date.

PR #2721, opened 2022-09-19
Squad SepMotor Mesh fixes

vader111:


GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
```diff
@@ -425,6 +425,7 @@
 }
 
 // Use RO-provided model instead of stock or Restock
+@PART[sepMotor1]:NEEDS[ReStock]:FIRST { %RestockIgnore = True }
```
StonesmileGit, line 428: This line seems to prevent the `ROSmallSpinMotor` from loading

GameData/RealismOverhaul/RealPlume_Configs/Squad/sepMotor1.cfg
```diff
@@ -1,4 +1,4 @@
-@PART[sepMotor1]:BEFORE[RealPlume]
+@PART[sepMotor1|sepMotorSmall|sepMotorLarge|ROSmallSpinMotor]:BEFORE[RealPlume]
```
StonesmileGit, line 1: The plume size is the same on all of these parts with this patch. The plume should be scaled to match the scale of the specific part

StonesmileGit (changes requested):

PR #2724, opened 2022-10-02
Better temp settings for Aviation Cockpits fighters

lpgagnon:
"skinMaxTemp = 900 // mach 3" is nonsense; sustained mach 3 flight at 18km on the Century Fighter cockpit only goes up to 560K skin temp, but going up to mach 3 for even 1 second is enough to bring Int temp over 400K before the skin can cool down

- Bring skinMaxTemp down to 600K, to avoid creating false expectations
- Reduce skin->int soak to allow spending at least a bit of time (tested 3 minutes) at mach 3 without melting down

400K is enough to stay at ~mach 2.2 indefinitely; seems adequate, so unchanged.

Mainly tested on Century Fighter cockpit, but applied to all fighters for consistency.

Also here: apply oxygen fix to f-104, to match Century and Voodoo

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/AviationCockpits/Cockpits.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/AviationCockpits/Cockpits.cfg
@@ -248,7 +248,7 @@
 	}
 }
 
-// ******* Hopefully by this point you have life support unlocked! ****
+
 
 // Generic Century Fighter Cockpit
 @PART[Trainer?Cockpit]:FOR[RealismOverhaul]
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SuggestedMods/AviationCockpits/Cockpits.cfg
```diff
@@ -264,7 +264,9 @@
 	@breakingForce = 2000
 	@breakingTorque = 2000
 	%maxTemp = 400
-	%skinMaxTemp = 900 // Mach 3
+	%skinMaxTemp = 600
```
StonesmileGit, line 267: Have these numbers been tested in game to be equivalent to the old skin temperature?

lpgagnon: they've been tested in-game to the extent described in the commit message:

- allow spending at least a bit of time (tested 3 minutes) at mach 3 without melting down
- enough to stay at ~mach 2.2 indefinitely

PR #2728, opened 2022-10-09
Configure some firespitter cockpits

lpgagnon:
- biplane cockpit: light and crappy, but someone is bound to find it fun
- early bubble cockpit, vaguely F-86ish, balanced to sit somewhere between Meteor and Century cockpits


Capkirk123:
There appears to be no adapter sized for the FS fighter cockpit? Otherwise looks good.

lpgagnon:
There isn't. stock size, both of those cockpits have the same shape and size; since I made the fighter cockpit larger, it'd need a larger copy of the adapter - but it wouldn't adapt to anything useful. So coming up with smooth area-ruled transitions will be left as an exercise for the player.

Capkirk123 (approved):
A little silly, but why not. Looks good.
Needs https://github.com/KSP-RO/RP-0/pull/1808

PR #2730, opened 2022-10-26
Merge ROKerbalism into RO

StonesmileGit:
This is a draft for merging ROKerbalism into RO.

TODO:
- [ ] Remove unused files copied from original stock kerbalism config
- [ ] Remove `:FOR[Kerbalism]`. Can't have that in RO
- [ ] Add `:NEEDS[Kerbalism]` in places that need it
- [ ] Move mod specific configs to the same place as where the RO configs exist

And probably more.


StonesmileGit:
Turns out this is way more annoying to pull off than first thought.

PR #2734, opened 2022-11-14
Add support for ROT-CMES-Tank

lpgagnon:
with some refactoring to avoid yet another copy of the same 'generate one part per tank type' block

needs https://github.com/KSP-RO/ROTanks/pull/81


Capkirk123:
Once again, seems to work fine.

PR #2737, opened 2022-11-17
Disable extremely long restart windows

### LINKED ISSUE #2736: s5.92 causing full game freeze in VAB
Since the update, when the s5.92 (and also potentially the s5.98M) is selected and placed, the game begins to lag, and about 10 seconds later completely freezes with the game no longer responding.

Log is attached.
[message.txt](https://github.com/KSP-RO/RP-0/files/10020373/message.txt)

Capkirk123:
Due to a TF bug of unclear origin, extremely long (300 day) TF restart windows cause a game freeze when the part is moused over. Just disable them until a solution can be found.
Resolves #2736

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/S5_92_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/S5_92_Config.cfg
@@ -143,8 +143,9 @@
 				{
 					key = 0 0 0 0
 					key = 360 1 0 0		//6 minutes between restarts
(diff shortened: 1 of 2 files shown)


Capkirk123:
Based on conversation in https://github.com/KSP-RO/TestFlight/issues/256, it seems like this is an issue on the TF side, not the RO side. Since it's unclear if it can be resolved on the TF side, I'm just gonna merge this now, it can always be undone later.

PR #2739, opened 2022-11-18
Give LR101 tankfed mode

Capkirk123:
Convert LR101 to use bimodal engine, capable of switching from pumpfed to tankfed mode.

LR101-NA-3 in pumpfed mode requires ground clamps to start, as early LR101s used pyrotechnic starters. All others can airstart.

This should discourage LR101 upper stages, as the LR101 in tank-fed mode performs worse and requires substantial pressurization.


lpgagnon:
While you're at it, make sense to also make the s-3d "vernier" ground-lit-only? Its parent engine is. It's technically possible to use the vernier with some other, non-ground engine, but that seems a weird edge case to want to support

PR #2740, opened 2022-11-18
Update/fix S5.23 TestFlight config, typos

ilmcamam:
Seems to have got lost in the mix, leaving the engine without TF support in-game (even though others with the older style, like the M-10, still work?). Also fixes a couple of typos (GRAU index 11D49, not 11D19, and Kosmos-3 based on R-14, not R-16).


Capkirk123:
Whoops

PR #2743, opened 2022-11-21
Remove ullage from 11D58

VeselyiKot:
According to known data (added in links), Buran OMS was fitted with fuel scoop. Given that for late Agena it gives "no ullage", it would be reasonable to also grant 11D58 with "no ullage" status, at least until someone will develop proper scoop system for tanks themselves.


Capkirk123:
Could've sworn that was already disabled...

PR #2744, opened 2022-11-22
Unbreak Mk1-3 rcs

### LINKED ISSUE #2566: Mk1-3 Stock Pod has RCS Issues
The integrated RCS on the Mk1-3 pod does not work Real fuel says it uses 50%MMH and 50%NTO this does not account for pressurant therefore the rotational thrusters do not fire.

lpgagnon:
Restock model uses thrustTransform. RO used to have a restock-specific patch to use the right name, got lost in one of the vens-vs-restock refactors. BUT the Squad model also uses thrustTransform, and Ven's doesn't seem to have a Mk1-3 model, so this was just broken everywhere? Guessing the stock model changed at some point.

Fixes #2566


BobKermanIndustries:
When you try to load stock and enable to show errors an error shows the the stock audio has the wrong transform name RCSTransform so there is a bug with the transform name for Mk1-3 in stock as well.

PR #2747, opened 2022-12-01
RD-107/108 vernier (Making History Cub) fix

jimmymcgoochie:
A single RD-107/108 vernier weighs as much as the entire RD-108. I've changed it to 40kg which makes sense from a vague "half the mass difference between the two" standpoint so it can actually be used as intended and also fixed a typo in rescaleFactor in the same config.


StonesmileGit:
With the now applied rescale factor, does the size make sense in-game? (Has the size been tested in-game with this fix so it isn't tiny)

jimmymcgoochie:
> With the now applied rescale factor, does the size make sense in-game? (Has the size been tested in-game with this fix so it isn't tiny)

It was tiny. I've removed the rescale entirely which makes this part closer in size to the verniers on the ROE RD-107/8. It could probably be a bit longer still but that's a bit beyond my config-tinkering skills.

Issue #2748, opened 2022-12-03
RS-68 burn time quick-fix


Capkirk123:
The RS-68 has limited burn time due to being an ablative engine, but throttles down during flight in some configurations, resulting in a longer burn. The throttle-based burn time limits in TF aren't working, and are not yet implemented in TL, so I made a quick fix by adding overburn to the RS-68. 

This should be removed whenever throttle based burn time limits are working.

PR #2753, opened 2022-12-19
Fix thrustmodifiers, remove quick-fix

Capkirk123:
Add thrustmodifiers to a few engines I missed, and remove the RS-68 burn time quick-fix since thrustmodifiers have been fixed in TF (see https://github.com/KSP-RO/TestFlight/pull/252)

fixes https://github.com/KSP-RO/RealismOverhaul/issues/2748


Capkirk123:
Merging as https://github.com/KSP-RO/TestFlight/releases/tag/2.3.0.0 should fix thrustmodifier bugs

PR #2763, opened 2023-01-11
KerbalFoundries 2 RO config

xFluffyDemon:
Think its how this works, first time doing this.

so theres probably a little more balance needed on some parts, but everything is working as far as i can tell


siimav (changes requested):
* Not a fan of 25m/s crash tolerance. Looks like RO-configured rover wheels in general have 12.
* The mod is _Kerbal Foun**d**ries_, not _Kerbal Founries_. Also the folder naming convention appears to be to replace spaces with _.
* How did you decide what max speed to assign? I was about to complain about those numbers but the current large rover wheels might have rather unrealistic numbers as well.

xFluffyDemon:
> * Not a fan of 25m/s crash tolerance. Looks like RO-configured rover wheels in general have 12.
Things became very prone to explosions at the speeds these wheels could go, i guess i could try to mess with suspension but im still kinda figuring it out, increasing crash tolerance was a diry fix to that

> * The mod is _Kerbal Foun**d**ries_, not _Kerbal Founries_. Also the folder naming convention appears to be to replace spaces with _.
Well thats embarrasing

> * How did you decide what max speed to assign? I was about to complain about those numbers but the current large rover wheels might have rather unrealistic numbers as well.
Mas speed is the same as the original mod, could probably halve them

siimav:
What even changed between now and when I last reviewed this? Sorry, I really can't tell from those commits. I'm still seeing rather high crash tolerances and speed limits.

xFluffyDemon:
Changed impact from 25m/s to 18m/s, tried messing around with suspensions to get them to work as they should but its still out of my league, speed is the same as the big rover wheels to i don't see a issue there atm (can change them if you really want)

GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
```diff
@@ -0,0 +1,438 @@
...
+    }
+}
+
+@PART[KF-WheelTiny]:FOR[RealismOverhaul]
+{
+    %RSSROConfig = True
+    @mass = 0.03
+    @crashTolerance = 18
+
+    @MODULE[ModuleWheelBase] { @mass = 0.1 }
```
DRVeyl, line 57: mass values mismatch

GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
```diff
@@ -0,0 +1,438 @@
...
+
+        @RESOURCE[ElectricCharge] { @rate = 10 }
+
+        !torqueCurve,*{}
+        torqueCurve
+        {
+            key = 0.0  2.0 0.0 0.0
+            key = 7.8  1.5 0.0 0.0
+            key = 15.6 0.5 0.0 0.0
+            key = 23.5 0.0 0.0 0.0
```
DRVeyl, line 246: Last torque key doesn't match wheel speed max

GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
```diff
@@ -0,0 +1,438 @@
...
+        }
+    }
+}
+
+@PART[KF-ALG-Small]:FOR[RealismOverhaul]
+{
+    %RSSROConfig = True
+    @mass = 0.1
+    @maxTemp = 2000
+    %skinMaxTemp = 2000
```
DRVeyl, line 405: That's very high, and probably should just be subject to the normal RO limits.  Same for the rest of the parts here.

DRVeyl (changes requested):
Few things to fix.
Overall balance from siimav's comments, might be good enough to get this in and used and then make adjustments.  No idea how the power draw stats are derived, might want something consistent or some particular rule?
Structurally, since there's a ton of repeated stuff in here, I would probably have structured it a bit to make things easier to change across everything.  For instance, build the torque curve using MM patch/math based on the wheelSpeedMax.  It's actually very formulaic here: the in/out tangents are all 0, and the 4 keys are 0, 1/4, 1/2, and 1x respectively of wheelSpeedMax.  And the ModuleWheelBase mass is always the part mass, so both of those could be put into a single generic patch.

Not requiring that technique to say I'd approve it, but it might help from a maintenance standpoint.

GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
xFluffyDemon: Im pretty sure there was a reason i made them able to get hotter, but cant remember now, ill just default it

xFluffyDemon:
> No idea how the power draw stats are derived, might want something consistent or some particular rule?

Had a spreadsheet where i kinda derived the values, but ive since lost it because it had the default name and was overwritten at some point



>Structurally, since there's a ton of repeated stuff in here, I would probably have structured it a bit to make things easier to change across everything. For instance, build the torque curve using MM patch/math based on the wheelSpeedMax. It's actually very formulaic here: the in/out tangents are all 0, and the 4 keys are 0, 1/4, 1/2, and 1x respectively of wheelSpeedMax. And the ModuleWheelBase mass is always the part mass, so both of those could be put into a single generic patch.

i have no idea how to do that, major skill issue lol

siimav:
Took another look and I think the skids deserve a slightly higher max speed than 50m/s. For reference try to find what the X-15 landed at.
Additionally I think you should configure the airplane landing gear parts with lower limits than the default 450m/s. Those also appear to have motors which I'm honestly not a big fan of. Motorized plane landing gears are more of a 2010+ concept and even then the idea is to save fuel while taxiing.

siimav:
Btw, any idea why the speed limit shown there doesn't align at all with what is shown in the PAW?

image

Capkirk123:
Any progress on this?

siimav:
Merging this to a WIP branch.

PR #2765, opened 2023-01-20
Fix EVA'd kerbal mass being the stock 45kg instead of 100kg

siimav:


Capkirk123:
image
Yup, works as intended

PR #2766, opened 2023-01-22
Convert AJE to MEC, more configs, fix turboFanEngine

Capkirk123:
Convert AJE jets to use MEC and global engineconfigs, and add some more configs. Duplicated parts used to represent different engine versions have been deprecated.

Also remove the broken realplume config from the J58 and just use stock effects.

Needs https://github.com/KSP-RO/RP-0/pull/1874


siimav:
How breaking is this when it comes to existing craft files?

Capkirk123:
A lot of stuff will get set back to the starting config. Although, I could just set the default config to whatever the engine was configured as before so it stays the same.

Capkirk123:
Ok, switched the default configs for MEC to the engine version used by AJE, so engines will stay on the same config when things are updated.

This will force the player to reselect a config they have actually unlocked in the future (unless we made MEC automatically change configs when we improved RP-1 integration), but it shouldn't be a huge issue.

PR #2767, opened 2023-01-23
Add ejection seat functionality

siimav:
Only a handful of cockpits are currently configured. When the RO dll is built in debug mode then there are handy onscreen markers that help with nailing down collider rotations, offsets and force vectors.

Among other changes, this also ups the naut G tolerance from 20 to 25 since ejection seats aim for ~20G.

Configured parts:
* X-1
* X-15 standalone
* X-15 inline
* Mk1 standalone
* Mk1 inline
* Mk2/SR-71-alike standalone
* Mk2/SR-71-alike inline
* SXTBuzzard aka the Stuka

**#2765 should be merged before this.**

### DIFF
--- a/GameData/RealismOverhaul/EVA.cfg
+++ b/GameData/RealismOverhaul/EVA.cfg
@@ -28,7 +28,7 @@
 	%thermalMassModifier = 4
 	%emissiveConstant = 0.89
 	%absorptiveConstant = 0.25
-	%gTolerance = 20
+	%gTolerance = 25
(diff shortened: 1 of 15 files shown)


Capkirk123:
Tested, seems to work well. Ejection seat successfully ejects kerbals from aircraft in unusual attitudes at all subsonic speeds. The addition of an EVA pack to the kerbal increases their weight enough that ejection becomes much less reliable, maybe a warning should be added?

PR #2768, opened 2023-01-28
Mk2 lander has rover variant again, slight weight reduction

xFluffyDemon:
Slight weight reduction, tank size the same because "sidepods" are empty space


Capkirk123:
Looks fine for now, but changing the variant should effect fuel tank volume if possible.

xFluffyDemon:
I thought about it, but the sides are more like cargo bays rather than tanks, so idk if i should've changed it

Capkirk123:
I see, they are cargo pods. Well, looks good to me then.

PR #2770, opened 2023-02-02
Hide the bugged miniLandingLeg

Test-Account-KSP:
Hide the bugged miniLandingLeg from the parts list in VAB/SPH and from R&D building


arrowmaster:
There is already an RO_Deprecation.cfg that will do the hiding. From some of the other configs I think the right way to do with would be with the following.

`@title = #$/title$ (Deprecated)
%deprecated = True`

siimav:
Deprecating isn't really the correct term here since ideally we would only be temporarily hiding the parts. If Restock finally gets the long-awaited release then we would just revert the changes in this PR.

Capkirk123:
Is this good to merge then?

siimav:
Yes, good enough for a temporary (???!) workaround.

PR #2774, opened 2023-02-08
Basic AirplanesPlus configs and new RO cockpits

Capkirk123:
Add configs for a few Airplanes Plus parts, and add a new starting cockpit and supersonic cockpit to RO using Airplanes Plus models.

Changes include:
-Added two-seater "Jet Rrainer Cockpit" using the Airplanes Plus fighter cockpit
-Added P-40 Cockpit using Airplanes Plus old fighter cockpit
-Converted X-1 cockpit to use RF tanks, and patched X-1 cockpit temperature limits so they aren't overridden
-Configured Airplanes Plus Fighter Cockpit, Old Fighter Cockpit, Zero Cockpit, J85 and CF34.

Needs https://github.com/KSP-RO/RP-0/pull/1881
Needs https://github.com/KSP-RO/ROKerbalism/pull/139


GameData/RealismOverhaul/RO_SuggestedMods/AirplanePlus/Command.cfg
```diff
@@ -0,0 +1,118 @@
...
+		{
+			name = ElectricCharge
+			rate = 1.2
+		}
+	}
+
+	MODULE
+	{
+		name = ModuleFuelTanks
+		type = SM-IV
```
arrowmaster, line 95: Shouldn't this be SM-I for where these are being placed in the tech tree?
```suggestion
		type = SM-I
```

Capkirk123: F-15B first flew in 1972

arrowmaster: It didn't show up well as a comment. Line 35 is the F-15B, line 95 is under the P-40 and A6M config. This was for line 95.

arrowmaster:
Putting the EC into a ModularFuelTank is increasing part costs and built times. These numbers are gathered from a craft consisting of only the cockpit as root part with no kerbals, and then clicking Remove All Tanks. The biggest difference is with the cost difference between the Junker and P-40 because of the different methods.

| Cockpit     | Uncrewed | Empty tanks |
| ----------- | -------- | ----------- |
| Junker      | 37       | N/A         |
| P-40        | 125      | 39          |
| X-1         | 394      | 308         |
| Jet Trainer | 697      | 611         |

Capkirk123:
Battery size and power consumption has been adjusted, and the batteries on early cockpits should be free now.

arrowmaster:
I have tested again and there is minimal cost and build time difference between using the Junker and P-40.

PR #2779, opened 2023-02-10
Remove ModuleDataTrasmitter from MK1CrewCabin

arrowmaster:
This part is not a command module and not caught by RealAntennas for conversion. A crew cabin that is not a command module has little need for an antenna of its own.


Capkirk123:
You probably wanna gate that behind :NEEDS[RealAntennas]

arrowmaster:
Updated to only removed when RealAntennas or RemoteTech are installed. I removed RealAntennas from my dev install and loaded with stock comms, the part gets a 100Km antenna but so do many other command pods. Stock comms in RO/RSS looks to be a mess.

Capkirk123:
The behavior stock comms in RO isn't really maintained, but it's better to avoid any unnecessary changes.

PR #2788, opened 2023-02-22
Change Jet Engine MEC default to first non upgrade in RP-0

arrowmaster:
Most of these were set to configs that unlocked in later nodes than the config they have been changed to. For the F100 and J75, the original default config and the new default unlock in the same node, but the originals were flagged as upgrades and the new default is the first entry in the MEC GUI and not flagged as an upgrade.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/AL31_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/AL31_Config.cfg
@@ -111,7 +111,7 @@
 	{
 		name = ModuleEngineConfigs
 		type = ModuleEnginesAJEJet
-		configuration = AL-31FM
+		configuration = AL-31F
 		modded = false
 		origMass = 1.520
 
--- a/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/Atar09_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/Atar09_Config.cfg
@@ -84,7 +84,7 @@
 	{
 		name = ModuleEngineConfigs
 		type = ModuleEnginesAJEJet
-		configuration = Atar09K-50
+		configuration = Atar09B
 		modded = false
 		origMass = 1.582
 
--- a/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/Avon_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/Avon_Config.cfg
@@ -127,7 +127,7 @@
 		name = ModuleEngineConfigs
 		type = ModuleEnginesAJEJet
 		useConfigAsTitle = false
-		configuration = Avon302
+		configuration = Avon107
 		origMass = 1.300
 		
 		CONFIG
--- a/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/CF6_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/CF6_Config.cfg
@@ -129,7 +129,7 @@
 	{
 		name = ModuleEngineConfigs
 		type = ModuleEnginesAJEJet
-		configuration = CF6-50E2
+		configuration = CF6-6
 		modded = false
 		origMass = 3.709
 
--- a/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/F100_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Jet_Engine_Configs/F100_Config.cfg
@@ -127,7 +127,7 @@
 	{
 		name = ModuleEngineConfigs
 		type = ModuleEnginesAJEJet
-		configuration = F100-PW-229
+		configuration = F100-PW-100
 		modded = false
 		origMass = 1.442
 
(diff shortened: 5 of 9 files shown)


Capkirk123:
This was explicitly done to maintain backwards compatibility.

arrowmaster:
Can this be changed in the future then when the compatibility is no longer needed? It seems very unintuitive to have parts that need to have their MEC changed before they can be used.

Capkirk123:
I don't see why not.

Capkirk123:
When a craft using an engine that has been converted to MEC is saved, it should write the config version to the craft file, and subsequently remember what version is selected even if the default version is changed. Therefore, reverting the default config to the first config after a few updates should be safe.

This could still cause issues with craft files that are not saved in the transition period, but this should be relatively minor.

Capkirk123:
I dunno how many update cycles we wanna wait before merging this. If you are reading this in 2024, please merge this.

arrowmaster:
If anyone wants the changes from this PR before it gets merged in the future, you can try this MM patch.

`
@PART[*]:HAS[#engineType[AL31]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[AL-31FM]]
	{
		@configuration = AL-31F
	}
}
@PART[*]:HAS[#engineType[Atar09]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[Atar09K-50]]
	{
		@configuration = Atar09B
	}
}
@PART[*]:HAS[#engineType[Avon]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[Avon302]]
	{
		@configuration = Avon107
	}
}
@PART[*]:HAS[#engineType[CF6]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[CF6-50E2]]
	{
		@configuration = CF6-6
	}
}
@PART[*]:HAS[#engineType[F100]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[F100-PW-229]]
	{
		@configuration = F100-PW-100
	}
}
@PART[*]:HAS[#engineType[F404]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[F404-GE-402]]
	{
		@configuration = F404-GE-400
	}
}
@PART[*]:HAS[#engineType[J57]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[J57-P-21]]
	{
		@configuration = J57-P-8
	}
}
@PART[*]:HAS[#engineType[J75]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[J75-P-19]]
	{
		@configuration = J75-P-17
	}
}
@PART[*]:HAS[#engineType[J79]]:AFTER[RealismOverhaulEngines]
{
	@MODULE[ModuleEngineConfigs]:HAS[#configuration[J79-GE-17]]
	{
		@configuration = J79-GE-3A
	}
}
`

NathanKell:
Hmm, should we do this in RP-0? I believe we do that with the J-2 as well, since RO defaults it to the Saturn V J-2 but RP-0 pulls it back to the 200klbf one. Unless that's changed recently?

Capkirk123:
> Hmm, should we do this in RP-0? I believe we do that with the J-2 as well, since RO defaults it to the Saturn V J-2 but RP-0 pulls it back to the 200klbf one. Unless that's changed recently?

There’s no reason to not do that. I believe the LR87 also selects the Titan II config by default.

NathanKell:
I believe you're correct there. Probably H-1 as well. I added a bunch of early configs when we first made RP-0, but left the defaults as is in RO.

arrowmaster:
The LR87 (I checked all 3 parts in game) all default to the LR87-AJ-3 with RP-0 installed.

The initial problem I had with the original change is that it is a terrible user experience to have parts that unlock in one tech node to have a default config that is unlocked in a later tech node. But Kirk explained that this was needed from his change to not break crafts that existed prior to these parts getting MEC configs. I tried looking through RP-0 repo and cannot see where MEC defaults might be changed unless its happening in C#?

NathanKell:
lol it's in RO after all.
https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/LR87_Config.cfg#L147

NathanKell:
So basically I'd suggest following that form in the PR--change the default with an @configuration

Probably ideally we'd do that in RP-0 (hence my thinking that was where we did it) but eh, this is the pattern we currently have.

arrowmaster:
So this still doesn't change the original problem. In a previous version of RO these parts did not have MEC configs. When Kirk added the MEC configs he set the defaults to match the pre MEC configs so that it would not break craft built prior to the change by having a too weak engine now. No matter how it gets changed it needs to stay in the current behavior until it is acceptable to break upgrading from a version prior to having MEC configs for these parts.

Once it is acceptable to make the change, the default with RP-0 installed needs a config that is unlocked at the same time as the parts. Changing the default between versions that already use MECs shouldn't have any impact other than user experience.

NathanKell:
Ah gotcha. Right. Whoops.

Guess we wait 2-3 months or something?

arrowmaster:
This could be merged in now with the rest of the breaking changes but I think there is still an engine added after Kirk's Water Injection patch that needs added to this.

I don't currently don't have a working install to test against with the on going PLC merges.

NathanKell:
Ok, I'll double-check the J47 and J75.

PR #2790, opened 2023-02-26
Create a unified Deprecation method that can be shared across RO mods

arrowmaster:
Set ROdeprecated = True on parts to have them automatically renamed and their description updated. If a folder exists in GameData named NoRODeprecated then soft hide all deprecated parts the same as NoNonRO and NoNonRP0 would do. If a folder exists in GameData named HardRemoveRODeprecated thend hard remove all deprecated parts the same as HardRemoveNonRO and HardRemoveNonRP0 would do. ROdeprecated can be used in other RO mods so that deprecation behavior is uniform across all the RO mods instead of the current behavior.

### DIFF
--- a/GameData/RealismOverhaul/Filter_Extensions_Configs/non_WIP_RO_FE_config.cfg
+++ b/GameData/RealismOverhaul/Filter_Extensions_Configs/non_WIP_RO_FE_config.cfg
@@ -8,9 +8,30 @@
 {
     @SUBCATEGORIES
     {
+		list = 95,RO-DEPRECATED
         list = 96,WIP-RO
         list = 97,non-RO
     }
+}
+
+//  ==================================================
+//  Create the RO-DEPRECATED category.
+//  ==================================================
+
+SUBCATEGORY
+{
+	name = RO-DEPRECATED
+	displayName = RO-DEPRECATED
+	icon = R&D_node_icon_nuclearpropulsion
+
+	FILTER
+	{
+		CHECK
+		{
+			type = title
+			value = (DEPRECATED)
+		}
+	}
 }
 
 //  ==================================================
--- a/GameData/RealismOverhaul/Filter_Extensions_Configs/non_WIP_RO_FE_config.cfg
+++ b/GameData/RealismOverhaul/Filter_Extensions_Configs/non_WIP_RO_FE_config.cfg
@@ -8,7 +8,7 @@
 {
     @SUBCATEGORIES
     {
-		list = 95,RO-DEPRECATED
+        list = 95,RO-DEPRECATED
         list = 96,WIP-RO
         list = 97,non-RO
     }
--- a/GameData/RealismOverhaul/Parts/Probes/ProbeVanguardXray.cfg
+++ b/GameData/RealismOverhaul/Parts/Probes/ProbeVanguardXray.cfg
@@ -24,7 +24,7 @@
 	!MODULE[TweakScale]{}
 
 	%RSSROConfig = true
-	%deprecated:NEEDS[RP-0] = True
+	%RODeprecated:NEEDS[RP-0] = True
 
 	@mass = 0.01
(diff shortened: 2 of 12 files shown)


arrowmaster:
While I had checked ROCapsules and ROEngines to find there was inconsistent deprecation of parts and no use of the existing `deprecated = true`, I missed usage of it in RP-0. So some parts are reappearing from RP-0 with this current change but are easily converted in a change for RP-0.

arrowmaster:
RO-RD9B is getting flagged as deprecated because its a clone of RO-J57P8 which has been deprecated. Will need to fix. Fast fix is to unset the deprecated flag, better fix would be to clone a different part.

arrowmaster:
Can I please have some external input on this before I finalize it and take it out of draft. I just want to verify its on a good path.

Is changing from `deprecated = True` to `ROdeprecated = True` good or should it stay with the original.

The plan after this is merged is to PR Express Install to soft hide deprecated parts. And to PR the other RO repos to change from any local deprecation scheme to all use the same deprecation logic from RealismOverhaul.

GameData/RealismOverhaul/Filter_Extensions_Configs/non_WIP_RO_FE_config.cfg
```diff
@@ -8,11 +8,32 @@
 {
     @SUBCATEGORIES
     {
+		list = 95,RO-DEPRECATED
         list = 96,WIP-RO
         list = 97,non-RO
     }
 }
 
+//  ==================================================
```
StonesmileGit, line 17: Does it make sense to have a category where you can see and use the deprecated parts? The main point of deprecating si to have the parts be removed from circulation over time, so that no crafts are affected when it is finally deleted.

StonesmileGit (commented):
Something to think about; `ROdeprecated` vs `RODeprecated` (Do note the capitalization). Is the latter more in line with other tags used in RO?

Capkirk123:
As the king of poorly deprecating parts, I do think this is a good idea.

GameData/RealismOverhaul/Filter_Extensions_Configs/non_WIP_RO_FE_config.cfg
Capkirk123: I second this, if something is deprecated you probably shouldn't be able to see it at all.

arrowmaster: The logic is copied from the soft and hard removal of nonRO and nonRP0 parts. If wanted I could strip it down to just always hiding parts with no control but that does make it harder to verify the deprecation is working as intended. I already found and fixed that RO-RD9B was being deprecated when it should not have been.

Because of how FilterExtensions works, its not possible to completely hide unlocked parts. I think moving to another category helps keep them from showing up in the main categories but setting to -1 may be good enough too.

Working on the assumption that :NEEDS[NoRODeprecated] passes because we have added it to the Express Install. When a user upgrades RO/RP-1/etc any unpurchased parts that have been newly deprecated will be moved to ORPHANS and completely hidden. However because KSP saves part purchases into the savegame under the tech tree node they were purchased from, any already purchased parts will remain unlocked. If the user does not have FilterExtensions they will not be able to find the part, but with FilterExtensions it is always possible to find the part if you go digging hard enough through the parts by module or resource type filtering.

StonesmileGit: The use case of finding incorrectly deprecated parts is a good motivation for having something like this. Could it be made more clear that the deprecated category shouldn't be used or something?

arrowmaster:
Rebased and minor updates. Next step is to update it to work better with #2838

StonesmileGit (approved):

PR #2793, opened 2023-03-05
Deprecate the Engine Pre-cooler and Engine Nacelle.

arrowmaster:
These parts only function as fuel tanks and have incorrect descriptions implying they have functions that they do not. Lets hide these noob traps.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Structural.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Structural.cfg
@@ -46,7 +46,7 @@
 @PART[radialEngineBody]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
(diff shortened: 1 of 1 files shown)


arrowmaster:
I didn't notice when making these that they do actually have an intake configuration. That intake configuration is being stripped out somewhere as everyone reports these parts do not have an intake in game. The parts should either be deprecated or the both the intake configuration needs fixed to not be stripped out AND the descriptions need updated to remove false information about such as precooling air that make it sound like they do more than just function as an intake.

Capkirk123:
Does this rely on https://github.com/KSP-RO/RealismOverhaul/pull/2790?

arrowmaster:
> Does this rely on #2790?

No this uses the existing deprecation behavior. If this PR gets merged then #2790 will need updated to change these to use its new behavior.

arrowmaster:
Updated to be compatible with #2790

PR #2795, opened 2023-03-05
Let ignoreMass, massOffset, etc work for bimodal engines

### LINKED ISSUE #2794: EngineVariants_Config.cfg doesn't support bimodal engines
[EngineVariants_Config.cfg](https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/EngineVariants_Config.cfg) applies a bunch of patches to ModuleEngineConfigs, and is unaware that there are ModuleBimodalEngineConfigs now.

Haven't gone looking whether any existing parts are impacted; hopefully not. But will be confusing to anyone trying to use these features with bimodal engines in the future.

I'll be fixing the massOffset bit because I'm poking at it; but won't be touching the rest.

lpgagnon:
fixes #2794

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/EngineVariants_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/EngineVariants_Config.cfg
@@ -5,7 +5,7 @@
 
 @PART[*]:HAS[#engineTypeMult[*],~minActiveEngines[*]]:FOR[RealismOverhaulEnginesPost]
 {
-	@MODULE[ModuleEngineConfigs]
+	@MODULE[Module*EngineConfigs]
 	{
 		@origMass *= #$../engineTypeMult$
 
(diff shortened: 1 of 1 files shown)


al2me6:
IMO you should just change the original patch to target `Module*EngineConfigs` instead.

lpgagnon:
I'd worry about that catching things it shouldn't, but whatever maintainers think is best.

Capkirk123:
I also support converting everything to Module*EngineConfigs, all of those patches should be safe to run on ModuleBimodalEngineConfigs.

lpgagnon:
I can change them all in this PR, but I won't test anything beyond what I've already tested

lpgagnon:
done. https://github.com/KSP-RO/ROEngines/pull/188 still works with the switchover; no further testing done

Capkirk123:
Tested, nothing obviously breaks. Should be good to go.

PR #2799, opened 2023-03-07
Decrease maximum speed for large landing gear

Test-Account-KSP:
Large landing gear have maximum speed of 300m/s, which is completely unrealistic. The tyres would be instantly destroyed.
This is a proposal to decrease maxSpeed to 150m/s, based on the fastest landing speeds I could find:
X-37 (unmanned drone) - 134m/s
Emergency landing of a damaged F-15 - 133m/s  (https://theaviationist.com/2014/09/15/f-15-lands-with-one-wing/)
Emergency landing of a damaged F-104 - 120m/s   (02 June 1965)
Space Shuttle, Buran - about 100m/s

!!! 150m/s allows some resizing of the gear, but much less than before, which will unfortunately break some craft.


marsh1832:
Seems good to me, it gives a smoother progression. Resizing gear is nice but it does create potential for ☣️ small wheels with very high speed limits

Capkirk123:
Seems pretty reasonable to me.

PR #2802, opened 2023-03-18
Fix old Juno IV configs, add new ones

Capkirk123:
Deprecate old Juno 6k and 45k configs and create new, more accurate ones. Also add configs for Juno IV growth options, including ClF3 upper stages and UDMH/LOX S-3D.

Needs https://github.com/KSP-RO/RealismOverhaul/pull/2838

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Juno45k_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Juno45k_Config.cfg
@@ -38,6 +38,8 @@
 //	Sources:
 
 //	Source: JPL 20-123 Juno IV Rocket Vehicle System: https://archive.org/details/nasa_techdoc_19730064143
+//	https://www.si.edu/media/NASM/NASM-NASM.2020.0031-Interim_Report_to_the_Special_Committee_on_Space_Technology_box1_folder4-000002.pdf
+//	http://libarchstor2.uah.edu/digitalcollections/items/show/11031
 
 //	Used by:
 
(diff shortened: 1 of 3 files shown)


Capkirk123:
RODeprecation tags applied, should be good to go once #2838 is merged

ilmcamam:
Is there a primary source that explicitly states 150klbf design thrust specifically for Juno IV? Having stumbled upon [two](https://www.si.edu/media/NASM/NASM-NASM.2020.0031-Interim_Report_to_the_Special_Committee_on_Space_Technology_box1_folder4-000002.pdf) [documents](http://libarchstor2.uah.edu/digitalcollections/items/show/11031) that suggest otherwise (165klbf), it seems a little strange to have such a low TWR (~1.1). If this is for Juno IV in particular (rather than Jupiter as a whole), would it make sense to have the higher thrust?

Capkirk123:
It does appear that some performance growth was expected for the Jupiter first stage of Juno IV.

arrowmaster:
I think this and https://github.com/KSP-RO/RP-0/pull/1925 should have been merged to master instead of a PLC branch. This shouldn't be save breaking because new engine configs were created and the old one left in place for any already built crafts.

NathanKell:
Oh, hmm, yeah. It redirects to the new ECM but I think that's ok, RF should treat it as unlocked? (I wasn't worried about the RO side, just the RP-1 side, but I can't merge the RO side to master and the RP-1 side to PLC because that would leave the new configs with no requirement or cost)

PR #2804, opened 2023-03-23
Revise Viking and Vikas

Capkirk123:
Revise the performance of the Viking and Vikas engines based on French sources. Also adjust the properties of UH25.
Changes include:

- Adjust ISP and thrust of all Viking configs based on French sources and RPA calculations
- Add Viking 5, 5B and 4C configs, and remove 2B config
- Revise and reorganize Vikas configs
- Adjust density and Hsp of UH25 to be more accurate

This is, unfortunately, mildly save breaking, so it should be held until the next save-breaking update (release of PLC?)


Capkirk123:
Actually, since this isn't waiting for the deprecation feature, it's ready to merge with the next save-breaking update.

PR #2808, opened 2023-03-25
BDB Antennas Configs

SkyPhoenix999:
RO Configs for the Antennas Folder


Capkirk123 (changes requested):
Most of these are deprecated, there's no reason to configure them

PR #2823, opened 2023-04-01
BDB Generic Configs

SkyPhoenix999:
All the General Stuff from BDB, including Cleanup, Payload Decouplers, KIS, Procedural Fairing Bases, and more


Capkirk123 (approved):

GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/Generic/RO_General_Stuff.cfg
```diff
@@ -22,11 +22,21 @@
 }
 
 
-// MPF Part Cleanup (SAF and PF Fairings kept, MPF not used in RO)
+//	Tank Mass Fix
+@PART[bluedog*,Bluedog*]:HAS[@MODULE[ModuleFuelTanks]]:AFTER[RealismOverhaulTankTypes]
+{
+	@MODULE[ModuleFuelTanks],*
+	{
+		@basemass = -1
```
Capkirk123, line 30: Please do not do this

SkyPhoenix999: idk what else is going to work then, because the RF Config keeps adding mass to the tanks, for small parts like Able the added mass makes it almost unusable.

Capkirk123: I have not experienced this issue. They may be configured wrong, or it may be BDB itself adding them mass (BDB multiplies the mass of parts as a function of the solar system rescale, a feature which should be removed with the next release but has not yet been)

SkyPhoenix999: I've installed that fix you made for the mass.

From everything I've checked it's not BDB, it's the RF config, the basemass part of each TANK_DEFINITION will add mass to the tanks once propellant is set on the tank.

Capkirk123: Then it's a config issue on the RF side.

Capkirk123: Can you point out a specific part that suffers this issue?

SkyPhoenix999: Technically speaking it's most parts

But the one I was testing on was bluedog_DeltaB_Tank from the Able parts

Capkirk123: Well bluedog_DeltaB_Tank has two tanks in it, one of which is a service module, which would explain why it's so heavy. I suspect the rest are also config issues of a similar nature.

SkyPhoenix999: Checking the configs (The Able PR configs) there's only one Tank. I'm not sure why there would be a second service module tank in there. #2807 

Will test further soon.

Capkirk123: image
And there you have it.

Base mass is being set in the RFTank module, which means RF will apply base mass as it sees fit.

RFTanks should be configured one of two ways. Either basemass should be set to -1, in which case RF will not try to alter the tank mass in any way, or basemass should be set to a sensible value for an empty structural shell, and RF allowed to add the appropriate tank mass such that the total dry mass adds up to the appropriate value. I would much prefer the latter method be used, but the former is acceptable for dumb fuel tanks and such if you don't want to spend the time setting the mass appropriately.

basemass should be set to -1 on a part-by-part basis, and not bulk patched to -1, since _I_ prefer to configure parts using the second method.

Capkirk123: Ok, Able has been fixed [here](https://github.com/KSP-RO/RealismOverhaul/commit/5e01c63c2280392d495ef98adf1d8eb079823273). Any other dry mass issues should be fixed in the same manner.

SkyPhoenix999: Made General corrections to the Able configs [here](https://github.com/KSP-RO/RealismOverhaul/pull/2956) with your suggestions and additions, corrected tank masses and relocating waterfall to match with the rest of the BDB PRs

PR #2828, opened 2023-04-03
BDB LDC Configs

SkyPhoenix999:
Configs for LDC parts from BDB.

Require the experimental RealFuels and B9PS dlls that support ModuleFuelTanks


Capkirk123 (approved):
Sure

PR #2829, opened 2023-04-09
Restore TF configs for generic thrusters

Capkirk123:
Restore TF configs for generic thrusters (only, since TF apparently does not support RCS failures). Configs are fairly generous, since they have to support a very wide range of thrusters and eras.

All configs get a 10 hour burn time (this should be enough for almost everything). Data rates are tuned so 10,000 du should be reached after ~100 minutes of operation (shorter in the event of a failure rewarding extra du).
Thruster types have been given the following stats

Cold Gas Thrusters (inert, only likely possible failure modes from valves)
ignition 0.98 -> 0.9999
cycle 0.99 -> 1.0

Monopropellant Thrusters (simple, but bed erosion and burn-through possible along with valve failure)
ignition 0.96 -> 0.9995
cycle 0.975 -> 0.9995

Bipropellant Thrusters (more complex, reliability based on AJ10)
ignition 0.90 -> 0.9994
cycle 0.95 -> 0.9994

This should give some incentive to choose something other than bipropellant thrusters the second they become available, and encourage installing backup engines, similar to many real space probes like Cassini-Huygens.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RCS_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RCS_Config.cfg
@@ -415,56 +415,68 @@
 	//FIXME: This should probably be shown in PAW, similar to MERF
 	@description ^= :$: <b><color=red> RCS requires High-Pressure tanks to function.</color></b>
 }
-@PART[*]:HAS[~useRcsMass[True],#engineType[RCSGeneric]]:FOR[RealismOverhaulEngines]
+@PART[*]:HAS[#engineType[RCSGeneric],@MODULE[ModuleEngines*]]:FOR[RealismOverhaulEngines]
+{
+	//just tag generic thrusters, since TF doesn't apply to RCS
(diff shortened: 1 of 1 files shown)


siimav:
I believe this can't be merged before RCS also has failures. Otherwise knowledgeable people would just replace all generic thrusters with oversized RCS parts.

Capkirk123:
I disagree. As is, a player that wishes to face _realistic_ reliability concerns when using generic thrusters cannot. Adding reliability to generic thrusters but not RCS means players will face more realistic constraints when designing spacecraft. Players that wish to exploit the properties of RCS still will be able too, but that's not a reason to not implement this feature at all.

marsh1832:
I second Kirk. This isn't that much different than using verniers without the main engine or its turbopump. If players want to cheese then they will, and they will get roasted for it in discord lol. Perhaps there will be a day for RCS failures, but that isn't a reason to not increase the realism of generics.

PR #2835, opened 2023-04-20
Update Raptor reliability

LuneoLucy:
nerf Raptor with flight test data

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Raptor_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Raptor_Config.cfg
@@ -163,9 +163,9 @@
 					key = 1.00 1.00 3 3
 				}
 
-				ignitionReliabilityStart = 0.802 // 1.98% of engines failed ignition on SH [9]
+				ignitionReliabilityStart = 0.736 // 2.64% of engines failed ignition on SH [9]
 				ignitionReliabilityEnd = 0.995 // ~ Slight Chance of the Torch-lit system failing
-				cycleReliabilityStart = 0.802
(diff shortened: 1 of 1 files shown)


Test-Account-KSP:
Ignition failure is irrelevant if pre-launch failures are off, which is the default setting.
Need to nerf cycleReliabilityStart to match IRL performance even when pre-launch failures are off.

LuneoLucy:
This should be ready for review I think

LuneoLucy:
Modifications have been tested in game on a clean express install, I am unable to troubleshoot the issue with the outer raptor testflight config, it appears correctly in the VAB but changes to 100% succes with 0s of rated burn time in flight

LuneoLucy:
The issue comes from the original config, I don't have enough knowledge to fix it myself so any help is welcome

Capkirk123:
> Ignition failure is irrelevant for first stage engines if pre-launch failures are off, which is the default setting. Need to nerf cycleReliabilityStart to match IRL performance even when pre-launch failures are off.

You should not do this

Capkirk123:
There are a few issues here. Most notably, Superheavy had Raptor 2s installed, not Raptor 1s (citation needed), so these configs are arguably being applied to the wrong engine.

Also, please use the TF engine reliability config tool attached to calculate failure chances.
[KSP Engine Reliability Calculator V2.0 (1).xlsx](https://github.com/KSP-RO/RealismOverhaul/files/11304731/KSP.Engine.Reliability.Calculator.V2.0.1.xlsx)

LuneoLucy:
I tried to fix the original TF config but it requires knowledge that I do not have, so PR is closed

PR #2845, opened 2023-05-03
Adding RSRM-I, RSRM-II and RSRM-III configs from the SRB-X studies

Its-Just-Luci:
Name has the basic details, I also included configs to enable the matching motors from photon corp as it's the mod I use, and the only one I know with these motors in it...


Its-Just-Luci:
Thrust curves need to be corrected, but this seems like a relatively easy (although unknown to me) thing to do

PR #2848, opened 2023-05-12
Add Titan variance configs

Capkirk123:
Add variance settings to LR87/91 based on the 1961 Launch Vehicle Handbook.

This is potentially slightly savebreaking, as crafts using the Titan 1 engines will experience a slight decrease in performance.


NathanKell:
Oh hell yes. This is why I added all those vars, in the hope someone (cough YOU cough) would do this :D

PR #2850, opened 2023-05-17
Add oxygen gas to HP tanks for the new RCS configs

TwistedGiraffe:
#2754 added several RCS configs that use oxygen gas as the oxidizer, which can only be added to SM tanks. This pull request will allow adding oxygen gas to HP tanks after the tech node `standardDockingPorts`, which unlocks such RCS configs. 

Needs https://github.com/KSP-RO/RP-0/pull/1939.


Capkirk123:
Looks good

PR #2854, opened 2023-05-22
Deprecate the SXT LEM if ROCapsules is present

arrowmaster:


arrowmaster:
Resolves https://github.com/KSP-RO/ROCapsules/issues/93

PR #2857, opened 2023-06-01
Delay tag removals so other mods can access them

TwistedGiraffe:
Move tag removals to `FOR[zzzTagCleanup]` a la https://github.com/KSP-RO/RealismOverhaul/blob/3de3a8cb47c64b14540b342e3c58502845f6458e/GameData/RealismOverhaul/RO_Deprecation.cfg#L11-L14. 
Thank @arrowmaster for the idea.


NathanKell:
Probably don't need to keep the lead, balloontemp, etc stuff around but it's fine. Cheers for the fix!

PR #2859, opened 2023-06-08
Add full names and abbreviations to resources

Capkirk123:
Since you can apparently specify separate abbreviations and display names for resources, give our resources more verbose display names.


NathanKell:
Given screen real estate premium, I'm a bit concerned about changing many resources to use two lines in the RF GUI?

Capkirk123:
Fair enough. Is the RF UI a fixed size? If so, I could try to keep things on one line.

NathanKell:
Height scrolls but width is fixed, yeah.

Capkirk123:
Names trimmed to fit on one line.

PR #2873, opened 2023-06-29
Add RD-0124MS

Its-Just-Luci:
Adding the Soyuz-5 second stage engine using data/info from russianspaceweb's 2023 postings.


Capkirk123:
Has conflicts

Capkirk123:
There's tons of files here that shouldn't be here, and it also seems to be trying to merge a very old copy of the RD-0124 config into the current one

PR #2876, opened 2023-06-30
Material based temps tagging

Capkirk123:
Add material tags to common RO parts, such as RO's own parts, squad/SXT/Restock, etc.


Capkirk123:
Almost everything in an express install should now be tagged. I'll do Kerbalism, RP-1, and the other ROMods in another PR. For now, ship it!

PR #2886, opened 2023-07-27
Shorten really long engine descriptions

NovemberOrWhatever:
Also added localisations for some of the longer engines that needed it.
Hopefully it's fine to switch to plume and sound from Waterfall, as it's a bit shorter.
Viking (5/6) and the AJ260 radial will need scroll bars until I finish localizing ROE
Currently, only the NAA75, LR89, LR105, Agena, RD58, Castor 4 and AJ260 SL radial have scroll bars, and those descriptions have all been shortened significantly.
Removed advice on how to use an RL10 as that should be clear from its stats.
In general, some review would be nice. I've tested it, but I have had to elide some information and want to make sure I'm not removing too much

### DIFF
--- a/GameData/RealismOverhaul/Localization/en-us-Engines.cfg
+++ b/GameData/RealismOverhaul/Localization/en-us-Engines.cfg
@@ -60,12 +60,12 @@ Localization
 		//		AJ260FL
 		#roAJ260FLTitle = AJ-260 FL
 		#roAJ260FLDesc = The AJ-260 was the largest rocket motor ever tested. This is the full length version of the engine. With a gross mass of 1,648 tons, it provides a thrust of 35,391 kN for more than 2 minutes. This version was designed as an alternative to the Saturn IB and as a booster for Apollo Applications launch vehicles.
-		#roAJ260FLSideTitle = AJ-260 FL
+		#roAJ260FLSideTitle = AJ-260 FL Radial Solid Rocket Booster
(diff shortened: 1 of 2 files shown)


GameData/RealismOverhaul/Localization/en-us-Engines.cfg
```diff
@@ -32,10 +32,10 @@ Localization
...
 		//		Agena
 		#roAgenaTitle = XLR81 (Agena) Vacuum Engine
-		#roAgenaDesc = A gas-generator nitric acid/UDMH vacuum engine used on Agena. The XLR81 family was derived from the Bell Hustler Rocket Engine, originally developed for use on an air-to-surface missile. Early engines were nearly identical to the Hustler engine, while later variants offered new capabilities and improved performance. Engine restart was introduced on the Agena B's XLR81-BA-7 (Model 8081). The XLR81-BA-11 (Model 8096) used on Agena D used propellant sumps to eliminate the need for ullage thrust. The XLR81-BA-13 (Model 8247) powered the Gemini Agena Target Vehicle (a modified Agena D) and was rated for up to 14 restarts.
+		#roAgenaDesc = A missile-derived gas-generator engine used on Agena. Early engines were nearly identical to the military version, while later variants improved capabilities and performance. Engine restart was added on the Agena B's XLR81-BA-7 (Model 8081). The XLR81-BA-11 (Model 8096) used on Agena D had propellant sumps to handle ullage. The XLR81-BA-13 (Model 8247) powered the Gemini Agena Target Vehicle (a modified Agena D) and was rated for up to 14 restarts.
```
NathanKell, line 35: I think this changes the meaning too strongly? "Missile-derived" make people think of, well, ballistic missiles, rather than AGMs. 
Instead try "Originally developed to power a standoff weapon for the B-58 Hustler, the XLR81 is a gas-generator engine with storable propellants. Later space variants improved performance and capability."
then it's fine except I'd go back from 'had propellant' to 'used propellant'.

PR #2892, opened 2023-08-06
Config cleanup and some more improvements

Capkirk123:
Continue to work on cleaning up engine configs, adding sources, etc. 
Also make improvements to a few of the more neglected configs. Changes include:

- Make Bell Model 117 a little lighter since it has no gimbal
- Use RPA to try to estimate BE-3 performance a little better
- Complete overhaul of E-1 configs. Using the E-1 Dynasoar RBS version, the only config for which comprehensive data exists, create configs of 380, 468, 500 and 575 klbf SL thrust, using RPA and the H-1 to estimate performance data. Old E-1 configs have been deprecated, and the Kolyma's Shadow E-1A is kept mostly the same
- Improve H-1 SL ISP using data from Skylab 2 and the H-1 users manual
- Revise HG-3 SL performance using RPA
- Reduce XLR83 T/W to a sane value by increasing dry mass
- Reduce LR87-LH2 T/W to a sane value by increasing dry mass. Also revise performance a little using RPA
- Correct MR-80/MR-80B SL ISP (it was previously using ISP values from Mars "sea level")
- Completely revise NK-9 engine configs based on better sources
- Completely revise NK-9V engine configs based on better sources
- Completely revise NK-15/33 engine configs based on better sources. Also add AJ26-58 and AJ26-59 configs.
- Completely revise NK-15V/43 engine configs based on better sources. Also add AJ26-60 config.
- Reduce mass of RD-0236
- Use RPA to better estimate SL performance of RL10-CECE configs
- Configure RS-2100 to use subcooled propellants
- Correct RS-68B ISP
- Completely revise RS-88/LAE configs based on better sources

Although many changes are made, this should not be save-breaking. Many changes were made, but most were minor. The only significant performance reductions were to the NK-15, NK-15V, RS-88, and LAE.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/18KS7800_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/18KS7800_Config.cfg
@@ -10,7 +10,7 @@
 //	Dry Mass: 27.21 Kg
 //	Thrust (SL): ??? kN
 //	Thrust (Vac): 34.09 kN
-//	ISP: ??? SL / 238.8 Vac
+//	ISP: 200? SL / 238.8 Vac
 //	Burn Time: 1.8
 //	Chamber Pressure: 6.89 MPa
 //	Propellant: NGNC
--- a/GameData/RealismOverhaul/Engine_Configs/25KS18000_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/25KS18000_Config.cfg
@@ -10,7 +10,7 @@
 //	Dry Mass: 117.9 Kg
 //	Thrust (SL): 82.7 kN
 //	Thrust (Vac): ??? kN
-//	ISP: 178 SL / ??? Vac
+//	ISP: 178 SL / 198.8? Vac
 //	Burn Time: 2.5
 //	Chamber Pressure: 9.24 MPa
 //	Propellant: NGNC
--- a/GameData/RealismOverhaul/Engine_Configs/A-4_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/A-4_Config.cfg
@@ -27,7 +27,7 @@
 //	Thrust (Vac): 288.68 kN
 //	ISP: 220 SL / 255 Vac
 //	Burn Time: 115
-//	Chamber Pressure: ??? MPa
+//	Chamber Pressure: 1.7? MPa	estimated with RPA
 //	Propellant: LOX / Hydyne
 //	Prop Ratio: 1.3253, same as A-6H?
 //	Throttle: N/A
--- a/GameData/RealismOverhaul/Engine_Configs/AJ10_Early_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/AJ10_Early_Config.cfg
@@ -9,7 +9,7 @@
 //
 //	Dry Mass: 84 Kg
 //	Thrust (SL): ??? kN
-//	Thrust (Vac): 35.585 kN
+//	Thrust (Vac): 33.8 kN
 //	ISP: 122 SL / 271 Vac		SL calculated with RPA
 //	Burn Time: 115
 //	Chamber Pressure: 1.4 MPa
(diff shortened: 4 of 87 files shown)


NathanKell:
Nice work!!
Question on the E-1 -- should the Isp really not increase? I would expect some improvement there, in the way that LR89s improved over time.

Capkirk123:
No, not really. I assumed it was a relatively modular design like the H-1, where chamber pressure was increased without making any changes to the injector.

NathanKell:
If you're increasing chamber pressure wouldn't you also increase the nozzle AR? LR87-AJ-5 -> LR87-AJ-9 style.

Capkirk123:
That's a can of worms I didn't really want to open, and besides, the H-1 and F-1 didn't

NathanKell:
Under the (speculative) world where the E-1 remains in regular use, I would expect it to get upgrades like the in-regular-use LR87, rather than the one-and-done H-1 and F-1.

NathanKell:
(one-and-done is slightly unfair to H-1, but the work on them broadly stopped in the early-mid 60s, right? Except for that 250klbf study. The "in regular use" H-1s were remanufactured into RS-27 rather than actually being new engines.)

Capkirk123:
I mean, that was the entire point of the modular design applied to the E-1, F-1, H-1, etc. That it could be upgraded to different roles without significant design changes, unlike first generation engines. And it does get the O/F leaned out in the later versions to help increase performance.

Capkirk123:
Also fwiw, the two datapoints we have on the E-1, the test stand model and the Dynasoar booster proposal, show very similar ISP despite a significant (380 to 468 klbf) increase in thrust

NathanKell:
Well, IMO it sucks for gameplay reasons to have basically no choice as US but to use LR87s if you want a specific impulse above 290 (discounting F-1), but I take your point.

NathanKell:
Also if you're renaming configs, you should use the upgrade tool provided (add entries to UpgradeData.cfg )

NathanKell:
@Capkirk123  More E-1 partisanship: since it unlocks later than the H-1, would we expect its in-service Isp to be comparable (or higher) to H-1? Or would there be too many losses from the gas generator? Because it has a higher chamber pressure and a higher AR nozzle, and yet a worse vacuum Isp. So I'm confused.

And does the UpgradeData.cfg file make sense?

Capkirk123:
Well, the one reliable source, dated May 1960, states an expected performance of 256/291. 

Throwing it into RPA, that seems to suggest a combustion efficiency similar to the F-1. The Rocketdyne baffled injectors that they used to deal with combustion stability at that scale really hurt efficiency. The F-1 just ran at higher pressure and had the film-cooled nozzle extension.

Although such a nozzle extension could also be fitted to the E-1, since it runs at lower pressure, this also hurts sea level ISP. That opens the can of worms as to what the E-1 would be used for. In a Titan-like role, with an airlit core, that tradeoff makes sense. In a more Kolyma's Shadow role, with a monolithic E-1 core (also suggested by the USAF study), this would likely not be performed.

I'm not sure using the upgrade cfg is a great idea. The old configs are still there, just deprecated, and the new configs are relatively dissimilar.

Capkirk123:
Obviously, with 1.5 reliable sources, this leaves a lot of latitude for creative config work. I chose the conservative approach, simply mirroring the upgrades applied to the H-1.

NathanKell:
Sorry, somehow I missed the notifications on these. Yeah, all that makes sense. Likely (as OTL for F-1 and H-1) E-1 upgrades would focus on thrust for a larger LV, rather than squeezing out extra performance. It's possible something like RS-27A might be done, as you surmise (either groundlit with a few UA1207s and a long-burning core, or airlit with many, many UA1207s).

Ok, at any rate I'm satisfied with the E-1. Thanks for humoring me. :) Consider this PR reviewed on my account, merge when ready!

Capkirk123:
Unless any further objections exist, I would like to merge this so I can get on with RD-0106/0107/0108 changes without fear of conflicts

PR #2895, opened 2023-08-08
Making generic solid rocket motors more consistent and realistic

Clayell:
This reduces the specific impulse of all of the generic solid rocket motors (spin motor, radial separation motor (small/medium/large), separation motor (small)) to 220 in a vacuum and 200 in the atmosphere. I'm unsure if these values are balanced, so I'm open to discussion for them. The purpose of this is to at least make them all consistent with each other.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1143,6 +1143,9 @@
 //  Spin Motor (small)
 //  ==================================================
 
+//Source for separation motors: https://www.northropgrumman.com/wp-content/uploads/NG-Propulsion-Products-Catalog.pdf (Page 49)
+//For dry mass:propellant mass, using a ratio of 1:0.9, a little better than the BSM
+
 @PART[ROSmallSpinMotor]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1183,7 +1183,7 @@
 			PROPELLANT
 			{
 				name = PSPC
-				ratio = 1.0
+				ratio = 2.069 //dry mass * 0.9 = propellant mass
 				DrawGauge = True
 			}
 		}
@@ -1200,7 +1200,7 @@
 	@title = Radial Separation Motor (Medium)
 	@description = Small solid motor use to help separate one stage from another or perform ullage. Best used with others.
 
-	@mass = 0.044
+	@mass = 0.032
 
 	@MODULE[ModuleEngines*]
 	{
@@ -1210,7 +1210,7 @@
 	MODULE
 	{
 		name = ModuleFuelTanks
-		volume = 20.34
+		volume = 16.552 //dry mass * 0.9 = propellant mass
 		type = PSPC
 		basemass = -1
 	}
@@ -1248,7 +1248,7 @@
 	@description = Larger solid motor use to help separate one stage from another or perform ullage. Best used with others.
 
 	%rescaleFactor = 2.0
-	@mass = 0.25
+	@mass = 0.256
 
 	@MODULE[ModuleFuelTanks]
 	{
@@ -1280,7 +1280,7 @@
 
 	@MODULE[ModuleFuelTanks]
 	{
-		@volume *= 0.25		// Should be *.125, but balancing vs snubOtron, which has volume=5.
+		@volume *= 0.25	//same mass and volume as SnubOTron
 	}
 
 	@MODULE[ModuleEngineConfigs]
@@ -1304,8 +1304,8 @@
 	{
 		%atmosphereCurve
 		{
-			@key,0 = 0 220
-			@key,1 = 0 200
+			@key,0 = 0 225
+			@key,1 = 0 190
 		}
 	}
 	@MODULE[ModuleEngineConfigs]
@@ -1314,8 +1314,8 @@
 		{
 			%atmosphereCurve
 			{
-				key = 0 220
-				key = 1 200
+				key = 0 225
+				key = 1 190
 			}
 		}
 	}
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
@@ -1153,7 +1153,7 @@
 	@title = Spin Motor (Small)
 	@description = Very small solid motor used to spin stabilize stages.
 
-	@mass = 0.004
+	@mass = 0.00381
 
 	MODULE
 	{
@@ -1165,7 +1165,7 @@
 		TANK
 		{
 		  name = PSPC
-		  amount = 1
+		  amount = 1.11
 		  maxAmount = 1
 		}
 	}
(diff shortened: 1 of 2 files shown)


NathanKell:
I mean, seems fine, but the reason these are horrific is because of their mass ratios, even though 250s Vac is fairly good for the time. It might make sense to make the mass ratio slightly better in that case (preserving total impulse vs when Isp was 250s)

NathanKell:
Good spot on the 230/220 one though, that's ludicrously low nozzle losses. Honestly I'd probably do something like 190/225 for them all, and then tweak the propellant amounts to keep burn time the same, and lower dry mass so total mass remains constant from before?

Clayell:
> Good spot on the 230/220 one though, that's ludicrously low nozzle losses. Honestly I'd probably do something like 190/225 for them all, and then tweak the propellant amounts to keep burn time the same, and lower dry mass so total mass remains constant from before?

Alright, thanks for the feedback.

Clayell:
I'll reduce the non-radial volume to .25 as well, to make it match with the radial version.

NathanKell:
Yep, that removal looks fine! You kept the one that adds the TF config, so that's still there.

Clayell:
> Yep, that removal looks fine! You kept the one that adds the TF config, so that's still there.

Alright good to know. I've run into a confusing problem though. In the config, it says that the separation motor (small) has a volume of only .25, but checking my previous in game screenshots shows it as around 5.09. Any idea what could cause this?

lpgagnon:
it's not 0.25, it's `*= 0.25`, i.e. 1/4 of its previous value, which is set here https://github.com/KSP-RO/RealismOverhaul/pull/2895/files#diff-a1c16e44803d091bae08f442608d5ac1832a1b75d133d315b40d0dff6d8b7b27R1213

Clayell:
> it's not 0.25, it's `*= 0.25`, i.e. 1/4 of its previous value, which is set here https://github.com/KSP-RO/RealismOverhaul/pull/2895/files#diff-a1c16e44803d091bae08f442608d5ac1832a1b75d133d315b40d0dff6d8b7b27R1213

Thank you!

Clayell:
Strangely, the dry masses of these separation motors seem to actually be realistic, as checking this source for the shuttle booster separation motor (https://www.northropgrumman.com/wp-content/uploads/NG-Propulsion-Products-Catalog.pdf) gives a dry weight of 35 kg and a wet weight of 76 kg. (aka a propellant mass of 31 kg, or ~1:1.13 ratio for propellant mass:dry mass) The "radial separation motor (medium)" has the closest ratio to this with about 1:1.243, so I'd say making this ratio about 1:1.1 for all the separation motors would give a good balance, which I'll do by changing the dry mass.

GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg
```diff
@@ -1143,6 +1143,9 @@
 //  Spin Motor (small)
 //  ==================================================
 
+//Source for separation motors: https://www.northropgrumman.com/wp-content/uploads/NG-Propulsion-Products-Catalog.pdf (Page 49)
+//For dry mass:propellant mass, using a ratio of 1:0.9, a little better than the BSM
```
NathanKell, line 1147: This actually should be a bunch worse, since this is supposed to be a 1940s solid, and the BSM referenced is a 1990s/2000s era solid, probably with a fiberglass casing. I'd suggest using 1.11 for the volume, to preserve the burntime.

NathanKell (changes requested):

NathanKell:
> Strangely, the dry masses of these separation motors seem to actually be realistic...

Per the review comment, that means they're wildly optimistic, since these are 40s or early 50s motors (which should probably be NGNC, but let's not open that can of worms, it'd probably put the variance too high) with steel casings, rather than 90s/2000s motors with (presumably) fiberglass.

Clayell:
> Per the review comment, that means they're wildly optimistic, since these are 40s or early 50s motors (which should probably be NGNC, but let's not open that can of worms, it'd probably put the variance too high) with steel casings, rather than 90s/2000s motors with (presumably) fiberglass.

Hm, what do you think I should put the ratio at? If irl it's around 1:0.9, should I put it at around 1:0.8 or 1:0.7?
Also, the ratio on the spin motor is incredibly terrible, it's around 1:0.435. If we want to keep the same burn time while having a reasonable ratio, then we should increase the thrust. (I have no idea how to calculate what thrust would be needed though)

NathanKell:
The ratio is bad, yes, which is very good for it being in the starting node in RP-1.

Clayell:
> The ratio is bad, yes, which is very good for it being in the starting node in RP-1.

Aw don't you think that's a little harsh? It's not like rocket motor technology drastically changed in a year or two. Also, there's no other rocket motors smaller than that for a long time, so I think it can be buffed a little bit for gameplay reasons. How about it being .05 or .1 less on the right side of the ratio compared to the next node? (which we still need to determine, what should the ratio be for the first unlockable node?)

lpgagnon:
> Also, there's no other rocket motors smaller than that for a long time, so I think it can be buffed a little bit for gameplay reasons. 

there's an obvious better solution to _that_ problem: add another, better, small SRM somewhere later. (pioneer 0/1 being the obvious use case)

Clayell:
> > Also, there's no other rocket motors smaller than that for a long time, so I think it can be buffed a little bit for gameplay reasons.
> 
> there's an obvious better solution to _that_ problem: add another, better, small SRM somewhere later. (pioneer 0/1 being the obvious use case)

My pull making the mercury posigrade motor 206 isp just got merged, so there's that. I agree that there should be more smaller srm's though.

NathanKell:
Yep, that should be fine (and the Mercury posigrades and retrogrades hopefully don't use a Mercury ECM, which means anyone can use them), so I think 'small higher-tech motors' are covered then.

It'd be great if there were more small SRMs, yes. Do BDB have any we're not using yet? We can put them in ROE...

Clayell:
> Yep, that should be fine (and the Mercury posigrades and retrogrades hopefully don't use a Mercury ECM

https://github.com/KSP-RO/RP-1/blob/8ebd7134d59e854e94d04b9a5078ba9add9248f8/Source/Tech%20Tree/Parts%20Browser/data/ROCapsules.json#L3742C20-L3742C20
https://github.com/KSP-RO/RP-1/blob/8ebd7134d59e854e94d04b9a5078ba9add9248f8/Source/Tech%20Tree/Parts%20Browser/data/ROCapsules.json#L3836C1-L3836C1
Needs to be fixed.

lpgagnon:
> pioneer 0/1 being the obvious use case

https://github.com/friznit/Unofficial-BDB-Wiki/wiki/Pioneer

sure looks like some tiny SRMs under there

Clayell:
> > pioneer 0/1 being the obvious use case
> 
> https://github.com/friznit/Unofficial-BDB-Wiki/wiki/Pioneer
> 
> sure looks like some tiny SRMs under there

Perhaps this? Unsure if it's referring to that. https://github.com/CobaltWolf/Bluedog-Design-Bureau/blob/master/Gamedata/Bluedog_DB/Parts/ProbeExpansion/Pioneer/bluedog_Pioneer_Engine.dds

Clayell:
> How about it being .05 or .1 less on the right side of the ratio compared to the next node? (which we still need to determine, what should the ratio be for the first unlockable node?)

@NathanKell Ah, you keep avoiding my question! Just want to know your thoughts, I'm thinking 1:0.8 for the first node and then 1:0.75 for the starting node. (dry mass:propellant mass)

NathanKell:
But I already told you, in my review and in the above, what I suggested doing for that motor: Change its resource to 1.11 liters, and remove the difference that makes in wet mass from the dry mass.

NathanKell:
> > pioneer 0/1 being the obvious use case
> 
> https://github.com/friznit/Unofficial-BDB-Wiki/wiki/Pioneer
> 
> sure looks like some tiny SRMs under there

Then it'd be great if someone added them to ROE and RP-1. :)

NathanKell:
Made the change to Mercury/Gemini solids.

PR #2897, opened 2023-08-08
Mk1 improvements and "Fixes"

Capkirk123:
Update Mk1 pod to remove massless tanks and generally better match ROC Mercury. Also remove integral heat shield, since it was broken anyway...


NathanKell:
Since it's already savebreaking because you can't return the capsule without adding a part (the heatshield), seems we might as well brick the RCS while saves are already broken?

Capkirk123:
I suppose I don't see why not

PR #2899, opened 2023-08-13
Pioneer Vernier Solid Motor

ryanc55:
Engine config for Pioneer vernier adjustment motor.


ryanc55:
https://github.com/KSP-RO/ROEngines/pull/215
https://github.com/KSP-RO/RP-1/pull/2175

PR #2909, opened 2023-08-30
Add RD-0108, improve RD-0110 configs

Capkirk123:
Add RD-0108 config, a more reliable, man-rated version of the RD-0107. Also adjust RD-0106 config to be the same as the RD-0107, give 107% throttle to RD-0110, and decrease mass of all RD-0110 variants

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD0110_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD0110_Config.cfg
@@ -9,13 +9,13 @@
 //
 //	Dry Mass: 410? Kg
 //	Thrust (SL): ??? kN
-//	Thrust (Vac): 304 kN
+//	Thrust (Vac): 297.9 kN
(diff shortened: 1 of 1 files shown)


NathanKell:
Yay!! :)

One thing--if the 0106 is the same as the 0107, why do both configs exist rather than (I think we do this elsewhere?) combining them?

Capkirk123:
I don't see why not

lpgagnon:
rp-1 was making 0106 unlock later but be cheaper: https://github.com/KSP-RO/RP-1/blob/ceec61b4557351f9e0adeaefc16dfae56e277968/Source/Tech%20Tree/Parts%20Browser/data/Engine_Config.json#L9836

can't speak to whether that was just nonsense (or to what the engine recosting has to say about it)

Capkirk123:
RD-0106 did fly later than the RD-0107. But since they were identical other than mounting, I don't really see any point to keeping it around.

PR #2937, opened 2024-02-03
Reduce minimum part RB mass from 2kg to 1kg

siimav:
https://github.com/KSP-RO/ROSolar/issues/18#issuecomment-1892106467

Does not entirely eliminate the issue but will significantly reduce the effect of it. Remains to be seen if this will introduce any interesting side effects.


siimav:
Turns out that the minimum RB mass feature was introduced in KSP v1.11, supposedly to mitigate PhysX issues with parts dropped from inventory.

PR #2942, opened 2024-02-18
RocketMotorMenagerie Updated RO Configs

SkyPhoenix999:
Updating the old RMM Configs

Changes:

- Fix Scale values due to model changes
- Inclusion of all engines in the mod and not just a few
- Waterfall support for all and adding nozzle glow


Capkirk123:
Looks good. However, the latest version is marked as compatible with KSP version 1.12.5, while most RO/RP-1 installs are using 1.12.3, and so the rescaled version will not be installed.

An issue has been opened to mark the latest version of RMM compatible with all KSP versions [here](https://github.com/KSP-CKAN/NetKAN/issues/9952). Once that issue is resolved, this should be safe to merge

PR #2944, opened 2024-02-21
Require SXT for RO

Clayell:
SXT is a hard dependency for RO, many textures are from SXT


Capkirk123:
RO doesn't require SXT, while some parts can use SXT models, they are correctly protected behind `NEEDS` tags.

PR #2955, opened 2024-04-09
Add Infernal Robotics Next Configs

lilynoord:
Super simple config file for Infernal Robotics Next. It doesn't do anything fancy, just makes it so that all of the parts in IR Next will show up if installed.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/MagicSmokeIndustries/RO_IR.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/MagicSmokeIndustries/RO_IR.cfg
@@ -1,0 +1,1257 @@
+@PART[IR_Rotatron_Bearing_Uncontrolled_v3]:FOR[RealismOverhaul]
+{
+        %RSSROConfig = True
+        @mass = 0.006
+	%TechRequired = start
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/RO_SuggestedMods/MagicSmokeIndustries/RO_IR.cfg
```diff
@@ -0,0 +1,1257 @@
+@PART[IR_Rotatron_Bearing_Uncontrolled_v3]:FOR[RealismOverhaul]
+{
+        %RSSROConfig = True
+        @mass = 0.006
+	%TechRequired = start
+	%cost = 5
+	%entryCost = 5
+	RP0conf = true
```
siimav, line 8: RO configs are not allowed to configure parts for RP-1. That part needs to be done in RP-1 itself using the Part Browser webapp.

lilynoord: That makes sense. I also just lost a craft due to some physics issues caused by IR, so I'm just going to close this PR.

PR #2956, opened 2024-04-15
BDB Able UPDATED Configs

SkyPhoenix999:
Updated Able Configs

- Waterfall Config Relocated
- bluedog_DeltaB_Tank Tank Mass Update
- Removed the default propellants from both the Delta A and Delta B tanks (Will start empty as AJ10 variants use different fuels)


GameData/RealismOverhaul/RO_SuggestedMods/Bluedog_DB/Able/RO_Able.cfg
```diff
@@ -65,25 +65,6 @@
 		type = Tank-Iso-Al-HP
 		volume = 1382.475
 		basemass = 0.11762		//188 kg - 70.38 kg tanks
-
```
Capkirk123, line 68: I already fixed these, you're just undoing my fixes

PR #2962, opened 2024-05-07
BDB-Atlas-Configs

SkyPhoenix999:
Accidently Deleted the Old Branch and PR at the same time, this restores it. _This time to the Proper Branch_

Renamed the Folder and Files for the Old Atlas Parts since the Atlas Update for BDB is Coming.

There are no changes to these configs from the last version of the former accidently deleted Atlas PR other than renaming the folder and files.


Capkirk123:
I already went over this and not much has changed since then so it's probably fine

Capkirk123 (approved):

SkyPhoenix999:
Updated RO Configs for the new Atlas Parts

Old Parts were moved to a Subfolder

PR #2966, opened 2024-05-29
Benjee10_SharedAssets RO Configs

SkyPhoenix999:
Used in multiple mods and needs to be standardized as such.

Covers the CBM, NDS/APAS/IDA, SSPV, and AJ10

Used in (almost) all Benjee10 Mods such as HabTech2, ACK, SOCK, MSSEV, etc

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Benjee10_SharedAssets/RO_Benjee10_sharedAssets.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Benjee10_SharedAssets/RO_Benjee10_sharedAssets.cfg
@@ -9,11 +9,13 @@
 
 
 //	APAS
+//	Source: https://forum.nasaspaceflight.com/index.php?PHPSESSID=4uisic1n140e3nlbf9fni2s7lj&action=dlattach;topic=32429.0;attach=819805;image
+//	0.8 m hatch, 1.25 m docking ring OD, 1.35 m docking port OD, 1.5 m max envelope size
 @PART[B10_APASv2]:FOR[RealismOverhaul]
 {
 	%RSSROConfig = True
 	!rescaleFactor,* = DEL
-	%rescaleFactor = 1.6
+	%rescaleFactor = 1.4286
 	@title = APAS-95 Docking Port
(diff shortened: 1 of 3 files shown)


Capkirk123:
Everything should now be scaled to the correct size. However, the scale for these parts is being set elsewhere in the mod configs (notably in the Habtech 2 configs), which means the settings from these configs are not being applied. Configs other than this one should probably be removed since these parts live in Benjee10 shared assets

SkyPhoenix999:
This line is what overrides the rescale factor in Benjee's other mods, idk why he put two separate rescaleFactors in some of his parts but this fixes it
`!rescaleFactor,* = DEL`

I can verify your NDS rescale values but not your CBM values, the 1.8x values were properly sized (and additionally are properly sized for HabTech2 at that value)

The outer ring diameter is 82 inches (2.083m) not 79 inches, [Per This Document](https://imlive.s3.amazonaws.com/Federal%20Government/ID42272755263273611130549092035837226177/SSP41148.pdf), 79 inches refers to the inner diameter, this 1.6737x rescale value is far too small. Should be between 1.7437x and 1.8x. depending on if you want to scale for inner diameter or outer. I chose 1.8x for consistency with HabTech parts while being within the correct size range.

Capkirk123:
I used Freedom-era documents for reference, so it's possible that they were changed.

The issue is not Benjee10s other mods, it is other RO configs. The rescaleFactors being set in this config are being changed by other rescale factors in other RO configs. They should be removed.

Capkirk123:
Well if the Habtech 2 configs are outdated, that's all the more reason to delete them in favor of the not-outdated configs here

Capkirk123 (changes requested):
Removing competing configs for Benjee10 shared assets parts (namely, those in Habtech2)

SkyPhoenix999:
I have a new set of HabTech configs but I still have some B9 errors to sort out, and I might wait until the new update with the large modules comes out, I'm undecided on that.

Capkirk123:
I have decided for you

Capkirk123 (approved):

SkyPhoenix999:
One Last thing, NDS Passive Port. Good to go now.

PR #2971, opened 2024-06-14
Added Waterfall support for jet engines J-58 and CR2 Ramjet

ArXen42:
Main changes:

1. Added two new templates, `rowaterfall-turbojet-afterburner` (inspired by SR-71 photos it would seem) and `rowaterfall-ramjet` (J-404 Panther in stock KSP)

2. Templates from Stock Waterfall Effects were used as starting point (do I need to put attribution somewhere, like in `Credits.txt`?)
Throttle values usage was replaced with appropriate "core throttle" and "afterburner throttle" values from AJE, some other changes and adjustments were made to exhaust/shock diamond effects, added barely visible smoke trail effect for non-afterburner thrust

3. The main challenge was adapting plume sizes to AJE's variable nozzle area feature
Currently it is still a bit messy - I had to resort to adding two controller sets which differ only in expected range of value received from `GetNozzleArea()` since it isn't normalized by default
I have currently only implemented two engines, so can't say for sure how portable nozzle area->scale curves will turn out to be later when more configs will be added

4. The audio I simply copied over from `ModuleManager` cache from unmodified RO install, so it should be exactly as it was without waterfall

At min and max nozzle area:
image
image

Current exhaust for comparison:
image

TODO:

1. Add configs for the rest of the jet engines
2. Maybe find more elegant solution to difference in how `nozzleArea` changes between different engines (if it actually will, haven't tested this thoroughly yet)

### DIFF
--- a/GameData/RealismOverhaul/Waterfall_Configs/AJE/aje.cfg
+++ b/GameData/RealismOverhaul/Waterfall_Configs/AJE/aje.cfg
@@ -1,0 +1,24 @@
+@PART[turboFanEngine]:BEFORE[ROWaterfall]:NEEDS[Waterfall]
+{
+    ROWaterfall
+    {
+        template = rowaterfall-turbojet-afterburner
+        audio = turbojet-afterburner
+        transform = thrustTransform
+        position = 0, 0, 0
+        rotation = 0, 0, 0
+        scale = 1.0, 1.0, 1.0
+    }
+}
+@PART[aje_ramjet]:BEFORE[ROWaterfall]:NEEDS[Waterfall]
+{
+    ROWaterfall
+    {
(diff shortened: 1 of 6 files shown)


Capkirk123:
Looks good! Is this ok to be merged, or is there more you are working on?

ArXen42:
> Looks good! Is this ok to be merged, or is there more you are working on?

Sorry didn't notice the message, but I think the only thing I missed is adding credits for Stock Waterfall Effects (which I used as a starting point) to `license.md`, something like this:
https://github.com/KSP-RO/RealismOverhaul/compare/master...ArXen42:RealismOverhaul:patch-1

Hopefully I will manage to make another PR with more engine configs based on these templates some time soon.

PR #2972, opened 2024-06-16
Consider all KSP 1.12 revisions the same in compatibility checks

siimav:
KSP 1.12.3 is the best KSP version. Fight me @Scorpuu.


NathanKell:
lol

PR #2986, opened 2024-07-29
Add deadly reentry to conflicts

Clayell:
Fix old players from thinking that its still required (https://discord.com/channels/319857228905447436/512556137380315139/12674272282418913990)

### DIFF
--- a/RealismOverhaul.netkan
+++ b/RealismOverhaul.netkan
@@ -113,6 +113,7 @@ conflicts:
   - name: TestFlightConfig
   - name: StockWaterfallEffects
   - name: WaterfallRestock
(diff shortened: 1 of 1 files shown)


Capkirk123:
Sounds good to me, between RO and (maybe) RSS everything is already configured for realistic thermal performance without DRE, and DRE has more than a few nasty bugs associated.

Just to check though, there's no reason we don't want to conflict DRE, right?

Clayell:
> Just to check though, there's no reason we don't want to conflict DRE, right?

It's not used as a dependency for anything RO-related so I can't see why there would be a reason.

Issue #2958, opened 2024-04-29
Some KerbalFoundries parts are physicsless


siimav:
That's how these are currently configured in the base mod itself. Figure out if RO can override this setting.

siimav:
lpg sez:
> might be a good idea to check where their CoM is before "fixing" that. I remember some of the stock gear had it in interesting places

Capkirk123:
CoM position seems pretty reasonable on all affected parts. I can't see any reason to set `PhysicsSignificance = 0` on the parts.

PR #2989, opened 2024-08-02
Enable physics on KF2 landing gear

Capkirk123:
Since there doesn't seem to be any reason not to, disable the physicless-ness of KF2 landing gear.

resolves https://github.com/KSP-RO/RealismOverhaul/issues/2958


siimav:
You sure that this change actually applies? My comment in the issue says:
> That's how these are currently configured in the base mod itself.

Capkirk123:
image
It certainly appears to. It also behaves as expected in the flight scene.

siimav:
Huh, I somehow thought that you removed those physicssignificance lines instead of adding them. IDK if I selectively saw that one and only red line.

PR #2990, opened 2024-08-05
Don't recommend outdated/incompatible mods

Clayell:
What decade are we in? /s

Reported by Cristi457 (https://discord.com/channels/319857228905447436/512556137380315139/1269741132284166255)

### DIFF
--- a/RealismOverhaul.netkan
+++ b/RealismOverhaul.netkan
@@ -97,6 +97,13 @@ suggests:
       - name: Restock
   - name: UniversalStorage
   - name: VenStockRevamp-NewParts
+supports:
+  - any_of:
+      - name: RemoteTech
+      - name: RealAntennas
+  - any_of:
+      - name: Kerbalism-Config-RO
+      - name: TACLS
 provides:
   - RealPlumeConfigs
   - RealFuels-Engine-Configs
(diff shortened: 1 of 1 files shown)


Capkirk123:
RCS Build aid changes seem fine. TACLS and RemoteTech are nominally still supported by RealismOverhaul, however. It's just RP-1 that absolutely does not work with them, which I believe has already been configured as such.

Clayell:
> RCS Build aid changes seem fine. TACLS and RemoteTech are nominally still supported by RealismOverhaul, however. It's just RP-1 that absolutely does not work with them, which I believe has already been configured as such.

Perhaps we can put them both as 'supports' instead of recommends? (as 'any of')

siimav:
> Perhaps we can put them both as 'supports' instead of recommends? (as 'any of')

I don't get what's the idea behind this. We do recommend that the player has a comms and a life support mod installed.
While TACLS and RT may be outdated, they're still supported on RO installs. It isn't up to RO to police any of this. Instead we have Express Install that lists the mods we think would be best for most players.

Clayell:
> We do recommend that the player has a comms and a life support mod installed.

Ah, ok now I get the reasoning behind this, I understand. I will revert them to how they were before.

Sorry for bugging you more, but do you mind checking https://github.com/KSP-RO/RP-1-ExpressInstall/pull/6? In my opinion thats a more urgent pr than this. (as in it fixes a ckan issue) (again sorry, ik im not obligated to have you do anything but people keep complaining about KER getting denied by ckan and I feel bad)

PR #2996, opened 2024-08-12
Add dependency on tweakscalerescaled-redist

JonnyOThan:
Since FAR depends on tweakscale-redist, users are prompted to choose between tweakscale-redist and tweakscaleRescaled-redist when installing.  Adding an explicit dependency here should remove that choice (which doesn't really matter) and prevent future problematic conflicts.

### DIFF
--- a/RealismOverhaul.netkan
+++ b/RealismOverhaul.netkan
@@ -11,9 +11,9 @@ license: CC-BY-SA
 release_status: stable
 depends:
   - name: ModuleManager
-  - name: AdvancedJetEngine
-    suppress_recommendations: true
   - name: TweakScaleRescaled-Redist
+    suppress_recommendations: true
+  - name: AdvancedJetEngine
     suppress_recommendations: true
   - name: FAR
     suppress_recommendations: true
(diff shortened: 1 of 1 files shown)


siimav:
Now the question is how does CKAN behave when someone tries to update their existing install?

JonnyOThan:
Yeah never mind, this doesn't work too well on the current version of CKAN.  It still forces you to choose when installing, and if you picked the wrong one it doesn't report the conflict until later

siimav:
To be clear - did you try with RO-only or with RO/RP-1 Express Install?

JonnyOThan:
RO only, but I don't think it would be any different.

JonnyOThan:
in ckan 1.34.5 (unreleased dev build), this change prompts an "update" action, and if you select it then you get:
image

So you'd still need to manually unselect tweakscale-redist and select tweakscalerescaled-redist (which does work).

Fresh installs still have the above problem:

image

JonnyOThan:
FYI, HebaruSan pointed out that since AJE depends on FAR, the dependency declaration needs to be before AJE for this to work properly.  I haven't yet re-tested CKAN's behavior with that change though.

JonnyOThan:
Still needs some testing but this should remove one of the "choice" screens when installing RO/RP-1

siimav:
Does the "needs some testing" part still apply?

Capkirk123:
image

Yeah sure ship it

PR #2997, opened 2024-08-13
Give KSPWheel parts some rotational resistance

siimav:
Something I've had in my install for a couple of years now and have mostly forgotten about.
Makes the wheels lose non-insignificant amounts of angular momentum over time. I believe I initially added this to combat plane landing gear breaking after takeoff. Added benefit is that rovers now actually slow down on level terrain. However it is possible that this slowdown effect is too large at high speeds so would be great if someone else gave the changes a try too.


Capkirk123:
Tested, seems fine. Wheels slow down relatively quickly. Even extremely underpowered aircraft like the Cessna don't have their takeoff roll extended too much, and most jet aircraft are barely effected at all. Rovers are also fine, wheel motors easily overpower extra drag.

PR #3006, opened 2024-09-25
Make U-1700 speculative instead of altHist

TwistedGiraffe:
It literally says it is "speculative" twice in the description. The U-1700 is about as unreal as some other engines labeled `speculative`, such as the later variants of E-1 and HG-3.
I am all ears about why the U-1700 was set to `altHist` besides early game balance.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/U1250_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/U1250_Config.cfg
@@ -151,7 +151,7 @@
 		{
 			name = U-1700
 			description = Speculative upgrade for the U-1250
-			specLevel = altHist
+			specLevel = speculative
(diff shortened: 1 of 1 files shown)


Clayell:
is there a source for the u-1700? i cant find any mention to it on the internet

Capkirk123:
> is there a source for the u-1700? i cant find any mention to it on the internet

No, because as the config mentions, it is entirely fictional.

Clayell:
Hm, I suppose this brings up a good point about what the real difference between speculative and althist should be. It's a little strange that if we have "documentation" of a fake part on a fan website somewhere, it would probably be considered alt history, but if there isn't any documentation and its being made up by RP-1, it would be speculation, which is prioritized over alt history?

PR #3013, opened 2024-11-04
Apply RO Material tags to Kerbal Foundries Parts

TwistedGiraffe:
Apply Aluminium / Instruments tags to rover wheels and small landing gears, and RCC / Instruments tags to medium and large landing gears, matching the stock wheels.


njits23:
Could this merge be reverted? Even the medium gears with the upgraded external temperature tolerance cannot survive a 2000 m/s hypersonic flight now, whilst the stock gears can. Additionally, if you put the stock gears fully inside your fuselage they experience no convective heating, whilst the KF2 gears do, making that not an option to protect them from heating
image
KF2 gears also have strange behaviour where their extended state experiences lower convective heating than the retracted state, exacerbating the overheating issue.
image
image

PR #3022, opened 2024-12-18
Airbag Description Fix

Eranos1:
Swaps the descriptions of the Mk-10 Inflatable Airbag and the Mk10-XL Inflatable Airbag, as they were opposite of intended based on description


Clayell:
I swear, this PR sounded so familiar to me until I realized that I [did exactly this](https://github.com/KSP-RO/RP-1/pull/2385/commits/32593d0c60496142ab4a1d4cada0dedbfede1999) 6 months, except that I only did it in RP-1 and forgot about RO. Pretty neat coincidence that we both stumbled upon the same issue!

PR #3037, opened 2025-02-02
Add some new engines engines

Its-Just-Luci:
Add RSRM-I, RSRM-II, RD-0124MS, and the unnamed LR91-AJ-5 modification from the SRB-X studies


Capkirk123:
I really don't see the point in the LR91 SRB-X config. It's functionally identical to the LR91-AJ-11, which would have been available at the same time (and is probably the engine the report is referencing anyway, since it also references Titan 34D).

The extended burn time is still well within the capability of the AJ-11, and the restart capability is optional, doesn't have much information on it, and doesn't really add anything considering this would have flown with a third stage?

PR #3039, opened 2025-02-07
Fix gear overheating during regular use

njits23:
Currently, all retractable landing gear directly exposed to heat levels experienced during things like hypersonic flight and faster are big fans of melting. 

This PR changes their skin and internal heat tolerance to be at minimum Inconel tier, preventing the gear from experiencing early death due to overheating when not fully shielded via clipping them inside parts. This clipping method is a workaround that only works for some gears. Now it should no longer be necessary.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
@@ -515,8 +515,8 @@
 	%RSSROConfig = True
 	@mass = 0.19
 	@PhysicsSignificance = 0
-	%skinTempTag = Aluminum
-	%internalTempTag = Instruments
+	%skinTempTag = Inconel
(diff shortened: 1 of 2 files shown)


Clayell:
Hm, it's certainly annoying that this shielding issue can't be fixed, as Inconel is very far above what most landing gear are actually made of (aluminum, including shuttle), and this PR could allow for some wacky shenanigans where suborbital (or even orbital if shallow enough?) re-entry can be done with the landing gear extended.

Shouldn't we also apply material tags to the stock landing gear so that they are the same?

njits23:
This PR also modifies the stock gear. 
As far as I can tell there's also zero change to gear thermal conduction from skin to interior based on extension status, so this doesn't even affect that. See the two screenshots I attached to #3013, the PR that broke KF2 gear thermals. It is an entirely different issue.

Clayell:
> This PR also modifies the stock gear.
Woops, I'm blind. Sorry about that.

Capkirk123:
This sucks

PR #3043, opened 2025-02-13
Make Engine Lighting Relit and Rocket Sound Enhancement conflicting in ckan

Clayell:
Normally, a mod not having configs for RO would not end up with it being listed as conflicts in ckan with RO. However, for some reason, when either of these mods are installed with ro, they seem to cause a lot of [errors](https://discord.com/channels/319857228905447436/512556137380315139/1141777198047572139), [lag](https://discord.com/channels/319857228905447436/331813459417235457/1339201005459603550), and glitches. For this reason, it is fair to list them as conflicting.

### DIFF
--- a/RealismOverhaul.netkan
+++ b/RealismOverhaul.netkan
@@ -113,6 +113,8 @@ conflicts:
   - name: StockWaterfallEffects
   - name: WaterfallRestock
   - name: DeadlyReentry
+  - name: EngineLightRelit
+  - name: RocketSoundEnhancement
 resources:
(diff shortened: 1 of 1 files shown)


Capkirk123:
Engine lighting relit is pointless with waterfall supplying it's own plume lights, but can we fix RSE? iirc it's actually useful

Capkirk123:
image
Ah fair enough yeah. Well I'm not fixing it any time soon so I guess we conflict for now

PR #3046, opened 2025-02-16
Change radiatorHeadroom value to restore analytical propellant cooling functionality

njits23:
Since PR #2876 radiators have not been working correctly. They provide zero cooling performance during analytical timewarp. The lowered temperature tolerances of the part somehow negatively affect its cooling capabilities.

Increasing the value of the radiatorHeadroom tag restores them to their old cooling performance. I don't actually know how this value is used to calculate radiator performance and the comments in the file are not particularly enlightening. Therefore, the 0.69 value for radiatorHeadroom was chosen to approximate the cooling performance of the radiators when they still had their old maxTemp = 1073.15 tags.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Utility.cfg
@@ -940,7 +940,7 @@
     %internalTempTag = Aluminum
 
     %fuelCrossFeed = False
-    @radiatorHeadroom = 0.69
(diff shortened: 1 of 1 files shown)


njits23:
At timewarp speeds of 1e6 and 6e6 there still are some issues where cooling is insufficient, but that is probably a separate issue.

njits23:
Siimav has done some digging in the code. This is how the radiator headroom functions
image
The old temperature tolerance was 1073K, and with the old 0.29 multiplier, the radiators worked up to 311K internal temperature.
The new temperature tolerance is 448K, and with the 0.69 multiplier, they work up to 309K internal temperature.

PR #2684, opened 2022-07-11
Combine RS-25 configs, update RMM

Capkirk123:
Combine the RS-25 configs into one config, add SSBE/LRBE configs, and update RMM mod support.

PR #2680, opened 2022-07-01
Add some basic localization data

Capkirk123:
Add english localization for a few generic actions, and all major North American and Soviet manufacturers.

PR #3054, opened 2025-03-28
Remove RSSROConfig lines from localizations

HebaruSan:
There are `RSSROConfig = True` lines in each of this mod's localization cfg files. Isn't that just supposed to be for parts? It'd be nice to have these removed, if so, to reduce confusion for the NetKAN bot (which currently thinks this mod has been translated into the commonly spoken natural language "RSSROConfig") and for other modders (who may use these files as a model).

Looks like @Capkirk123 has worked with these files a lot recently, tagging for notification.

Noticed during discussion of Cruesoe/VABOrganizer-RP-1-Configs#8, and this mod appears to be the origin of the phenomenon (ROCapsules is also affected).

I checked the history of the files, and nobody appears to have given a reason for these lines in any of the commits or PRs, so I don't think it was solving a problem anyone was having:

- https://github.com/KSP-RO/RealismOverhaul/commit/14dcb9aa54f366c7463e244cf2cfab8da2c45c18
- KSP-RO/RealismOverhaul#2680
- KSP-RO/RealismOverhaul#2684
- KSP-RO/RealismOverhaul#2722
- KSP-RO/RealismOverhaul#2872
- KSP-RO/RealismOverhaul#2912


siimav:
Around here, "RSSROConfig" is a *commonly spoken natural language* though!

siimav:
It's true that this tag does nothing in RO. The idea was to possibly implement something similar to what RP-1 has [here](https://github.com/KSP-RO/RP-1/blob/f6b9081ce6dc083d7e3ab24ce3458c7e2837a053/GameData/RP-1/Localization/LocPatches.cfg#L1).
NK said this:
> It's because I'm changing the meaning of some stock loc tokens so we need to make sure that incorrect translations of them don't exist (since we only replace the English tokens)

PR #3055, opened 2025-04-05
Reduce minimum part RB mass even further to 0.1kg

siimav:
Continuation of #2937.
Closes https://github.com/KSP-RO/ROSolar/issues/18.


Capkirk123:
Well I can't say no to that level of certainty

Issue #3064, opened 2025-04-16
The BE-3 engine is misconfigured


Adrts:
RealismOverhaul/Engine_Configs/BE-3_Config.cfg

TESTFLIGHT:NEEDS[TestLite|TestFlight] was incorrectly entered as TESTFLIGHT

TESTFLIGHT:NEEDS[TestLite|TestFlight] 可能被错误地输入为 TESTFLIGHT

siimav:
There's many more such cases that need review.
![Image](https://github.com/user-attachments/assets/7dec85fa-b454-440e-98b8-5626ed6d127e)

PR #3065, opened 2025-04-16
Add some missing :NEEDS[TestLite|TestFlight]

Gordon-Dry:
see https://github.com/KSP-RO/RealismOverhaul/issues/3064


Capkirk123:
resolves https://github.com/KSP-RO/RealismOverhaul/issues/3064

PR #3067, opened 2025-04-26
Nerf heatshields when their ablator runs out

siimav:
Will greatly increase skin-to-internal and skin-to-skin conduction values.

### DIFF
--- a/GameData/RealismOverhaul/RO_HSClobberer.cfg
+++ b/GameData/RealismOverhaul/RO_HSClobberer.cfg
@@ -6,6 +6,9 @@ HS_CLOBBERER_SETTINGS
 	effectTime = 5					// number of seconds over which the effects are applied
 
 	// names of parts that are excluded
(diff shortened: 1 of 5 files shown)


njits23:
This PR seems to exclude crew capsules from the conduction nerf. Why? Currently Vostok (as well as other capsules with their heatshields included) can reenter without any ablator and if its excluded from this fix it will continue to be able to do so.

Capkirk123:
Tested, seems to work as intended. With some tweaks to heat shield internal temperatures, it pretty quickly destroys heat shields if they run out of ablator during re-entry.

The only issue is the part exclusion doesn't appear to work, Mercury/Gemini also get massively increased conduction if they run out of ablator.

Capkirk123:
The part exclusion works properly now. Looks good to me!

Capkirk123 (approved):

PR #3086, opened 2025-08-16
Standalone S5.4 mass

Capkirk123:
Try to set a sensible mass for a bare S5.4 engine


Capkirk123:
Oh yeah that's true, Vostok is just HP for the nitrogen RCS

Issue #3088, opened 2025-08-17
Reliability data of RD-0105 incorrectly assigns it a cycle failure


njits23:
Currently, the RD-0105 reliability data says it suffered a cycle failure on one of the Luna 8K72 flights (for some reason labeled as Vostok in the .cfg). https://github.com/KSP-RO/RealismOverhaul/blob/460695cdaa8ed34c841d8be8b1c60c834fefa9e5/GameData/RealismOverhaul/Engine_Configs/RD0105_Config.cfg#L133 However, I have found a source that says this wasn't an engine failure, but a failure of ground personnel to fuel up the kerosine tank fully.

The book Soviet Robots In The Solar System by Wesley T. Huntress says the following on page 84: 
<img width="687" height="94" alt="Image" src="https://github.com/user-attachments/assets/c2cef19b-b770-4d7d-9d3e-16ba643a4bc7" />

Based on this, I believe the reliability data of the RD-0105 should be revised upwards. It suffered no failures on any of its flights. The only flights that included it which failed did so because the earlier stages failed.

PR #3089, opened 2025-08-17
Update RD0105 reliability to reflect its lack of actual cycle failures

njits23:
Solves https://github.com/KSP-RO/RealismOverhaul/issues/3088


Capkirk123:
resolves #3088

PR #3095, opened 2025-08-26
Added Chinese localization support 

Adrts:
Added Chinese localization support
Although it is an AI translation, a basic review has been done


Adrts:
Some errors need to be corrected first

PR #3097, opened 2025-08-30
Add some Chinese translations

Adrts:
This translation is an AI translation, but it has been reviewed.


Capkirk123:
I'd like some confirmation from a native chinese speaker

al2me6:
Oh dear.

To be direct: this is an unreasonable amount of jargon to ask someone to review. I lack the time, the will, and (most crucially) the domain expertise to review this thoroughly, since it is less about checking the grammar and more to do with looking up each technical term to ensure that the correct translation/transliteration was used. 

Given that this entire PR is AI-generated, I have serious doubts as to the correctness of the technical terms therein. Short of finding a Chinese-language rocket enthusiast (not to be confused with an English-language rocket enthusiast who knows Chinese) to properly review it for technical correctness, I recommend not merging this PR.

Adrts:
我是中国人，我自认为语言还算是比较通顺的，做了基本的审查，符合国人的语言习惯。我觉得没问题。

al2me6:
> _I am Chinese and consider myself fairly decent at the language. I did a basic review to make sure that the translations are idiomatic and don't see any problems._

You are missing the point.

There is little concern with how idiomatic the generated text is; AIs are in fact quite adept at generating idiomatic-sounding nonsense.

_That_ is not acceptable for our purposes. The text here is largely comprised of a) proper names of rather obscure (and often defunct) entities, and b) rocketry jargon. An AI does not have the necessary context to translate this text correctly. The only way to do so is to dig through journal articles (and other published, reliable sources) to find attested forms of each term. Doing a review of your AI translation will require no less effort, at which point I may as well be producing a hand translation anyway.

eggrobin:
> The text here is largely comprised of a) proper names of rather obscure (and often defunct) entities, and b) rocketry jargon. An AI does not have the necessary context to translate this text correctly. The only way to do so is to dig through journal articles (and other published, reliable sources) to find attested forms of each term.

As an example of this (albeit with obscure orbital mechanics terminology rather than obscure rocketry terminology), see https://github.com/mockingbirdnest/Principia/blob/21a8763f076599cffc1918cdf424c3ec450f2c13/ksp_plugin_adapter/localization/zh-cn.cfg#L146-L185, where the bracketed abbreviations are entries in https://github.com/mockingbirdnest/Principia/blob/master/documentation/bibliography.pdf. The technical terms there are obscure enough that we can sometimes trace them back to a single source, and so their translations likewise need to be tracked down to works on the same subjects. This takes extremely specific domain expertise in addition to knowledge of the language.

PR #3098, opened 2025-08-31
Nerf landing gear, make separate spaceplane gear

Capkirk123:
Since apparently some people can't behave when given re-entry related parts in the start node, nerf landing gear heat tolerance and create spaceplane versions instead.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
@@ -515,6 +515,7 @@
 	%RSSROConfig = True
 	@mass = 0.19
 	@PhysicsSignificance = 0
+	//Aluminum probably, but I don't wanna make an X-15 rated version too so just say Inconel hot structure
 	%skinTempTag = Inconel
 	%internalTempTag = HotStructure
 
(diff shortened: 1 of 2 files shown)


njits23:
Also I'm pointing this out at my own peril, but you didn't do anything about the Kerbal Foundries landing gears. https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/KerbalFoundries_2/KF-RO.cfg
Please just change the pre-spaceplane gear to temp resistances that survive a 2000 m/s flight instead of titanium-instrument. They'll melt with instrument internal temperatures, because the conduction will just be too great. I suppose as an alternative to re-upping the temp tolerances you could try and reduce the conduction to their internals from all sources, but what you have here now simply won't work.
I'm not sure about if the titanium skin temp will be exceeded. because I forget its exact value. That might be fine.

Capkirk123:
Alright fine but if somebody figures out how to use those for re-entry I'm making X-15 tier gear too

njits23:
Based on my experiences with heatsink heatshields I don't see anything with a 1100K or lower temperature tolerance survive an Earth reentry without a massive amount of braking dV. Besides that, fairings with their 2000K temp tolerance and their density turned all the way up make way better cheese heatshields than non-RCC landing gears ever could.

PR #3099, opened 2025-09-01
Make ORM-65 performance more realistic

Capkirk123:
Considering the given ORM-65 performance numbers almost certainly refer to sea level performance, compute vacuum performance with RPA (assuming similar efficiency to WAC-Corporal) and adjust ORM-65 stats to match. Also increase variance a little and tweak throttled performance.


Clayell:
Can we confirm the reliability? Wikipedia (quoting astronautix) says that the missile was cancelled due to the numerous explosions of the engines.
(this config should also have some sources listed)

jjang36524:
Currently the varience is centered on 210s ISP, which leads to lower than IRL performance
It should be centered on 212,5s ISP

PR #3118, opened 2025-12-26
Add persistent rotation to conflicting on ckan

Clayell:
RO provides persistent rotation, so installing the mod along with RO is bound to cause some problems


siimav:
No issues there. RO implementation will disable itself in presence of PR or Principia.

Clayell:
Ah I forgot about that. Probably should have checked first

PR #3119, opened 2026-01-16
Fix unrealistically long LR87 spool up time

MrTagnan:
Sets a throttle response rate so that the LR87s don't take unrealistically long to spool up. Based on available footage and audio spool up took around a second for the hypergolic variants, in game they take around ~3 seconds right now. Kero and LH2 variants are my best guess when accounting for the need for a separate ignition system.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/LR87LH2_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/LR87LH2_Config.cfg
@@ -160,7 +160,7 @@
 			heatProduction = 175
 			ignitions = 2
 			massMult = 1.25149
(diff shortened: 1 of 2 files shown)


MrTagnan:
I can compile a full list of sources I referenced to estimate the spool up time if needed. Mostly audio from the various gemini-titan launches and slow mo footage of a test firing. Couldn't find anything on the LH2 variant, and couldn't find any good video of the AJ-1/3

MrTagnan:
Actually I might adjust the numbers somewhat. MJ launches too quickly after engine start

siimav:
Please update the same PR instead of making a new one.

PR #3120, opened 2026-02-20
Allow SoundingPayload in all HP tanks

periodically-makes-puns:
Currently only the 4 earliest HP tank types (first 3 Conventional, first Isogrid) are configured to allow SoundingPayload. https://github.com/KSP-RO/RP-1/pull/2692 (Operational and Advanced Earth Observation) has a contract that uses SoundingPayload, intended to be taken with 1972 tech available. By that point most players will have better conventional and isogrid tanks.

### DIFF
--- a/GameData/RealismOverhaul/RO_RealFuels.cfg
+++ b/GameData/RealismOverhaul/RO_RealFuels.cfg
@@ -130,6 +130,7 @@ TANK_DEFINITION
 	addResources = true
 	addResourcesHP = true
 	addLead = true
+	addSoundingPayload = true
 	maxUtilization = 90
 	defaultMass = 0.00009
 }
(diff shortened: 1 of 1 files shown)


periodically-makes-puns:
Closing as irrelevant, as the contract that requires sounding payload in EOS2/EOS3 has been dropped. See https://github.com/KSP-RO/RP-1/pull/2707#issuecomment-4085470489.

WithCrow:
While I deleted the contract, I do think that this PR would still be good nevertheless

Clayell:
I also agree, there is no reason for sounding payload to not be available in all HP tanks, their exclusion seems like an oversight.

siimav:
Why litter the list when nobody (sane at least?) is launching SRs with isogrid tanks

WithCrow:
The first tier of isogrids already have SP, though - the latter do not. I made use of it as a stand-in for experiments in the EOS successor programs, so I do see some value in it being an option.

PR #3122, opened 2026-02-22
Remove multiple ignition capability from S1.5400

Nazfib:
It had a single set of pyrotechnic ignition charges (main combustion chamber, gas generator, pressurization/roll control system, turbopump starter).

In the 90's, a drop-in replacement chemical igniter was developed after production of the original solid-fueled charges had stopped; those still only provided the single ignition, though.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD58_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD58_Config.cfg
@@ -17,7 +17,7 @@
 //	Prop Ratio: 2.48?
 //	Throttle: N/A
 //	Nozzle Ratio: 100?
-//	Ignitions: 5
+//	Ignitions: 1
 //	=================================================================================
 //	11D33
 //	Molnyia-M (Blok ML, Blok MVL)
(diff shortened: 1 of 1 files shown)


siimav:
Looks like you are correct but I have to say that I hate it. With these changes you are forced to go from really reliable 11D33(M) to the rather abysmal RD-58 to get any kind of multi-ignition capability.

Nazfib:
True, but the reliability is no worse (better, even) than the RL-10A-1 (which is the first American non-hypergolic restartable engine, unless I'm forgetting one; it was also developed for a very similar purpose, high-energy upper stages for heavy lunar/interplanetary probes and geostationary comsats).

Note that the RD-58 was a completely new engine; it was only related to the 11D33/S1.5400 in that it was developed at the same OKB, which means that there was a bunch of tech transfer.

siimav:
Yeah but the second config of RL10 is awesome already. Second and third S1.5400 configs used to be equally great but now to get >1 ignition you have to use the 4th config which is kind of a turd reliability-wise. So realistically more like 5th config if you want something reliable with multiple ignitions.

PR #3127, opened 2026-03-18
Add Knes Configs

WithCrow:
This PR aims to configure most of the relevant parts in Knes, which adds more European spacecraft.


WithCrow:
I've added the Multi-Role Capsule & speclevels (currently all are speculative). I deem this to be somewhat reasonable, as all craft (Viking, Hermes & MRC) are based on real proposals.

WithCrow:
Requires ROKerbalism configs, which can be found in [this PR](https://github.com/KSP-RO/ROKerbalism/pull/191)

GameData/RealismOverhaul/RO_SuggestedMods/Knes/KNES_Viking.cfg
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    RESOURCE
+    {
+        name = Ablator
+        amount = 120
+        maxAmount = 120
+    }
+
+    !RESOURCE,* {}
```
pap1723: This is going to delete the Ablator resource that you just added above it.

GameData/RealismOverhaul/RO_SuggestedMods/Knes/KNES_Viking.cfg
```diff
@@ -0,0 +1,437 @@
...
+        TANK
+        {
+            name = LithiumHydroxide
+            amount = 1.0
+            maxAmount = 1.0
+        }
+    }
+}
+
+@PART[Knes_Heatshield_09375]:FOR[RealismOverhaul]
```
pap1723, line 84: This should use the new heat shield tagging system instead of manually setting values here.

GameData/RealismOverhaul/RO_SuggestedMods/Knes/KNES_Viking.cfg
```diff
@@ -0,0 +1,437 @@
...
+            amount = 280
+            maxAmount = 280
+        }
+        TANK
+        {
+            name = Oxygen
+            amount = 1000
+            maxAmount = 1000
+        }
+        TANK
```
pap1723: Best to use the existing configuration system here. Makes it more robust for anything that might change in the future 

%setDefaultTACLS = true
%defaultTACLSPax = 1.0
%defaultTACLSSupply = <duration>

GameData/RealismOverhaul/RO_SuggestedMods/Knes/KNES_MRK.cfg
```diff
@@ -391,9 +391,36 @@
 	{
 		name = ModuleFuelTanks
 		type = ServiceModule
-		volume = 631.008
+		volume = 750.008
 		basemass = -1
 
+		TANK
```
pap1723, line 462: Same comment as the Viking about using the configured additions with TACLS

pap1723 (changes requested):
A few things to clean up, but otherwise it'll be nice to have some new crew options.

GameData/RealismOverhaul/RO_SuggestedMods/Knes/KNES_Viking.cfg
WithCrow: Wow, it's *far* easier to use as well!

WithCrow:
I hope to have implemented all of your requested changes now - and wow, if only I had been doing that from the start! Far easier than doing it all manually!

WithCrow:
In order to prepare parts for the RP-1 part browser, those with a leading underscore are instead copied, with the prefix 'RO' (i.e. 'RO_Knes_French_Hermes' instead of '_Knes_French_Hermes'). Prior to this change the entry cost modifiers would not apply in RP-1, as the leading underscore is turned into a leading hyphen during the RP-1 config generation

WithCrow:
As per Veyl's suggestion, the original part names are now overwritten instead of copying the entire part.

WithCrow:
FAR configs for Callisto & Hermes have been added - I could run from them no longer for VentureStar, so I also went ahead and measured them here.

WithCrow:
I've switched everything over to using the RO_Materials instead of 'hardcoded' thermal values & changed to ROLS.

WithCrow:
I have had enough! Callisto is such an aerodynamic nightmare, it will simply not fly! Unless anyone has an idea other than not giving it FAR configs, I am going to simply remove it. It handles like hell & it is useless!

WithCrow:
I did not have it in me to kill Callisto. The solution I took was to use procedural control surfaces, note how large they needed to be and then apply those stats to the evelons. Furthermore, I've increased Callisto's size overall, which should make it a bit more useful.

WithCrow:
image
Kliper (shown on the left) and three additional smaller station parts have also been added. The former is a real proposal by Energia to replace Soyuz, the latter an extrapolation of 'What if they wanted to do Columbus Man-Tended Free Flyers on Ariane 4?'

WithCrow:
Slow and steady wins the race! PAP (Ariane 4) and EAP (Ariane 5) solids now work

PR #3130, opened 2026-03-26
Limit AdjustableCoMShifter offsets to within part bounds

### LINKED ISSUE #2628: The CoM should only be able to be moved within the volume of the avionics
Suggested by [Dank_Jeb](https://discord.com/channels/319857228905447436/1115321536807718992/1409282342375260371). Obviously not an easy problem to solve for non-cylindric avionics, but it would help to curb abuse of the CoM shifter.

siimav:
Upgrade pipeline is still scuffed and I refuse to sink more hours into this. So if anyone has a vessel in SCM queue or storage then you need to launch it; revert to VAB; launch again.

Resolves https://github.com/KSP-RO/RP-1/issues/2628

### DIFF
--- a/Source/AdjustableCoMShifter.cs
+++ b/Source/AdjustableCoMShifter.cs
@@ -16,16 +16,22 @@ namespace RealismOverhaul
         [KSPField]
         public Vector3 DescentModeCoM = Vector3.zero;   // If configured, disable customization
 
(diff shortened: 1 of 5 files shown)


siimav:
# PSA: the version numbers need to be updated to reflect the exact RO version this change ships in

PR #3143, opened 2026-05-09
Fix issues with Vexin/Valois/Vesta ignitions

### LINKED ISSUE #280: Vesta/Vexin/Valois (all three) cannot ignite mid-air despire stats saying it can
https://medal.tv/games/kerbal-space/clips/mmvFdO2g0f0mPWZM4?invite=cr-MSw3Y00sNDExMDMyNjEw&v=27

so i was testing out my downrange rocket and read that the french gemstone engine can be ignited. tested it and it refuses to work, as shown in the medal link. double checked my ullage, fuel tank and i notice it can only ignite with clamps on the launchpad.

Glloyd:
Fixes https://github.com/KSP-RO/ROEngines/issues/280
and https://github.com/Capkirk123/ROEngines-Extended/issues/9

All three engines were improperly configged and could only be groundlit despite having one ignition. I've fixed this, and also made the ignition resources align with those for Veronique and its children, but let me know if you'd like it implemented differently.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/Vexin_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/Vexin_Config.cfg
@@ -149,6 +149,12 @@
 
 			IGNITOR_RESOURCE
 			{
+				name = ElectricCharge
+				amount = 0.5
+			}
+
+			IGNITOR_RESOURCE
+			{
 				name = Furfuryl
 				amount = 1
 			}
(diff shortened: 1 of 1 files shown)


Gordon-Dry:
As I stumbled upon this PR while I needed to find a solution after assuming initially it was an issue with MechJeb or whatever I just can say that the fix is simple and works.
Should be merged.

PR #3147, opened 2026-05-12
Add VentureStar Configs

WithCrow:
This PR adds configs for the MK-33 mod, with them specifically 'interpreted' as VentureStar. VentureStar would have featured 7 RS-2200s, yet the mod only allows for four to be placed next to each other - due to this, the current RS-2200 part is actually 1.75 RS-2200s. I could not think of a better solution.


WithCrow:
Sigh - branch in my fork was based off the Knes branch. Will open new PR

PR #3148, opened 2026-05-12
Add VentureStar Configs

WithCrow:
This PR adds configs for the MK-33 mod, with them specifically 'interpreted' as VentureStar. VentureStar would have featured 7 RS-2200s, yet the mod only allows for four to be placed next to each other - due to this, the current RS-2200 part is actually 1.75 RS-2200s. I could not think of a better solution.


WithCrow:
There are a few important To-Dos/problems, highlighted thanks to Creativit's expertise with SSTOs - which is why this has been converted to a draft.

Another thing I am considering: it would be possible to also add a proper X-33. While it is somewhat useless on its own (as it is merely suborbital), by combining it with a first stage (or adding an upper stage to its back)

WithCrow:
The above flaws have been fixed! Should be a-okay now.

WithCrow:
Thanks to Chris for testing it out and making some small fine tuning changes!

PR #3154, opened 2026-05-17
Add Moderately Plane Related Configs

WithCrow:
This PR adds configs for the Moderately Plane Related cockpit parts. Initially, this is the
- F-100
- F-104
- U-2 (Using the model of the F-104 cockpit) 
- F-4 

If there is interest in them, I can add configs for the other more modern cockpit parts.


WithCrow:
image
Family photo!
image
And compared to some of the other cockpits.

PR #3168, opened 2026-06-06
Revert RD-119 ISP to 352

Blothorn:
The fact that this early engine outperforms the staged-combustion UDMH/LOX engines despite its much lower chamber pressure and only marginally greater expansion ratio has been bothering me, and I went down a rabbit hole.

This was changed to 362s on the basis of https://web.archive.org/web/20230814113330/http://library.voenmeh.ru/cnau/elr03216.pdf, contradicting a consensus of 352s from all other sources (including some Russian ones, including an NPO Energomash document). However, that source is itself inconsistent: it gives three different values for ISP/exhaust velocity (all vacuum).
- Page 4, text: 3551m/s (362s)
- Page 8, summary of overall engine statistics: 3551m/s (362s)
- Page 10, text: 352s
- Page 14, summary of the combustion chamber: 3512m/s (358s)
 
We can also calculate ISP from the mass flow and thrust; 100.5kN from 30.82kg/s mass flow gives an even lower value of 332s. (Notably the combustion chamber table gives somewhat different mass flow rates, 170g/s less oxydizer and 1020g/s less fuel, and a different nominal mixture ratio; I suspect the discrepancy comes from the monopropellant gas generator. Still, using the lower combustion chamber mass flow rate still only brings ISP to 345. Using the higher thrust value of 106kN found in other sources gives 351s--but the exclusion of the gas generator mass flow makes that optimistic.) I think these inconsistencies, along with oddities such as the claimed expansion ratio of 1280, call into question the reliability of this source. While the author may have had access to superior primary sources, they seem to have uncritically mixed that information with some from less reliable sources. Since they give a bibliography but not footnotes, it is impossible to tell which of these values is most trustworthy without gaining access to the primary sources.

### DIFF
--- a/GameData/RealismOverhaul/Engine_Configs/RD109_Config.cfg
+++ b/GameData/RealismOverhaul/Engine_Configs/RD109_Config.cfg
@@ -25,7 +25,7 @@
 //	Dry Mass: 168 kg
 //	Thrust (SL): 65.6 kN
 //	Thrust (Vac): 105.5 kN
-//	ISP: 220 SL / 362 Vac
+//	ISP: 220 SL / 352 Vac
 //	Burn Time: 260
 //	Chamber Pressure: 7.9 MPa
 //	Propellant: LOX / UDMH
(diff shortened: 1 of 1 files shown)


GameData/RealismOverhaul/Engine_Configs/RD109_Config.cfg
periodically-makes-puns: The vacuum Isp needs to be changed here.

Nazfib:
The specific impulse value of 362s is certainly too high.

That said, I'm not sure that the Matveev/Semelov document is entirely unreliable, with the notable exception of the specific impulse.

The mass flow rates _do_ add up correctly. The overall engine takes in 18.47 kg/s of oxidizer and 12.35 kg/s of fuel. The combustion chamber takes in 18.3 kg/s of oxidizer and 11.33 kg/s of fuel; the gas generator burns 0.985 kg/s of fuel. The missing 0.170 kg/s of oxidizer goes through a heat exchanger before it's used to pressurize the oxidizer tank. The missing 0.055 kg/s of fuel gets mixed with some of the gas generator exhaust and is then used to pressurize the fuel tank.

Regarding the thrust; it has a "nominal" thrust of 100.5 kN, and can vary the thrust +/- 5%. The maximum thrust is therefore 105.525 kN, which matches with other sources. Notably, this includes the thrust of the gas generator exhausts.

Regarding the expansion ratio; it seems that the Soviet convention was to measure the expansion ratio as the ratio of gas pressure between the combustion chamber and pressure at the exhaust plane; with the stated combustion pressure and nozzle exit pressure, I get a value of 7.89e6/6.18e3=1276. This is in unlike the American convention, where the expansion ratio is the ratio of the nozzle exit area to the throat area (which would give an expansion ratio for the RD-119 of 102.2).

Still, this leaves the specific impulse as a mystery. The stated overall specific impulse (either 352s or 362s) don't fit with the given thrust and mass flow rate; nor the specific impulse of the main nozzle (358s) with the mass flow rate given for the combustion chamber. Only if I take the _nominal_ propellant consumption, and the _maximum_ thrust, do I get values anywhere near 352 m/s. If I take nominal thrust and nominal mass flow, I get an Isp of 336s; which is still _very_ good for the time, but not an enormous leap above everything else. 

Finally, there is only one start powder charge and several pyrotechnically activated startup/shutdown valves; the -119 was definitely not restartable. The "pumped idle" mode mentioned in the config is from the S5.23 on Kosmos 1/3/3M, not the RD-119 on Kosmos 2.

For now, setting the specific impulse back to 352s, making the thrust 95.5-105.5 kN, and removing the extra ignition, is the best we can do with the available information. If any details about how the Isp was measured come to light (or if someone can make a better argument based on the available data), it can always be revisited.

PR #3172, opened 2026-07-01
Module Manager fixes for RN mods

StonesmileGit:
The same two errors seems to have been copied to multiple locations. See the last two commits, one for each type of error.
(the first commit includes another fix for Tantares with the same issue)


raidernick (approved):
looks good

PR #3175, opened 2026-07-13
Adds FFT extra engine/utilities

gabo463:
this pr aims to adds extra very endgame engine and utilities to rp1/RO

more info bellow


gabo463:
Pretty big update on that one :

Adjusted all engine to be more inclined into real life concept and research
New subtype and propellent to tanks (fusion) to be more inclined on real life research and concept ( Fusion Pellets and Tritium)
new fuel mixtures 
Adjusted the ISRU instrument to integrate some of the new subtype and to be more inclined into real life concept and research
Adjusted the Fusion reactors to be more inclined into real life concept and research
image_2026-07-28_192321759

WithCrow:
ISRU would be better placed in ROKerbalism

gabo463:
Even tho there are no kerbalisim elements being used ?

WithCrow:
They should use Kerbalism elements

gabo463:
alright on it

gabo463:
lithium integration. lithium is primarily used at the moment for tritium production using the the Integrated Isotope & Pellet Fabrication Facility

gabo463:
i do not agree with the fact of an NSWR  beeing 20 meters i dont see no documentation on this and its way bigget than anyting in the game i left the Orion drive alone because that is what some documentation says and so: 

reduced NSWR to 10 meters in diameter 
reduced 2 20 meter NSWR tank to 5 meters

PR #3177, opened 2026-07-13
Adds additional kerbal-atomics engines

gabo463:
this pr goal is to pretty much complete the nuclear branch of the tree while also introducing unique nuclear engines whit 3 additional nuclear engine 1 NTR and 1 Aerospikes


gabo463:
updated outdated code

PR #3180, opened 2026-07-21
Add an High-Altitude Scientific Research Balloon from the Kerballoons mod

gabo463:
this pr adds the universal balloon form the  kerballoons mod and so contains a :

High-Altitude Scientific Research Balloon (Universal)

Capture d’écran 2026-07-21 175817

image from:https://youtu.be/0Oj4jqqJQOA?si=Nh_Pg8Eo0rrs1liJ


siimav:
That mod isn't MM-patchable and I'm pretty sure you get some nonsensical stock configs by default.

gabo463:
i only added the balloon of the mod i dont see why it not being MM patchable would cause a problem since its a balloon  also i did not notice any crazy stock config  

i havent tested it yet myself but testing from others using the config showed that it is functional

PR #3181, opened 2026-07-22
Fix LK size

LuneoLucy:
RN's LK is much bigger than it should, this PR fix this to match real life's stowed numbers (according to RussianSpaceWeb).

LK is now 2.26m wide stowed (legs excluded) and 4.5m wide deployed (legs included).

Might make interfacing with RN's N1 awkward though.

### DIFF
--- a/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_LK_LOK.cfg
+++ b/GameData/RealismOverhaul/RO_SuggestedMods/RaiderNick/RO_RN_LK_LOK.cfg
@@ -3,9 +3,9 @@
 	%RSSROConfig = True
 	@MODEL
 	{
-		@scale = 1.0, 1.0, 1.0
+		@scale = 0.74, 0.74, 0.74
 	}
-	@scale = 1
+	@scale = 0.74
 	@rescaleFactor = 1
 	@mass = 1.336 //1.436-0.1 for one crew mass
 	@manufacturer = MOM - Korolev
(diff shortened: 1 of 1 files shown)


siimav:
Closing this — the numbers don't hold up, even against the source cited. The RussianSpaceWeb page you linked lists max landing gear span as 5.4m, not the 4.5m used in the PR, and the claimed 2.26m stowed diameter isn't physically plausible for this vehicle based on available reference imagery.

Current model: stowed diameter 4.5m, deployed/landing-gear span 5.4m, height 5.2m (matched to imagery for the height/diameter ratio), and it fits the fairing correctly. Fair point that the stowed landing gear sticks out a bit more than it should, due to the folding animation (~4.5m vs 4.3m gap) — that's a valid critique.

But resizing the LK to match the PR's figures isn't feasible without remodeling the N1 as well, since the N1 assembly depends on the LK's current spacing. Not something we're doing for this.

LuneoLucy:
> and the claimed 2.26m stowed diameter isn't physically plausible for this vehicle based on available reference imagery.
image_2026-08-12_131436501
It is according to every schematic/cutaway I could find online, and it is consistent across Nick Stevens' renders, the most reliable western source for the N1 program, here's a quick and dirty proof using QCAD, the image is cropped to match the LK's height and scaled to 5200px, the width of the base is roughly 2300px
Even if you disagree with the changes I am reopening this PR until a solution is found, the LK is unusable in replicas as is (it needs to fit inside a 4.1m fairing) and a remodel is definitely not necessary in this case

LuneoLucy:
> I am reopening this PR

Nevermind I do not have the permissions to do so :p

LuneoLucy:
@siimav Please take another look at this

PR #3190, opened 2026-08-19
Update Raptor Config

SkyPhoenix999:
Change Raptor 3 thrust from 280tf to 250tf, and RVac thrust from 306tf to 275tf inline with real Raptor 3 performance.

Added a subconfig to allow for the original 280tf spec which is a stretch goal for the real engine.

Slight change to the mixture ratio to be inline with Starship Expansion Project and Flip&Burn.


gabo463:
Flip&Burn has been discontinued recently I’m pretty sure. Just a heads up in case it affects anything

SkyPhoenix999:
Not an issue, SEP runs on the same mixture ratios and this isn't dependent on either of them, the numbers were ran for those mods and I'm bringing the MEC inline with it.

PR #3191, opened 2026-08-20
adds a 7 meter docking port (Extra docking ports)

gabo463:
pretty much what the title says


gabo463:
squashed the part  to be more pleasing for the eyes

